PR #2, opened 2017-07-24
Add basic layout.

tpendragon:
It looks like typography's still off, but haven't been able to figure it
out yet.


tpendragon:
screen shot 2017-07-24 at 3 54 45 pm

app/views/shared/_add_content.html.erb
```diff
@@ -0,0 +1,41 @@
...
+        </li>
+      <% end %>
+      <% if include_collections_link %>
+        <li class="divider"></li>
+        <li><%= link_to 'Add a Collection', main_app.new_collection_path, class: 'menu-heading new-collection', role: 'menuitem' %></li>
+      <% end %>
+      <% if include_roles_link %>
+        <li class="divider"></li>
+        <li><%= link_to 'Manage Roles', role_management.roles_path, class: 'menu-heading manage-roles', role: 'menuitem' %></li>
+        <li><%= link_to 'Workflow Roles', hyrax.admin_workflow_roles_path, class: 'menu-heading manage-workflow-roles', role: 'menuitem' %></li>
```
escowles: These won't work, right?

config/locales/en.yml
```diff
@@ -17,3 +17,13 @@ en:
 
   titles:
     application: Figgy
+  valkyrie:
+    institution:
+      name: 'Princeton University Library'
+      homepage_url: 'https://library.princeton.edu'
+    product_name: 'Digital PUL'
```
escowles: I'm going to create a separate ticket for this, but I've been thinking that DPUL is the brand for the frontend, so Plum/Figgy/whatever should have a different brand.

jpstroop: 👍

app/views/shared/_add_content.html.erb
tpendragon: This whole block is commented out for now.

tpendragon: (Well, not commented - bypassed. At the top there's an `<% if false %>`)

config/locales/en.yml
tpendragon: Totally. Just making the layout work for now.

PR #6, opened 2017-07-26
Import Metadata from Pulfa/Bibdata

tpendragon:
I'm doing my best to avoid VCR.


app/models/plum_schema.rb
```diff
@@ -1,328 +1,346 @@
 # frozen_string_literal: true
 module PlumSchema
   extend ActiveSupport::Concern
-  included do
-    # Hyrax
-    attribute :depositor
-    attribute :title
-    attribute :label
-    attribute :visibility
+  def imported_schema
```
escowles: I like the idea of `imported_schema`, but it does make me think we should have a `MarcRelatorsSchema` etc. classes to clearly identify where some of these are coming from.

app/models/remote_record.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def attributes
+    JSONLDBuilder.for(jsonld).result
+  end
+
+  private
+
+    def jsonld_request
+      @jsonld_request ||= Faraday.get("https://bibdata.princeton.edu/bibliographic/#{source_metadata_identifier}/jsonld")
```
escowles: TODO: we shouldn't be generating bibdata and pulfa URLs in our apps — I've added a ticket to pul_metadata_services for this: https://github.com/pulibrary/pul_metadata_services/issues/18

spec/support/stub_bibdata.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+module BibdataStubbing
```
escowles: 👍  This is a lot simpler than VCR, and in particular, I'm looking forward to having human-readable sample records on disk instead of VCR's gnarly raw captures.

Gemfile
```diff
@@ -16,7 +16,7 @@ gem "simple_form"
 gem "sprockets", ">= 3.0.0"
 gem "title"
 gem "uglifier"
-gem "valkyrie", git: "https://github.com/samvera-labs/valkyrie.git"
+gem "valkyrie", git: "https://github.com/samvera-labs/valkyrie.git", branch: "fix_date_parsing"
```
escowles: You can point this at master now that samvera-labs/valkyrie#180 is merged

app/models/plum_schema.rb
tpendragon: Can we make an issue for that? I'm not sure the neatest way to bisect them. The only reason this exists is so that importing metadata won't wipe out all the non-imported metadata (IE `source_metadata_identifier`)

escowles: 👍  https://github.com/pulibrary/figgy/issues/7

PR #9, opened 2017-07-27
File Manager with FileSets

tpendragon:


tpendragon:
https://vimeo.com/227323204 Very messy demo of file manager (files are no longer required, though.)

valhalla/Gemfile.lock
```diff
@@ -0,0 +1,213 @@
+PATH
```
escowles: Do you want the engine's Gemfile.lock to be checked in?

PR #10, opened 2017-07-27
Test Postgres in CI

tpendragon:
It's more accurate, and the runtime is basically the same.

### DIFF
--- a/config/valkyrie.yml
+++ b/config/valkyrie.yml
@@ -3,7 +3,7 @@ development:
   storage_adapter: disk
 
 test:
-  metadata_adapter: memory
+  metadata_adapter: postgres
   storage_adapter: test_disk
 
 production:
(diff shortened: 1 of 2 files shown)


config/valkyrie.yml
```diff
@@ -3,7 +3,7 @@ development:
   storage_adapter: disk
 
 test:
-  metadata_adapter: memory
+  metadata_adapter: postgres
```
escowles: Can we make this configurable with env variables?  Or some other way to make it easier to run the tests locally without having to have postgres setup?  Not that it's a huge burden, but it would be nice not to need postgres setup locally to run the tests.

tpendragon: I think for our case it'd be best to just require having postgres up. We need it for Blacklight anyways, other projects require it, and it means we can make some assumptions about features (like transactions)

escowles: I'm fine with this for now — but I do think it's worth exploring this a bit locally and with the Data Mapper WG.  If different backends have different characteristics (which of course they do, or why bother?), do we need to be more explicit about those differences, and the tradeoffs of choosing one over another?

PR #37, opened 2017-08-01
File Characterization

tpendragon:
Adds height/width/mime_type and three different checksums to files on ingest.


app/characterization_services/tika_file_characterization_service.rb
```diff
@@ -0,0 +1,67 @@
...
+    @file_characterization_attributes = { width: result['tiff:ImageWidth'], height: result['tiff:ImageLength'], mime_type: result['Content-Type'], checksum: checksum }
+    @file_node = @file_node.new(@file_characterization_attributes.to_h)
+    @persister.save(resource: @file_node) if save
+    @file_node
+  end
+
+  # Provides the SHA256 hexdigest string for the file
+  # @return String
+  def checksum
+    MultiChecksum.new(
```
escowles: In Java, you can read a stream and run it through multiple checksum tools to generate multiple checksums with one file read.  Can we do something like that here?  This would presumably benefit from file system caching, but I would expect it to be considerably slower than a single read.

tpendragon: I can't seem to find a way to do it - does the thing you mention have a name?

escowles: Looks like the process is very similar to how it works in Java: you read the file in chunks and update multiple checksum objects with each chunk: https://gist.github.com/escowles/acc2c0d5c5b3fc5724ae2e768c398d0b

tpendragon: Great! Done

PR #42, opened 2017-08-03
Add thumbnails to Catalog/File Manager/Structure

tpendragon:
This does a maybe-slow-if-postgres-wasn't-so-fast recursive lookup to
find the derivative FileMetadata resource ID, so it can generate a IIIF
path to it. Right now it seems fast enough - we'll have to keep an eye
on it.


app/models/file_set.rb
```diff
@@ -4,4 +4,8 @@ class FileSet < Valhalla::Resource
   attribute :id, Valkyrie::Types::ID.optional
   attribute :title, Valkyrie::Types::Set
   attribute :member_ids, Valkyrie::Types::Array
+
+  def thumbnail_id
+    @thumbnail_id ||= Valkyrie.config.metadata_adapter.query_service.find_members(resource: self).find { |x| x.use.include?(Valkyrie::Vocab::PCDMUse.ServiceFile) }.try(:id)
```
escowles: This took me a second to figure out before understanding that we are finding the pcdmuse:ServiceFile which the image server is using to generate the thumbnail.  Maybe we should move the image API suffix in here instead of returning the service file from `thumbnail_id`?

tpendragon: I agree this is confusing, but what do you mean by image API suffix? I wondered if we just need a derivative_file_metadata method

escowles: I mean the "/full/!200,150/0/default.jpg" part that gets added at https://github.com/pulibrary/figgy/pull/42/files#diff-55f651679ae90d845871424d7d4e9bd3R180

So this method would return the entire thumbnail URL

tpendragon: I want to have it not return the whole URL so we have only one place generating manifest URLs, which isn't attached to the model itself, so that it can vary between dev/staging. However, I moved some code around so hopefully the intent is more clear - have a look/

PR #46, opened 2017-08-04
Add Rights Statement CV/Show

tpendragon:


app/decorators/scanned_resource_decorator.rb
```diff
@@ -20,6 +20,21 @@ def metadata_adapter
     Valkyrie.config.metadata_adapter
   end
 
+  def rights_statement
```
escowles: I recently added a `rights_statement_uri` method to Plum to make it easier to access the URI instead of the HTML rendering.  Is that the best way to allow access to both forms?  Or would it make more sense to have this have a different name like `rendered_rights_statement`?

app/assets/javascripts/rights_note.js
```diff
@@ -0,0 +1,16 @@
+/* toggle rights note field based on which rights statement is selected */
+$(document).ready(function(){
+    $('.rights-statement').change(rights_note_visibility);
+});
+
+function rights_note_visibility() {
+    var stmt = $('.rights-statement');
+    var note = $('.rights-note');
+    if ( stmt.val() == 'http://rightsstatements.org/vocab/NKC/1.0/' ) {
```
escowles: I wonder if we could attach the notable boolean to the `<option>` as an attribute or something to avoid duplicating the logic of which statements are notable here?

escowles:
This looks good - just a couple of minor questions.

app/decorators/scanned_resource_decorator.rb
tpendragon: I can change it to `rendered_rights_statement`

app/assets/javascripts/rights_note.js
tpendragon: Oh, sure - I just copied what existed.

escowles: I'm happy to make a separate ticket for that

tpendragon: Nah I went ahead and did it.

PR #50, opened 2017-08-07
Add holding location CV.

tpendragon:


app/models/controlled_vocabulary.rb
```diff
@@ -14,6 +14,10 @@ def all
     []
   end
 
+  def find(value)
+    all.find { |x| x.value == value }
```
escowles: This is fine for small vocabs — but it won't scale well...

tpendragon: Default implementation I think.

escowles: 👍

PR #73, opened 2017-08-07
Store imported attributes in a nested record.

### LINKED ISSUE #8: Try out using nested records for imported metadata
A `ScannedResource` could have a `imported_metadata` property which is another `Valkyrie::Resource`. When you re-import metadata you can just wipe out that property.

tpendragon:
Closes #8

### DIFF
--- a/spec/models/scanned_resource_spec.rb
+++ b/spec/models/scanned_resource_spec.rb
@@ -6,14 +6,6 @@ require 'valkyrie/specs/shared_specs'
 RSpec.describe ScannedResource do
   let(:resource_klass) { described_class }
   it_behaves_like "a Valkyrie::Resource"
-  let(:schema_klass) do
-    Class.new(Valkyrie::Resource) do |k|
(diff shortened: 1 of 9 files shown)


spec/models/scanned_resource_spec.rb
```diff
@@ -11,9 +11,6 @@
       k.include PlumSchema
```
escowles: Should this go too?

escowles:
The one thing I thought of is that there might be a big overlap between the fields that are moving to the nested resource here, and the EphemeraFolder.  But I think we can probably refactor PlumSchema to make it easy to include those in a reasonable way later.

tpendragon:
@escowles Yeah, I figured we'd tackle that when we do MVW.

spec/models/scanned_resource_spec.rb
tpendragon: Yup, removed.

PR #79, opened 2017-08-08
Refactors the PlumSchema into Schema Modules

### LINKED ISSUE #20: Add all the appropriate fields to show view of Scanned Resources/Multi-Volume Works

jrgriffiniii:
Resolves #20 and #7 by addressing the following:
- Refactoring metadata within PlumSchema into `Schema::*` Modules
- Adds explicitly defined "suppressed" attributes for decorated Valkyrie Resources


app/models/schema/plum/hyrax.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+##
+# Defining the attributes for all Resources within Plum
+module Schema
+  module Plum
+    ##
+    # Defines the attributes migrated from Hyrax core
+    module Hyrax
```
escowles: I am wary of naming any of our modules or classes "Hyrax" — mostly for symbolic reasons and to avoid confusion if/when Hyrax support for Valkyrie is undertaken.

I also think it would be nice to split up the terms by source vocabulary: many of these are from DC/DCTerms or MARC Relators.

app/models/schema/plum/imported.rb
```diff
@@ -0,0 +1,309 @@
...
+  module Plum
+    ##
+    # Defines the attributes migrated from Hyrax core
+    module Imported
+      extend ActiveSupport::Concern
+
+      def self.attributes
+        [
+          # Used for a license,
+          :license,
```
escowles: As with the comment above, breaking these out by source would be nice.

app/models/schema/plum/local.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+module Schema
+  module Plum
+    ##
+    # Defines the attributes local to Plum
+    module Local
```
escowles: Same as above: breaking these out by source would be nice, since some of these properties are our own local properties, but some (alternative, identifier, etc.) are DC.

app/models/schema/common.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    ##
+    # Retrieve the attributes from Hyrax, imported from external systems, and local
+    def self.attributes
+      Plum.attributes + Schema::IIIF::Book.attributes
+    end
+
+    included do
+      Common.attributes.each do |field|
+        attribute field
```
escowles: Is there a way to avoid this boilerplate code in each schema?

jrgriffiniii: That was my primary struggle with this PR, I'll definitely see about this immediately.

app/decorators/valkyrie/resource_decorator.rb
```diff
@@ -1,5 +1,25 @@
 # frozen_string_literal: true
 class Valkyrie::ResourceDecorator < ApplicationDecorator
+  self.suppressed_attributes = [
```
escowles: I'm not sure I understand what's going on here, since at first blush these look like attributes we'd want to display.  Is this related to moving the imported attributes to a nested resource, so these get suppressed on the parent resource?

app/models/schema/plum/local.rb
jrgriffiniii: Apologies, I misunderstood this from within #7 - I can address this as well.

app/models/schema/common.rb
escowles: It's fine if it's not possible — or we could make a separate ticket for looking into that later so it doesn't block this work.

app/decorators/valkyrie/resource_decorator.rb
jrgriffiniii: No, I will document this in the code base for clarity.  This an attempt to structure the following View logic from Plum:

https://github.com/pulibrary/plum/blob/master/app/views/hyrax/base/_attributes.html.erb#L8
https://github.com/pulibrary/plum/blob/master/app/schemas/plum_schema.rb#L302

app/models/schema/common.rb
jrgriffiniii: Apologies, it is likely to continue to block this work, as I would likely need to abandon any (simple) approach using Modules.  Within the scope of `included`, I do not believe that I can easily reference the Module being mixed into the Class.  From what I could determine, this is not a limitation of `ActiveSupport::Concern`:

https://gist.github.com/jrgriffiniii/8cf9ab64d4ecaff44a2f59284d23ba85

escowles: 👍

tpendragon: You could always make them not modules - add a method like `apply_schema` to Valhalla::Resource (or via a module I guess) and each schema be a class with an interface that `apply_schema` can utilize.

jrgriffiniii: Agreed, something similar was what I was considering for #80 , but I was wary of further blocking this issue.

PR #95, opened 2017-08-09
ARK minting

### LINKED ISSUE #54: Mint Arks on Complete

escowles:
* Mint ARKs when records are completed
* Update ARK metadata when remote metadata is updated

Fixes #54

### DIFF
--- a/app/change_set_persisters/plum_change_set_persister.rb
+++ b/app/change_set_persisters/plum_change_set_persister.rb
@@ -56,7 +56,6 @@ class PlumChangeSetPersister
 
     def after_save(change_set:, updated_resource:)
       append(append_id: change_set.append_id, updated_resource: updated_resource) if change_set.append_id.present?
-      IdentifierService.mint_or_update(updated_resource) if change_set.try(:state_changed?) && change_set.new_state == 'complete'
     end
 
     def before_delete(change_set:)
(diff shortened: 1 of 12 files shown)


app/change_set_persisters/plum_change_set_persister.rb
```diff
@@ -56,6 +56,7 @@ def before_save(change_set:)
 
     def after_save(change_set:, updated_resource:)
       append(append_id: change_set.append_id, updated_resource: updated_resource) if change_set.append_id.present?
+      IdentifierService.mint_or_update(updated_resource) if change_set.try(:state_changed?) && change_set.new_state == 'complete'
```
tpendragon: I'm shocked this works - by the time after_save has worked the ChangeSet has `sync`d already, right?

escowles: It works because it saves the resource directly: https://github.com/pulibrary/figgy/pull/95/files#diff-41aae6e0457a6b7d231238cba013d0d1R8

Should we be minting the ARK earlier and getting it in before the change_set is persisted?

tpendragon: Oh, I don't think this works. I don't think it ever gets called. There's only one spec, which sets a source_metadata_identifier, so it's being called as part of refreshing remote metadata. Part of why I don't like postfix conditionals - SimpleCov would have caught this

escowles: Oh, I'll add another test that calls this one and see what we need to do to make sure it gets called when there's no remote metadata.

tpendragon: Probably need a before_save hook to stash whether or not you need to identify.

app/models/remote_record.rb
```diff
@@ -21,7 +21,7 @@ def success?
     end
 
     def client_result
-      @client_result ||= PulMetadataServices::Client.retrieve(source_metadata_identifier)
+      @client_result ||= PulMetadataServices::Client.retrieve(Array.wrap(source_metadata_identifier).first)
```
tpendragon: This isn't necessary if you run `prepopulate!` on the change_set

PR #100, opened 2017-08-09
Add workflow gated search.

### LINKED ISSUE #78: Ensure gated search works for workflow states

tpendragon:
Closes #78.


app/models/search_builder.rb
```diff
@@ -16,6 +17,25 @@ def filter_parented(solr_params)
...
+    state_string += "has_model_ssim:Collection"
+    solr_params[:fq] << state_string
+  end
+
+  def readable_states
+    all_states - unreadable_states
+  end
+
+  def all_states
+    BookWorkflow.new(nil).aasm.states.map(&:name).map(&:to_s)
```
escowles: This can be `BookWorkflow.new(nil).valid_states`

PR #101, opened 2017-08-09
Add state transition widget to show pages.

### LINKED ISSUE #48: Add state transition widget to show pages
Should display current state and allow transitioning to other valid states, and commenting

### LINKED ISSUE #68: Add workflow note property

tpendragon:
Closes #68, closes #48.

### DIFF
--- a/app/change_sets/workflow_note_change_set.rb
+++ b/app/change_sets/workflow_note_change_set.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+class WorkflowNoteChangeSet < Valkyrie::ChangeSet
+  validates :author, :note, presence: true
+  property :note, multiple: false, required: false
+  property :author, multiple: false
+end
--- a/app/assets/stylesheets/components/form.scss
+++ b/app/assets/stylesheets/components/form.scss
@@ -3,3 +3,6 @@
     margin-left: 0;
   }
 }
+.form-group .radio input.radio_buttons {
+  margin-left: 0;
+}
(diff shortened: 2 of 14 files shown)


app/change_sets/workflow_note_change_set.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class WorkflowNoteChangeSet < Valkyrie::ChangeSet
+  validates :author, :note, presence: true
```
escowles: The property definition below says note isn't required (which is what I'd expect), but this is validating its presence.  Why define the property as non-required and then enforce its presence?

escowles:
Do we want to have the state also available on the edit form?  Or should we change that to a hidden field so there's only one place to edit state?

app/change_sets/workflow_note_change_set.rb
tpendragon: Because it's weird required. You can not fill it out, it just won't make a comment. If you set required to true here it'll make a required label and chrome will prevent the form from submitting without a comment.

tpendragon:
I should be able to just remove it from the edit form.

app/change_sets/workflow_note_change_set.rb
escowles: 👍

tpendragon: Basically I'm using the validations here too determine whether or not to create a comment

PR #115, opened 2017-08-11
Add a content-addressable storage adapter.

### LINKED ISSUE #111: Write a content-addressable storage adapter.
There's no reason to waste space or time on the same file twice if we don't have to. They don't have to share FileSets or even FileMetadata, but the file on disk should be the same.

tpendragon:
Closes #111

This doesn't actually take advantage of the content-addressable storage - theoretically we could use this to make Characterization and
  Derivative Generation of duplicate files MUCH faster, but we'd have to
assume that we're using content addressable storage.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -77,3 +77,5 @@ RSpec/VerifiedDoubles:
     - 'spec/validators/viewing_hint_validator_spec.rb'
     - 'spec/validators/viewing_direction_validator_spec.rb'
     - 'spec/models/user_spec.rb'
+Lint/UnusedMethodArgument:
+  AllowUnusedKeywordArguments: true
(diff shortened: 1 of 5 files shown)


tpendragon:
Been thinking about this - in order to use this to de-dup derivatives and source files, I think FileMetadata resources need to not be embedded. The only issue is that would slow down displaying derivatives and generating manifests, because the type of the resource is stored as part of the resource.

I think we can fix that if we do an embedded node on FileSets which just has a file_identifier property and a type - so we can immediately pluck out the derivative.

tpendragon:
Oh! This is actually hard. The only way to avoid creating a new FileMetadata resource is to query for it. Which means we'd need a new query - `find_by(file_identifier:)`

tpendragon:
https://github.com/pulibrary/figgy/tree/dont_reingest_duplicates uses this to drastically speed up characterization/derivatives of duplicate content. (300 duplicates with a worker that has 10 threads finishes in ~ 58 seconds.)

PR #128, opened 2017-08-23
Add deploy configuration for Staging

tpendragon:
Includes:

1. Cantaloupe path generation
2. BrowseEverything parallelization
3. Capistrano deploy files.


config/fedora.yml
```diff
@@ -0,0 +1,20 @@
...
+  password: fedoraAdmin
+  url: http://localhost:8984/rest
+  base_path: /dev
+test:
+  user: fedoraAdmin
+  password: fedoraAdmin
+  url: http://localhost:8986/rest
+  base_path: /test
+production:
+  user: <%= ENV["PLUM_FEDORA_USER"] || "fedoraAdmin" %>
```
escowles: Do we want to make these `FIGGY_...`?

config/deploy.rb
```diff
@@ -0,0 +1,55 @@
...
+set :application, 'figgy'
+set :repo_url, 'https://github.com/pulibrary/figgy.git'
+
+# Default branch is :master
+# ask :branch, `git rev-parse --abbrev-ref HEAD`.chomp
+set :branch, ENV['BRANCH'] || 'master'
+
+# Default deploy_to directory is /var/www/my_app_name
+# set :deploy_to, '/var/www/my_app_name'
+set :deploy_to, '/opt/figgy'
```
escowles: Do we want to use `/opt/rails_app`?  I have mixed feelings about this: it's nice to be able to tell which app you're dealing with, but it's also nice to know the path regardless of which app it is.

config/fedora.yml
tpendragon: Oh. I guess. It never actually uses this file, staging just won't spin up without it

config/deploy.rb
tpendragon: I like /opt/figgy personally. We only had rails_app because the puppet recipe was easier to write that way, but I think we have that solved in ansible. Maybe we can ask everyone's opinion?

escowles: Not hearing any objection on Slack, I think we can stick with `/opt/figgy`

config/fedora.yml
escowles: Maybe we should file a Valkyrie ticket for that?  Seems strange to require a Fedora configuration file when no Fedora adapter is configured...

tpendragon: Agreed

PR #135, opened 2017-08-25
Force SSL.

tpendragon:


escowles:
We should force ssl in staging too, right?

tpendragon:
Staging requires the production the config, so this does it for both

escowles:
👍

PR #137, opened 2017-08-28
Extends IndexingAdapter to save metadata for member resources

### LINKED ISSUE #83: Translate Visibility to Child Works / File Sets on Save

jrgriffiniii:
Resolves #83 by extending the IndexingAdapter to propagate permissions and state for member resources during a save

### DIFF
--- a/app/change_set_persisters/plum_change_set_persister.rb
+++ b/app/change_set_persisters/plum_change_set_persister.rb
@@ -52,6 +52,13 @@ class PlumChangeSetPersister
     def before_save(change_set:)
       apply_remote_metadata(change_set: change_set)
       create_files(change_set: change_set)
+
+      return if !change_set.changed?(:visibility) && !change_set.changed?(:state)
+      members(change_set: change_set).each do |member|
+        member.read_groups = change_set.read_groups if change_set.read_groups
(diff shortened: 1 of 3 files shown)


app/adapters/indexing_adapter.rb
```diff
@@ -41,7 +41,23 @@ def index_persister
     # @note This saves into both the `persister` and `index_persister`
     #   concurrently.
     def save(resource:)
-      composite_persister.save(resource: resource)
+      if resource.respond_to?(:member_ids) && !resource.member_ids.empty?
```
tpendragon: This should be part of `PlumChangeSetPersister`

app/change_set_persisters/plum_change_set_persister.rb
```diff
@@ -52,6 +52,12 @@ def with(metadata_adapter:)
     def before_save(change_set:)
       apply_remote_metadata(change_set: change_set)
       create_files(change_set: change_set)
+
+      members(change_set: change_set).each do |member|
+        member.read_groups = change_set.read_groups
```
escowles: We should probably only do this if the read_groups or state have changed, otherwise, we'll be touching all the FileSets in a ScannedResource any time we make a metadata change.

PR #140, opened 2017-08-28
Display original_file's tech metadata on FileSet show page

### LINKED ISSUE #63: Display technical metadata in FileSet properties.

escowles:
Fixes #63

### DIFF
--- a/app/models/file_metadata.rb
+++ b/app/models/file_metadata.rb
@@ -10,6 +10,7 @@ class FileMetadata < Valkyrie::Resource
   attribute :original_filename, Valkyrie::Types::Set
   attribute :file_identifiers, Valkyrie::Types::Set
   attribute :use, Valkyrie::Types::Set
+  attribute :size, Valkyrie::Types::Set
 
   def self.for(file:)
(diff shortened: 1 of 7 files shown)


app/models/file_metadata.rb
```diff
@@ -10,16 +10,23 @@ class FileMetadata < Valkyrie::Resource
...
 
   def derivative?
-    use.include?(Valkyrie::Vocab::PCDMUse.ServiceFile)
+    use_uris.include?(Valkyrie::Vocab::PCDMUse.ServiceFile)
   end
+
+  private
+
+    def use_uris
+      use.map { |u| u.is_a?(Hash) ? u[:@id] : u }
```
tpendragon: Why's this necessary? Valkyrie should support mapping back/forth to RDF::URI.

escowles: When the use values are coming back from Solr, they come back as hashes like `{"@id":"http://pcdm.org/use#OriginalFile"}` — I'll try to figure out why Valkyrie isn't converting them to URIs.

escowles: Oh, it's probably because the use values are nested objects inside the FileMetadata nested objects.

escowles:
I've updated this now that https://github.com/samvera-labs/valkyrie/pull/206 fixed the nested URI issue.

PR #144, opened 2017-08-29
Extends the File Management interface for regenerating derivatives

### LINKED ISSUE #64: Regenerate Derivatives Button

jrgriffiniii:
Resolves #64 by adding a route for updating FileSet derivatives and providing a jQuery extension for the File Management interface

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,6 +4,11 @@ machine:
   services:
     - redis
 dependencies:
+  pre:
+    # Install the latest release of Google Chrome for functional JavaScript testing
+    - curl -L -o google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+    - sudo dpkg -i google-chrome.deb
+    - rm google-chrome.deb
   cache_directories:
     - kakadu
   post:
--- a/spec/controllers/file_sets_controller_spec.rb
+++ b/spec/controllers/file_sets_controller_spec.rb
@@ -26,4 +26,24 @@ RSpec.describe FileSetsController do
       expect { get :edit, params: { id: file_set.id.to_s } }.not_to raise_error
     end
   end
+
+  describe "PUT /file_sets/id" do
+    context 'with a derivative service for images in the TIFF' do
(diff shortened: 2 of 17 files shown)


circle.yml
```diff
@@ -4,6 +4,11 @@ machine:
   services:
     - redis
 dependencies:
+  pre:
+    - curl -L -o google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+    - sudo dpkg -i google-chrome.deb
+    - sudo sed -i 's|HERE/chrome\"|HERE/chrome\" --disable-setuid-sandbox|g' /opt/google/chrome/google-chrome
```
escowles: I assume we're adding the `disable-setuid-sandbox` because we need that option for our tests to work properly.  Can you add a comment here?

spec/controllers/file_sets_controller_spec.rb
```diff
@@ -26,4 +26,23 @@
...
+      before do
+        allow(original_file).to receive(:mime_type).and_return('image/tiff')
+        allow(file_set).to receive(:original_file).and_return(original_file)
+        allow(create_derivatives_class).to receive(:perform_later).and_return(success: true)
+      end
+
+      it "can regenerate derivatives" do
+        put :derivatives, params: { id: file_set.id.to_s }
+
+        expect(response).to redirect_to(file_set)
```
escowles: Can you add an expectation that the derivatives class has received the perform_later call?

spec/support/capybara_selenium.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+require 'capybara/rspec'
+require 'selenium-webdriver'
+
+Capybara.register_driver(:headless_chrome) do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: { args: %w(headless disable-gpu) }
```
escowles: Now that I see this, I wonder if we need to modify the script above.  Can we just add the `disable-setuid-sandbox` option here instead?

escowles:
Looks good, just a few little things.

PR #153, opened 2017-08-31
Adds support for RabbitMQ messaging 

### LINKED ISSUE #104: Fire RabbitMQ Events

jrgriffiniii:
Resolves #104 by adding support for RabbitMQ messaging for the creation, updates, and deletion of Figgy resources

In addition to the improvements identified by @tpendragon , the following must be addressed:
- [x] Rebase using the latest commits on `master`
- [x] Refactor the messaging methods in `PlumChangeSetPersister` as registered handlers
- [x] Increase the coverage and quality of the test suites
- [x] Implement handlers for `after_save_commit` and `after_delete_commit`

### DIFF
--- a/app/change_set_persisters/plum_change_set_persister.rb
+++ b/app/change_set_persisters/plum_change_set_persister.rb
@@ -19,13 +19,20 @@ class PlumChangeSetPersister
       after_save: [
         AppendToParent
       ],
+      after_save_commit: [
+        PublishMessage::Factory.new(operation: :update)
+      ],
       before_delete: [
         CleanupMembership::Factory.new(property: :member_of_collection_ids),
         CleanupMembership::Factory.new(property: :member_ids),
         DeleteReferenced::Factory.new(property: :member_of_vocabulary_id)
       ],
+      after_delete_commit: [
+        PublishMessage::Factory.new(operation: :delete)
+      ],
       after_commit: [
-        Characterize
+        Characterize,
+        PublishMessage::Factory.new(operation: :create)
       ]
     }
   end
@@ -51,6 +58,7 @@ class PlumChangeSetPersister
       before_save(change_set: change_set)
       persister.save(resource: change_set.resource).tap do |output|
         after_save(change_set: change_set, updated_resource: output)
+        after_save_commit(change_set: change_set, updated_resource: output) unless transaction?
         after_commit unless transaction?
       end
     end
@@ -58,6 +66,7 @@ class PlumChangeSetPersister
     def delete(change_set:)
       before_delete(change_set: change_set)
       persister.delete(resource: change_set.resource).tap do
+        after_delete_commit(change_set: change_set) unless transaction?
         after_commit unless transaction?
       end
     end
@@ -90,6 +99,10 @@ class PlumChangeSetPersister
       yield self.class.new(metadata_adapter: metadata_adapter, storage_adapter: storage_adapter, transaction: true, characterize: @characterize, handlers: handlers)
     end
 
+    def messenger
+      @messenger ||= ManifestEventGenerator.new(Figgy.messaging_client)
+    end
+
     private
 
       def before_save(change_set:)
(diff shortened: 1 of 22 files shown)


app/change_set_persisters/plum_change_set_persister.rb
```diff
@@ -19,10 +19,15 @@ def save(change_set:)
   end
 
   def delete(change_set:)
+    if change_set.resource.is_a?(FileSet)
```
tpendragon: Can we move this to `before_delete`?

tpendragon: Or even better, a method that `before_delete` calls.

app/change_set_persisters/plum_change_set_persister.rb
```diff
@@ -19,10 +19,15 @@ def save(change_set:)
...
+      updated_resource = change_set.resource.decorate.parents.first unless change_set.resource.decorate.parents.empty?
+    end
     before_delete(change_set: change_set)
     persister.delete(resource: change_set.resource).tap do
       after_commit unless transaction?
     end
+    messenger.record_updated(updated_resource) if updated_resource
```
tpendragon: Same, can we move to `after_commit`?

tpendragon: And maybe we need like a...`after_delete_commit`

app/change_set_persisters/plum_change_set_persister.rb
```diff
@@ -69,6 +81,12 @@ def before_save(change_set:)
 
     def after_save(change_set:, updated_resource:)
       append(append_id: change_set.append_id, updated_resource: updated_resource) if change_set.append_id.present?
+      # Update after having appended a FileSet to a Work, or updated a Work
```
tpendragon: Extract to method

app/decorators/file_set_decorator.rb
```diff
@@ -4,4 +4,12 @@ class FileSetDecorator < Valkyrie::ResourceDecorator
   def manageable_files?
     false
   end
+
+  def parents
+    Valkyrie::MetadataAdapter.find(:indexing_persister).query_service.find_parents(resource: self).to_a
```
tpendragon: Extract `query_service` to its own method please.

spec/controllers/file_sets_controller_spec.rb
```diff
@@ -46,4 +52,14 @@
       end
     end
   end
+
+  describe "DELETE /concern/file_sets/id" do
+    render_views
```
tpendragon: Is this necessary?

valhalla/app/controllers/concerns/valhalla/resource_controller.rb
```diff
@@ -23,6 +27,7 @@ def create
         change_set_persister.buffer_into_index do |buffered_changeset_persister|
           obj = buffered_changeset_persister.save(change_set: @change_set)
         end
+        messenger.record_created(obj)
```
tpendragon: I wonder if we need `after_created` in our change set persister.

app/change_set_persisters/plum_change_set_persister.rb
```diff
@@ -17,14 +17,19 @@ def self.registered_handlers
...
         CleanupMembership::Factory.new(property: :member_of_collection_ids),
         CleanupMembership::Factory.new(property: :member_ids)
       ],
+      after_delete: [
+        PublishMessage::Factory.new(operation: :delete)
+      ],
       after_commit: [
-        Characterize
+        Characterize,
+        PublishMessage::Factory.new(operation: :create)
```
tpendragon: So I think we need things like `after_save_commit` and `after_delete_commit`. We don't want to fire messages for objects that are still locked up in a transaction.

tpendragon:
So I don't think those `after_delete_commit` messages are firing if you're actually in a transaction. This spec fails:

```ruby
      it 'publishes messages in a transaction', run_real_derivatives: false, rabbit_stubbed: true do
        resource = FactoryGirl.build(:scanned_resource)
        change_set = change_set_class.new(resource, characterize: false)
        change_set.files = [file1]

        output = nil
        change_set_persister.buffer_into_index do |buffer|
          output = buffer.save(change_set: change_set)
        end

        expected_result = {
          "id" => output.id.to_s,
          "event" => "UPDATED",
          "manifest_url" => "http://www.example.com/concern/scanned_resources/#{output.id}/manifest",
          "collection_slugs" => []
        }

        expect(rabbit_connection).to have_received(:publish).at_least(:once).with(expected_result.to_json)
      end
```

Fixing this is hard though. I think we need to queue up events from the buffered persister, pass them into the un-buffered persister, and then run them.

PR #185, opened 2017-09-11
Refactor PlumChangeSetPersister

tpendragon:
I'm not sure if this is better. This would reduce the churn in
PlumChangeSetPersister by a lot, but it makes knowing what exactly is
going on at each step a lot harder by moving that information out to the
registration.

### DIFF
--- a/app/change_set_persisters/plum_change_set_persister.rb
+++ b/app/change_set_persisters/plum_change_set_persister.rb
@@ -1,6 +1,29 @@
 # frozen_string_literal: true
 class PlumChangeSetPersister
+  class_attribute :registered_handlers
+  self.registered_handlers = {}
+
+  def self.register_handler(action, handler)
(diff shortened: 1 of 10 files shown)


app/change_set_persisters/plum_change_set_persister.rb
```diff
@@ -1,6 +1,29 @@
...
+  class_attribute :registered_handlers
+  self.registered_handlers = {}
+
+  def self.register_handler(action, handler)
+    raise "Invalid action #{action} given. Must be one of #{valid_actions.to_sentence}." unless valid_actions.include?(action)
+    registered_handlers[action] ||= []
+    registered_handlers[action] << handler
+  end
+
+  def self.valid_actions
```
jrgriffiniii: Just out of curiosity, why can't these be modified in the same manner as registration handlers?

PR #195, opened 2017-09-13
Add Ephemera Folder Workflow

tpendragon:
Also DRYs this up a bit to make it easier to add another workflow later.


spec/controllers/ephemera_folders_controller_spec.rb
```diff
@@ -168,6 +168,7 @@
 
   describe "update" do
     let(:user) { FactoryGirl.create(:admin) }
+    it_behaves_like "a workflow controller", :ephemera_folder
```
jrgriffiniii: 👍

Issue #168, opened 2017-09-11
Basic Ephemera Models


escowles:
Create Valkyrie models for Ephemera Projects, Boxes and Folders

PR #197, opened 2017-09-14
Supports controlled vocabularies for Ephemera Projects

jrgriffiniii:
Addresses #168 with the models for controlled vocabularies within Ephemera Projects


tpendragon:
I think there's some refactoring and generalizations we'll want to do soon, as well as some cleanup (if you delete a vocab it leaves a bunch of empty references), but for now I think we're good.

Merging.

PR #288, opened 2017-09-29
Load controlled vocabs for Templates.

### LINKED ISSUE #279: Ephemera template form should have the same controlled vocabs as the folder form

tpendragon:
Closes #279.

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -9,6 +9,7 @@ class TemplatesController < ApplicationController
   )
 
   before_action :find_parent, only: [:new, :create, :destroy]
+  before_action :load_fields, only: [:new]
 
   def find_parent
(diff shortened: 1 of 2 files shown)


app/controllers/templates_controller.rb
```diff
@@ -35,4 +36,19 @@ def resource_params
...
+      when 'subject'
+        @subject = field.vocabulary.categories
+      else
+        instance_variable_set("@#{field.attribute_name}", field.vocabulary.terms)
+      end
+    end
+  end
+
+  def fields
+    @fields ||= find_parent.decorate.fields
```
jrgriffiniii: Not blocking, but are there any general guidelines for when related resources should be retrieved from an instance variable and dynamically decorated, or when the decorate resource should, itself, be stored in the instance var.?

Issue #223, opened 2017-09-20
Refactoring for Valhalla Decorators and Controllers


jrgriffiniii:
As identified during a discussion on 09/20/17, @eliotjordan identified elements of functionality which could be refactored between the decorators and controllers.  Concerns blocking this refactoring were possibly some mixins which are more appropriately structured for the Valhalla Gem.

tpendragon:
Cold we expand on this ticket into specifics?

jrgriffiniii:
Certainly, I would imagine that the following methods _may_ be generic to any Valkyrie-based repository application to warrant their relocation to the `Valhalla` Gem.  However, I welcome input from @escowles, @tpendragon, @eliotjordan, and all others should these remain specific to `figgy`:

# Decorators
## `metadata_adapter`
- ScannedResourceDecorator#metadata_adapter
- EphemeraProjectDecorator#metadata_adapter
- EphemeraBoxDecorator#metadata_adapter
- EphemeraFolderDecorator#metadata_adapter
- EphemeraTemplateDecorator#metadata_adapter
- EphemeraFieldDecorator#metadata_adapter
- EphemeraVocabularyDecorator#metadata_adapter
- EphemeraTermDecorator#metadata_adapter

## `members`
- ScannedResourceDecorator#members
- EphemeraProjectDecorator#members
- EphemeraBoxDecorator#members
- EphemeraFolderDecorator#members

## `member_of_collections`
- ScannedResourceDecorator#member_of_collections
- EphemeraBoxDecorator#member_of_collections
- EphemeraFolderDecorator#member_of_collections

## `rendered_rights_statement`
- ScannedResourceDecorator#rendered_rights_statement
- EphemeraFolderDecorator#rendered_rights_statement

# Controllers
## `manifests`
- CollectionsController#manifests
- ScannedResourcesController#manifests
- EphemeraFoldersController#manifests

## `load_collections`
- ScannedResourcesController#load_collections
- EphemeraBoxesController#load_collections
- EphemeraTemplatesController#load_collections
- EphemeraFoldersController#load_collections

## `load_vocabularies`
- EphemeraFieldsController#load_vocabularies
- EphemeraFoldersController#load_vocabularies
- EphemeraTermsController#load_vocabularies

hackmastera:
Revisit after #641

tpendragon:
[...]

PR #291, opened 2017-09-29
Add VectorWork work type

### LINKED ISSUE #211: Migrate VectorWork
- [ ] Model / metadata
- [ ] ChangeSet
- [ ] Controller
- [ ] Decorator

hackmastera:
fixes #211, refs #223


hackmastera:
@eliotjordan do we need to add `VectorWork` to this list? https://github.com/pulibrary/figgy/blob/0fe25d1811cc6d0df43295a564753c70a8aa0786/app/models/ability.rb#L145

eliotjordan:
@hackmastera Yes we do. Good catch.

hackmastera:
@eliotjordan cool I will add it thanks

app/controllers/concerns/browse_everythingable.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+module BrowseEverythingable
```
tpendragon: Should we just add this to `Valhalla::ResourceController`?

app/decorators/concerns/member_of_collections_display.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+module MemberOfCollectionsDisplay
```
tpendragon: This makes me pretty uncomfortable. I'm not willing to stop the PR over it, but I wonder if we need a different pattern.

tpendragon: (I don't know what that pattern is - maybe some sort of global property rendering registration?)

tpendragon: In fact, maybe it's better to just add `member_of_collections` to the default decorator and have the individual decorators add it to `display_attributes` if they need it?

app/decorators/ephemera_template_decorator.rb
```diff
@@ -2,12 +2,6 @@
 class EphemeraTemplateDecorator < Valkyrie::ResourceDecorator
   self.display_attributes = []
 
-  delegate :query_service, to: :metadata_adapter
-
-  def metadata_adapter
```
tpendragon: All of this moving `metadata_adapter` to the default decorator is really great, btw.

app/controllers/concerns/browse_everythingable.rb
eliotjordan: Some controllers don't make use of BrowseEverything, `EphemeraFieldsController` for instance, but do include `Valhalla::ResourceController`. If that isn't an issue, then this seems reasonable. Alternately, we could put this into `BaseResourceController`.

app/decorators/concerns/member_of_collections_display.rb
eliotjordan: I think the default decorator is a good spot for `member_of_collections`.

app/models/scanned_map.rb
```diff
@@ -15,11 +15,16 @@ def self.can_have_manifests?
     true
   end
 
+  # How is this used?
```
tpendragon: This is a good question - I'm not sure it is anymore. It used to be the label on display in a lot of places for Hyrax, but I think that's a different method in the decorator now.

tpendragon:
The PR looks great to me. Couple questions about the use of modules, but those can just be tickets we create.

app/controllers/concerns/browse_everythingable.rb
tpendragon: I'm not sure it's a problem. I wonder if there's some sort of trigger or organization we should be thinking about which is along the lines of "has filesets" and "doesn't have filesets"

hackmastera:
Re: browse_everythingable and member_of_collections_display: in both cases I argued against including the behavior in a base class since not all inheriting classes would use the behaviors. I'm not sure why that bugs me! But I will accept the collective "it's cool" and get over it :)

hackmastera:
I got rid of those two modules. @tpendragon what do you think of the BaseResourceController in general? would you prefer to see all this as part of Valhalla::ResourceController?
 I wasn't sure how to make that decision. https://github.com/pulibrary/figgy/blob/441e8e39613258ad14620c694fb10c1653d40e9b/app/controllers/base_resource_controller.rb

tpendragon:
I'm not sure, really. It's true not everything needs `browse_everything`. Maybe keep the BaseResourceController and see how it goes over time?

hackmastera:
@tpendragon that sounds good. In which case I think this PR is ready to go, unless there's something I lost track of.

PR #302, opened 2017-09-29
Add start & end dates to form.

### LINKED ISSUE #280: Ephemera should have start and end dates
These can be year only, and we should be able to retrieve by intermediate years (e.g., start=2000, end=2012, retrieve by 2006)

tpendragon:
Closes #280.

This is a little more complicated than needs be, but it creates a nested object for date ranges so we could potentially add multiple pairs later on. It's a need we've recognized before when talking about Visuals stuff, so I thought it'd be a good idea to see what it looks like to implement.

### DIFF
--- a/spec/change_sets/ephemera_folder_change_set_spec.rb
+++ b/spec/change_sets/ephemera_folder_change_set_spec.rb
@@ -39,6 +39,16 @@ RSpec.describe EphemeraFolderChangeSet do
       result = change_set.validate(date_range_form_attributes: { start: "2018", end: "2017" })
       expect(result).to eq false
     end
+    it "is invalid if only start is given" do
+      change_set.prepopulate!
+      result = change_set.validate(date_range_form_attributes: { start: "2018", end: "" })
+      expect(result).to eq false
+    end
+    it "is invalid if only end is given" do
+      change_set.prepopulate!
+      result = change_set.validate(date_range_form_attributes: { start: "", end: "2018" })
+      expect(result).to eq false
+    end
     it "has a default" do
       change_set.prepopulate!
       expect(change_set.date_range_form.start).to be_nil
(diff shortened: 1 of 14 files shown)


spec/change_sets/ephemera_folder_change_set_spec.rb
```diff
@@ -22,6 +22,37 @@
...
+      change_set.prepopulate!
+      expect(change_set.date_range_form.start).to be_nil
+      expect(change_set.date_range_form.required?(:start)).to eq false
+    end
+    context "when there's a date range set" do
+      it "sets it single-valued appropriately" do
+        change_set = described_class.new(FactoryGirl.build(:ephemera_folder, date_range: DateRange.new(start: "2017", end: "2018")))
+        change_set.prepopulate!
+        expect(change_set.date_range_form.start).to eq "2017"
+      end
```
hackmastera: What behavior is expected if one of the values is given but the other is blank?

tpendragon: I figured it would be fine, but we should probably make it required.

Regarding `date_created`, I wonder if we should keep it around for the weird dates (thinks with question marks and such)

hackmastera: What is the indexing logic for date_created? If that is where single dates live it should be possible to search for them alongside date ranges.

tpendragon: It's just indexed the same as everything else - into a `date_tesim` field. I don't think you can search for it in Figgy, atm.

tpendragon: Maybe we should create a ticket for the old date field and ask the Ephemera people what they want?

hackmastera: New ticket sounds good!

tpendragon: Created #313 and added a new validation to make sure if start is set then end must be too and vice versa.

PR #303, opened 2017-09-29
Run derivatives if they're missing from Plum

### LINKED ISSUE #263: PlumImporter should generate derivatives if it can't find a derivative file

tpendragon:
Closes #263


tpendragon:
Given that some derivatives legitimately can't generate with this PR, I also added some commits to
1. Let the manifest still render in UV
2. Make the catalog page not break if it's set to a thumbnail without a derivative.

app/services/plum_importer.rb
```diff
@@ -31,6 +31,12 @@ def generate_derivative(member)
       handlers: change_set_persister.handlers
     )
     derivative_change_set_persister.save(change_set: file_change_set)
+  rescue Errno::ENOENT
```
hackmastera: It's a little confusing that this method is called generate_derivative but I think from conversation that it actually imports / saves the derivative. It looks like it is saving from a file location; are these applications running on the same box?

hackmastera:
If there are derivatives that still aren't generating, should those be logged somewhere for future attention?

app/services/plum_importer.rb
tpendragon: :+1: I'll fix the name.

tpendragon:
@hackmastera Take another look?

Issue #257, opened 2017-09-27
Vocabulary Terms should serialize RDF


tpendragon:

tpendragon:
JSONLD for Vocabulary: https://plum.princeton.edu/vocabularies/22.jsonld
JSONLD for Term: https://plum.princeton.edu/vocabulary_terms/693.jsonld
JSONLD for Category: https://plum.princeton.edu/vocabularies/2.jsonld

Issue #226, opened 2017-09-22
EphemeraVocabulary resource URI's should default to a plum.princeton.edu NS


jrgriffiniii:
Discovered while discussing #197 , `EphemeraVocabulary` resources should generate a URI using `https://plum.princeton.edu/ns` as a namespace when not supplied by the user.

jrgriffiniii:
It was clarified by @tpendragon that all terms should bear the `https://plum.princeton.edu/ns` namespace, and that a `skos:exactMatch` predicate [1] shall link these to any additional URI provided by the user.

[1]: https://www.w3.org/TR/skos-reference/#mapping-cycles-exactMatch

PR #310, opened 2017-10-02
Vocab URI's for EphemeraFolders

### LINKED ISSUE #243: JSONLD of Ephemera Folders should have Vocabulary URIs.

jrgriffiniii:
Resolves #243, #257, and #226


jrgriffiniii:
I don't see how this can be fully resolved without also ensuring that #226 is resolved as well

app/services/linked_resource_builder.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+class LinkedResourceBuilder
```
tpendragon: What's the purpose of this abstraction? Wouldn't it be less code to just `LinkedResourceFactory.new(resource: resource).to_json, rather than `LinkedResourceBuilder.new(resource: resource).build`?

jrgriffiniii: Certainly, yes.  I believe that my reasoning here was that the `LinkedResourceBuilder` remain structured along consistent lines with that of the `ManifestBuilder`.  Both model a service which transforms a Valkyrie resource into an RDF graph (rather than simple serialization to JSON), and, perhaps, there might be some interfaces which could be generalized in a future sprint.

PR #323, opened 2017-10-05
Implementing a HumanReadableTypeIndexer for indexing Multi-Volume Works

### LINKED ISSUE #254: Works with children should index a different type
So we can facet on Multi Volumed Works

jrgriffiniii:
Resolves #254


tpendragon:
Oh! Using human readable type is really smart! Very cool. Merging.

PR #334, opened 2017-10-09
Ensures that local identifiers are indexed

### LINKED ISSUE #304: Ensure search by local identifier works.

jrgriffiniii:
Resolves #304


spec/controllers/catalog_controller_spec.rb
```diff
@@ -35,6 +35,20 @@
       expect(facets["state_ssim"]).to eq ["complete", 1]
       expect(facets["display_subject_ssim"]).to eq [output.imported_metadata[0].subject.first, 1]
     end
+    it "can search by local identifiers" do
+      persister.save(resource: FactoryGirl.create_for_repository(:scanned_resource, local_identifier: "p3b593k91p"))
+
+      get :index, params: { q: "p3b593k91p" }
+      expect(assigns(:document_list).length).to eq 1
+    end
+    context 'with imported metadata' do
```
tpendragon: I don't think there's such a thing as an imported local identifier. @escowles ?

escowles: We're not mapping it now, but I think that records with local identifiers in 024 (like the Cicognara records) should have those imported into local_identifier.

tpendragon: Okay. This test still isn't right though - you need something like `FactoryGirl.create_for_repository(:scanned_resource, imported_metadata: [ImportedMetadata.new(local_identifier: "p3b593k91p")])`

jrgriffiniii: Understood, I shall proceeding by using a fixture from 8543429

jrgriffiniii:
This is ready for another review

PR #341, opened 2017-10-09
Migrating geographic_origin

escowles:


app/change_sets/ephemera_folder_change_set.rb
```diff
@@ -20,7 +20,7 @@ class EphemeraFolderChangeSet < Valhalla::ChangeSet
   property :creator, multiple: false, required: false
   property :contributor, multiple: true, required: false
   property :publisher, multiple: true, required: false
-  property :geographic_origin, multiple: false, required: false, type: Valkyrie::Types::ID
+  property :geographic_origin, multiple: false, required: false
```
tpendragon: This is interesting. I'd expect this to make it so when it saves it saves it as a string and not a Valkyrie::ID. Can you add a test below to make sure it's an instance of `Valkyrie::ID`?

escowles: I've updated the test to check all the values that should be IDs, and they are.  When the `type: Valkyrie::ID` is included, they get messed up (I see objects that have an `id` property with the `Valkyrie::ID` value).  Maybe an object that is already a `Valkyrie::ID` gets coerced when it shouldn't be?

tpendragon: That's the json serialization of a Valkyrie ID. I wonder if this is a Valkyrie bug.

PR #358, opened 2017-10-11
Extending the BulkIngestService and implements the IngestDiskFolderJob for bulk directory ingestion

### LINKED ISSUE #56: Migrate over bulk directory ingest script

jrgriffiniii:
Resolves #56 by:
- Restructuring the BulkIngestService to properly handle ScannedMap Resources and ingestion for Multi-Volume Works
- Restructuring the BulkIngestService in order to ensure that Multi-Volume Works are ingested with child resources of ScannedResources while EphemeraFolders have ScannedResources ingested as children


jrgriffiniii:
I suspect that there are various areas of improvement, but I'm eager to capture feedback for this regarding the general approach.

lib/tasks/bulk.rake
```diff
@@ -0,0 +1,102 @@
...
+    @logger.warn "No BIB id specified" unless bib
+    @logger.info "ingesting files from: #{dir}"
+    @logger.info "ingesting as: #{user.user_key} (override with USER=foo)"
+    @logger.info "adding item to collection #{coll}" if coll
+
+    begin
+      if background
+        IngestDiskFolderJob.perform_later(directory: dir, collection: coll, source_metadata_identifier: bib, local_identifier: local_id, replaces: replaces)
+      else
+        change_set_persister = PlumChangeSetPersister.new(
```
escowles: Maybe just do `perform_now` to avoid duplicating the job logic here?

app/services/bulk_ingest_service.rb
```diff
@@ -3,51 +3,136 @@ class BulkIngestService
...
+      attach_dir_for_new(directory_path: base_directory_path, file_filter: file_filter, **attributes)
+    end
+  end
 
-    logger.info "Found #{resource.id} for #{property}:#{property_value}"
-    attach_files(base_directory: base_directory, resource: resource, file_filter: file_filter)
+  def attach_dir_for_existing(directory_path:, property:, value:, file_filter:)
+    root_resource = find_resource_by(property: property, value: value)
+    return unless root_resource
+    attach_files(path: directory_path, resource: root_resource, file_filter: file_filter)
```
escowles: I agree that this looks like it could be refactored to be simpler.  One possibility would be using `attach_children` here.  We currently only have a single-volume use case for attaching ephemera images to existing objects, but I can imagine using a similar workflow for multi-volume items if they were scanned by a vendor.

If that was the case, then you could just have the logic to find an existing resource or fallback on creating a new resource.

jrgriffiniii:
This has been restructured and is prepared for a second review.

lib/tasks/bulk.rake
escowles: I meant that lines 30-37 could just be `IngestDiskFolderJob.perform_now` with the same arguments as the `perform_later` call above.  Line 28 should still be `perform_later` since `background` is true.

jrgriffiniii:
Further refactored and prepared for another review.

PR #359, opened 2017-10-12
Fix ephemera display and data migration issues

escowles:


app/decorators/ephemera_folder_decorator.rb
```diff
@@ -12,21 +12,23 @@ class EphemeraFolderDecorator < Valkyrie::ResourceDecorator
...
     :description,
     :date_created,
     :rendered_date_range,
     :dspace_url,
     :source_url,
-    :visibility
+    :visibility,
+    :rendered_rights_statement
   ]
+  self.suppressed_attributes -= [:description]
```
hackmastera: Interesting. Do we know why it was in this list to begin with? Maybe it is supposed to be somewhere else on the page layout? But considering plum/figgy isn't really a user-facing application maybe that's not important? Should we just remove it from the list altogether?

escowles: I'm not sure — it looks like it was added to the suppressed list here: https://github.com/pulibrary/figgy/commit/2c34528cbc251bc928058b19f41fb28ee3c060b3#diff-e9e53c864d6bc24d5ceb99e0ebdb8357 — but I don't see anything removed in that commit that would have suppressed it.

Absent any reason to suppress it, I'm going to remove it from the global `suppressed_attributes` list and we can suppress it on other decorators if a use case presents itself.

hackmastera: 👍

PR #362, opened 2017-10-12
Add a factory to MultiChecksum, use it from plum importer; fixes #151

### LINKED ISSUE #151: Plum ingests should do checksums.

hackmastera:


app/models/multi_checksum.rb
```diff
@@ -3,4 +3,13 @@ class MultiChecksum < Valkyrie::Resource
   attribute :sha256, Valkyrie::Types::SingleValuedString
   attribute :md5, Valkyrie::Types::SingleValuedString
   attribute :sha1, Valkyrie::Types::SingleValuedString
+
+  def self.factory(file_object)
```
tpendragon: We've used `.for` for the Factory Method pattern before. I don't really care which we use, but it should match https://github.com/pulibrary/figgy/blob/master/app/models/file_metadata.rb#L15

hackmastera: i like `.for`; sounds good.

app/services/plum_importer.rb
```diff
@@ -16,11 +16,20 @@ def import!
...
     file_sets.each do |member|
       id_cache[member.local_identifier.first] = member.id.to_s
+      add_checksums(member)
       import_derivative(member)
     end
     update_structure(output, members, change_set_persister)
   end
 
+  def add_checksums(member)
+    member.original_file.file_identifiers.map do |id|
```
hackmastera: oops don't need map *and* append; changed logic midstream here; will fix.

Issue #374, opened 2017-10-13
Add a property for provenance


jrgriffiniii:
As discussed within the meeting on 10/13/17, a stakeholder requested that an additional property for Ephemera resources in order to track provenance.

hackmastera:
cardinality? I would assume it can be multiple but should we confirm?

escowles:
Given the use case, I think this should be singular — if there's some complicated situation where there is more than one donor/contributor, then I would expect that to be reflected in the text of a single value, rather than having a list of donors.

escowles:
Closed with #382

PR #382, opened 2017-10-13
Add provenance attribute to ephemera folders, refs #374

hackmastera:
Indexing piece to follow


hackmastera:
apparently indexing is not actually required, in which case resolves #374

PR #384, opened 2017-10-13
Use Valkyrie's custom query functionality

tpendragon:


hackmastera:
Looks good; instead of having to instantiate and use a separate query object you can go through the existing query service.

hackmastera:
Thanks for calling that out, @jrgriffiniii

PR #385, opened 2017-10-16
Ensures that Ephemera Projects have IIIF Manifests

### LINKED ISSUE #364: EphemeraProjects should generate IIIF collection manifests

jrgriffiniii:
Resolves #364 by:
- [x] Ensuring that Ephemera Projects have IIIF Manifests (exposing Boxes)
- [x] Ensures that Ephemera Projects have slugs generated for them
- [x] Ensures that these slugs are published using RabbitMQ when EphemeraBoxes and EphemeraFolders are updated


app/change_sets/ephemera_project_change_set.rb
```diff
@@ -3,7 +3,7 @@ class EphemeraProjectChangeSet < Valkyrie::ChangeSet
   validates :title, presence: true
   property :title, multiple: false
   property :member_ids, multiple: true, required: false, type: Types::Strict::Array.member(Valkyrie::Types::ID)
-  property :slug, multiple: false, required: false
+  property :slug, multiple: false, required: true
```
tpendragon: This makes it required in the form, but there should be a validation too.

app/assets/javascripts/ephemera_project_slug.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Ephemera Project slug generation
+ * Generata slug identifier when the client enters or modifies the title for a project
+ */
+
+// UUID Generation without Bower, AMD's, or the NPM
+function uuidv4() {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
```
hackmastera: does this become the ID of project, somehow?

jrgriffiniii: Thank you for the review, documentation should indeed be added in order to clarify this feature.

> Shouldn't there be a routing piece? What happens if a slug changes? Is there some 'preferred' or 'primary' slug [...]

This does not become the ID of the EphemeraProject.  As the slug is entered by the user prior to ingestion (when the ID is assigned), this is simply a UUID-based alternative for generating unique slugs for EphemeraProject Objects.

A unique slug is required for the purposes of publishing messages via the RabbitMQ, but I don't believe that users ever actually reference or otherwise use this slug (from what I've gathered this is used for synchronization with other services such as https://github.com/pulibrary/lae-blacklight).  Hence, I don't believe that routing or redirection should be offered.  Please correct me if I am incorrect @tpendragon or @escowles.

Unless there is a preference for users to generate the slug using their own scheme (i. e. if they happened to track these slugs for purposes of which I'm unaware), then I would think that the automatic generation using JavaScript is entirely unwarranted.

escowles: The slug is the user-readable shortname for the project/collection — e.g., for use in DPUL, where the collection URL is dpul.princeton.edu/[slug].  So it should be entered by the user, and shouldn't be auto-generated or have any suffixes, etc. to make sure it's unique.  If there is another collection or ephemera project with the slug, an we should just display an error message to the user.

hackmastera: Ah, okay, my understanding was that this would be the basis of a distributable URL to aid in replacing standalone sites like the one for lae with easy-to-access collection pages within figgy/plum. Was I mistaken here? @escowles? Totally possible, considering I didn't know about the messaging use case.

escowles: @hackmastera I think that's right, except the collection pages are in DPUL, not figgy/plum.  The point of the slug property is for plum/figgy to tell DPUL what it should use for the collection shortname for building its URLs.

hackmastera: Okay, but we don't need routing because it's not really for use in figgy / plum -- it's just for use in DPUL.

jrgriffiniii:
This has been updated and is prepared for another review.

hackmastera:
While you're adding a validation for uniqueness I would also like to see a validation for suitability in a url. like `validate :slug_characters` or something that ensures it's only lowercase alphanumeric plus dashes.

app/decorators/ephemera_project_decorator.rb
```diff
@@ -25,4 +29,20 @@ def manageable_structure?
   def title
     super.first
   end
+
+  def slug
+    Array.wrap(super).first
```
hackmastera: If we want it to be single-value, we should enforce that on the changeset. But I'm not sure we want it to be single-value. I'm concerned we actually need 2 properties, like current_slug and all_slugs, so that the front-end app can continue routing to old / bookmarked urls. I think this may need more discussion / a new issue.

hackmastera: Oh, it already is enforced on the changeset, okay. Well the rest of my concern still stands :)

escowles: I've created a new issue for handling changed slugs gracefully — not sure if it's ultimately something that needs to change in DPUL or Figgy, but it's something that should work: https://github.com/pulibrary/figgy/issues/395

jrgriffiniii:
Slug validation has been improved upon as requested and this is prepared for another review.

hackmastera:
Looks good to me!

PR #389, opened 2017-10-16
Adding code documentation for ControlledVocabulary and EphemeraField

### LINKED ISSUE #388: Improve upon the documentation for ControlledVocabulary and EphemeraVocabulary Models
Currently, ambiguity in naming and nested Class definitions render it difficult to modify or improve upon the functionality for the `ControlledVocabulary` and `EphemeraVocabulary` Classes (and related `Term`'s).

Code documentation, at minimum, should be offered for these.

jrgriffiniii:
Resolves #388

### DIFF
--- a/app/models/controlled_vocabulary.rb
+++ b/app/models/controlled_vocabulary.rb
@@ -1,48 +1,79 @@
 # frozen_string_literal: true
+
+# Base Class for controlled vocabularies for local authorities
+# Terms within these vocabularies are stored within YAML files
+# *Please* note that these are distinct from EphemeraVocabulary and Vocab Classes
 class ControlledVocabulary
   class_attribute :handlers
   self.handlers = {}
+
+  # Mutates the Class attribute mapping a key symbol to a ControlledVocabulary subclass
+  # @param key [Symbol] the symbolized name of the registered vocabulary
+  # @param klass [Class] the ControlledVocabulary subclass
   def self.register(key, klass)
     handlers[key] = klass
   end
 
+  # Accesses the Class mapped to a key symbol for a ControlledVocabulary subclass
+  # @param key [Symbol] the symbolized name for the registered vocabulary
+  # @return [Class] the ControlledVocabulary subclass
   def self.for(key)
     (handlers[key] || self).new(key: key)
   end
 
+  # Initializer for the ControlledVocabulary
   attr_reader :key
+  # @param key [Symbol] the symbolized name for the vocabulary
   def initialize(key: nil)
     @key = key
   end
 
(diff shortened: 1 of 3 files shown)


app/models/controlled_vocabulary.rb
```diff
@@ -1,75 +1,123 @@
 # frozen_string_literal: true
+
+# Base Class for controlled vocabularies for local authorities
+# Terms within these vocabularies are stored within YAML files
+# *Please* note that these are distinct from EphemeraVocabulary and Vocab Classes
```
hackmastera: Where is the "Vocab" class? cannot find.

app/models/controlled_vocabulary.rb
```diff
@@ -1,75 +1,123 @@
...
+  # Base Class for terms within a ControlledVocabulary
+  # Please note that this Class does *not* contain a property for term ownership
   class Term < Valkyrie::Resource
     attribute :label, Valkyrie::Types::String
     attribute :value, Valkyrie::Types::Any
     attribute :notable, Valkyrie::Types::Bool
     attribute :definition, Valkyrie::Types::String
+
+    # Whether or not this Term has been marked as notable
+    # @return [Boolean] whether or not this Term is notable
```
hackmastera: what does it mean to be notable?

app/models/ephemera_field.rb
```diff
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
+# Class for EphemeraFields as independent resources
+# Please note that these are global in namespace and distinct from ControlledVocabulary::EphemeraField
+# Also note that :field_name was used (as opposed to :name) for the purposes of disambiguation
```
hackmastera: It's actually the `name` part of this variable that I find misleading. Since it's effectively a number, but `name` to me implies a word. maybe `key` or `field_key` would be better.

app/models/controlled_vocabulary.rb
escowles: Whether or not a general note/description field should be shown with this term.  This is used in Plum to trigger the display of a rights note for some rights statements but not others.

hackmastera: okay, but the note itself is fixed / not editable? And I guess these are pulled from locales.

escowles: In Plum, there is a rights note field that is hidden or displayed on the edit form based on which rights statement is selected.  So if you select a rights statement that is "notable", the rights note input box is shown so the user can enter a rights note, and if you select a rights statement that isn't "notable", the note field is cleared and hidden.

PR #397, opened 2017-10-17
Allow adding folders directly to projects

### LINKED ISSUE #378: Ephemera without boxes
We should be able to add ephemera items without being in a box.

These should be accessible in the browse in a "Boxless Items" pseudo-box, and the form should not require a barcode.

escowles:
Fixes #378

### DIFF
--- a/app/change_sets/boxless_ephemera_folder_change_set.rb
+++ b/app/change_sets/boxless_ephemera_folder_change_set.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+class BoxlessEphemeraFolderChangeSet < EphemeraFolderChangeSet
+  def self.new(record, *args)
+    return DynamicChangeSet.new(record, *args) unless record.is_a?(EphemeraFolder)
+    super
+  end
+
+  validates :title, :language, :genre, :width, :height, :page_count, :visibility, :rights_statement, presence: true
+  property :barcode, multiple: false, required: false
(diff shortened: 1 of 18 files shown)


escowles:
As a side-effect, barcode and folder number are no longer required.  In theory, it would be nice to require them when in a box, but not require them when attached directly to the project.

tpendragon:
> In theory, it would be nice to require them when in a box, but not require them when attached directly to the project.

This should be easy - just use a different change set if it has a box.

hackmastera:
Title should maybe be "directly to folders"?

escowles:
I'm a little wary of creating different ChangeSets, if only because there are a ton of property defs and stuff (that could be abstracted out, I guess).

But could we just make the requirement dependent on the parent being a box, something like:

```
property :barcode, multiple: false, required: parent_box?

def parent_box?
  parent.is_a?(EphemeraBox)
end
```

tpendragon:
@escowles You could do that, but ChangeSets are really meant to be able to be attached to situations so you can solve these issues with polymorphism. If we're wary of having multiple change sets, I'd be inclined to figure out how we can solve that wariness.

escowles:
OK, I'll try out creating a separate ChangeSet class for when a folder is attached directly to a project, and see how it goes.

tpendragon:
This looks good to me, just needs rebased.

PR #412, opened 2017-10-18
Separate local from imported metadata properties for ScannedResources

### LINKED ISSUE #331: Improve Imported Metadata Display
Imported metadata should probably be in its own sub-section of metadata, without the "imported" label before the property name.

jrgriffiniii:
Resolves #331


app/decorators/application_decorator.rb
```diff
@@ -1,36 +1,97 @@
...
+    self.displayed_attributes += attribute_names
   end
 
-  def display_attributes
-    local_attributes(self.class.display_attributes - self.class.suppressed_attributes)
+  # Remove a set of attributes from display
+  # @param attribute_names [Symbol]
+  def self.suppress(attribute_names)
+    attribute_names = Array.wrap(attribute_names)
+    attribute_names.each { |attribute_name| self.displayed_attributes.delete(attribute_name) }
```
escowles: I think several of these could be shortened, e.g. couldn't this be:

```
self.displayed_attributes -= Array.wrap(attribute_names)
```

app/views/catalog/_resource_attributes_ephemera_folder.html.erb
```diff
@@ -0,0 +1,28 @@
+<div id="attributes">
```
escowles: Can we avoid some of these duplicate views?  Could we make one of them `_resource_attributes_default.html.erb`? Could we have one copy and then have this view and the other one call out to it with `render partial` ?

escowles:
I like the general approach here, and think this is a big improvement on making the logic for what to display clearer.  I left a couple of comments around streamlining a couple of things.

PR #424, opened 2017-10-19
Provide "Category -- Subject" links to faceted results, fixes #377"

### LINKED ISSUE #377: Subject categories are duplicated
If two distinct terms with distinct categories are selected for a given EphemeraFolder, the category of the term first selected is duplicated on the Show Page.

hackmastera:


app/decorators/ephemera_folder_decorator.rb
```diff
@@ -127,6 +127,20 @@ def language
     super.map { |value| controlled_value_for(value) }
   end
 
+  # Provide "Category -- Subject" with each linked to faceted search results
+  def rendered_subject
+    subject.map do |value|
+      value = value.decorate if value.is_a? EphemeraTerm
+      display = [
+        link_to_facet_search(field: 'display_subject_ssim', value: value.vocabulary.label),
```
escowles: I think you could simplify this by doing the first `link_to_facet_search` call, and then returning it `if value.label == value.vocab.label`.

hackmastera: lmk how that looks to you!

PR #427, opened 2017-10-19
Project Manifests should show Folders as members.

### LINKED ISSUE #422: Project Manifests should have IIIF Folders as members, not boxes

### LINKED ISSUE #423: Project Manifests should not display EphemeraTerms as members.

tpendragon:
Closes #423, closes #422.

### DIFF
--- a/app/models/concerns/linked_data/linked_node.rb
+++ b/app/models/concerns/linked_data/linked_node.rb
@@ -25,7 +25,7 @@ module LinkedData
         resource.label
       else
         {
-          "@id" => resource.h.polymorphic_url(resource),
+          "@id" => helper.url_for(resource),
           "@type" => "skos:Concept",
           "pref_label" => resource.label,
           "exact_match" => {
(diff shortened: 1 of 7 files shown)


app/queries/find_grandchildren.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  attr_reader :query_service
+  delegate :resource_factory, to: :query_service
+  delegate :orm_class, to: :resource_factory
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def find_grandchildren(resource:)
+    query_service.find_members(resource: resource).flat_map do |child|
```
escowles: With https://github.com/pulibrary/figgy/pull/397, the members could either be folders or boxes.  So we should probably filter the members to make sure we're only getting the members of boxes (and not getting FileSets too).  And then we'll want to get the directly-attached folders too.

tpendragon: Oh, yeah, okay.

hackmastera: sounds like `find_grandchildren` is no longer the correct method name.

escowles: `find_children_and_grandchildren_of_type` doesn't have the same ring to it

PR #435, opened 2017-10-20
Make sure we're working with a string when checking text direction

escowles:


app/helpers/catalog_helper.rb
```diff
@@ -9,6 +9,10 @@ def render_document_heading(*args)
     document ||= @document
 
     val = presenter(document).heading
-    content_tag(tag, val, itemprop: "name", dir: val.dir)
+    content_tag(tag, val, itemprop: "name", dir: dir(val))
```
tpendragon: Why not just val.to_s.dir always?

escowles: Good point, the string `.to_s` implementation probably doesn't add any overhead...

PR #443, opened 2017-10-20
Restore rendered_subject to eph folder decorator

hackmastera:
This was functionality introduced in #424 that regressed in #412


hackmastera:
@escowles thanks for the assist :)

PR #444, opened 2017-10-20
Fix in_scheme missing from JSON-LD output.

tpendragon:


spec/models/linked_data/linked_ephemera_folder_spec.rb
```diff
@@ -96,14 +96,22 @@
...
-          "@id" => "http://www.example.com/concern/ephemera_terms/#{ephemera_term.id}",
+          "@id" => "http://www.example.com/concern/ephemera_terms/#{ephemera_child_term.id}",
           "@type" => "skos:Concept",
-          "pref_label" => ephemera_term.label,
-          "exact_match" => { "@id" => ephemera_term.uri.first }
+          "pref_label" => ephemera_child_term.label,
+          "exact_match" => { "@id" => ephemera_child_term.uri.first },
+          "in_scheme" => {
+            "@id" => "http://www.example.com/concern/ephemera_vocabularies/#{ephemera_vocabulary.id}",
+            "@type" => "skos:Concept",
```
escowles: Shouldn't this be a `skos:ConceptScheme`?

tpendragon: Probably, fixing.

PR #464, opened 2017-10-24
Don't render date range in JSON-LD if empty.

tpendragon:


app/models/concerns/linked_data/linked_date_range.rb
```diff
@@ -3,6 +3,7 @@
 module LinkedData
   class LinkedDateRange < LinkedResource
     def local_fields
+      return {} if resource.start.blank? || resource.end.blank?
```
escowles: Maybe this should be `&&` instead of `||` — if we had start=1900 and end=blank, wouldn't we want to display `1900-` ?

tpendragon: Honestly, it should be impossible for a date range to get in there without start/end set. It's invalid (per a discussion we had before.) I'm not sure where they come from.

escowles: OK, I'm fine with this as-is if we agreed that start and end are both required.

PR #476, opened 2017-10-25
Report for records with ARKs not present in Voyager metadata

### LINKED ISSUE #465: Report of objects with ARKs where the ARK isn't in the MARC record
Produce a CSV report listing the bibid and ARK for all objects that have an ARK, but the ARK is not present in the imported metadata.

escowles:
Fixes #465

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+class ReportsController < ApplicationController
+  def identifiers_to_reconcile
+    authorize! :show, Report
+    @resources = find_identifiers_to_reconcile_query.find_identifiers_to_reconcile
+    respond_to do |format|
+      format.html
+      format.csv do
+        send_data to_csv(@resources, fields: { source_metadata_identifier: 'bibid', identifier: 'ark' }),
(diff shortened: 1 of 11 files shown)


app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      CSV.generate(headers: true) do |csv|
+        csv << fields.map { |_k, v| v }
+        records.each do |record|
+          csv << fields.map { |k, _v| Array.wrap(record.send(k)).first }
+        end
+      end
+    end
+
+    def find_identifiers_to_reconcile_query
+      FindIdentifiersToReconcile.new(query_service: Valkyrie.config.metadata_adapter.query_service)
```
tpendragon: Should we register this as a custom query so we can do `query_service.custom_queries.find_identifiers_to_reconcile`?

escowles: I added it to the initializer (https://github.com/pulibrary/figgy/pull/476/files#diff-2aaca0f4b98ee23805a01ecacbd74a07R113) — is there anything else we need to do?

tpendragon: Nope. Just use it via custom_queries instead of the class

Issue #487, opened 2017-10-27
Ensure that `DeleteMembers` properly handles FileSets and Collections


jrgriffiniii:
Please see https://github.com/pulibrary/figgy/pull/483/commits/9859c656ef0cda15663b16f2f70b0732fb915488#diff-a0fc55c9f78aa928eb67dc80beee3d80R19

jrgriffiniii:
Resolved by #501

PR #501, opened 2017-10-30
Deletes member resources for Collections, FileSets and general Resources

### LINKED ISSUE #484: Is `post_save_resource` a useful argument for `DeletedReferenced`
This was discovered after a code review by @hackmastera, and I question whether it is unnecessary.  Please see:

https://github.com/pulibrary/figgy/pull/483
 https://github.com/pulibrary/figgy/blob/master/app/change_set_persisters/plum_change_set_persister.rb#L136
 https://github.com/pulibrary/figgy/blob/master/app/change_set_persisters/plum_change_set_persister/delete_referenced.rb#L10

jrgriffiniii:
Resolves #484 and #487

### DIFF
--- a/spec/change_set_persisters/plum_change_set_persister_spec.rb
+++ b/spec/change_set_persisters/plum_change_set_persister_spec.rb
@@ -432,15 +432,14 @@ RSpec.describe PlumChangeSetPersister do
 
   describe "collection interactions" do
     context "when a collection is deleted" do
-      it "cleans up associations from all its members" do
+      it "deletes members" do
         collection = FactoryGirl.create_for_repository(:collection)
         resource = FactoryGirl.create_for_repository(:scanned_resource, member_of_collection_ids: collection.id)
         change_set = CollectionChangeSet.new(collection)
 
         change_set_persister.delete(change_set: change_set)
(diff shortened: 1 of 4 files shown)


spec/change_set_persisters/plum_change_set_persister_spec.rb
```diff
@@ -432,15 +432,14 @@
...
+      it "deletes members" do
         collection = FactoryGirl.create_for_repository(:collection)
         resource = FactoryGirl.create_for_repository(:scanned_resource, member_of_collection_ids: collection.id)
         change_set = CollectionChangeSet.new(collection)
 
         change_set_persister.delete(change_set: change_set)
-        reloaded = query_service.find_by(id: resource.id)
 
-        expect(reloaded.member_of_collection_ids).to eq []
+        expect { query_service.find_by(id: resource.id) }.to raise_error(Valkyrie::Persistence::ObjectNotFoundError)
```
escowles: I don't think we want to delete objects that link to a collection when the collection is deleted.  I think the old behavior is what we should do (the collection id should be removed from the member_of_collection_ids property, but the object should remain).

PR #506, opened 2017-10-31
Add top_language property to ephemera projects

### LINKED ISSUE #490: Put Spanish, Portuguese, and English at top of language select menu

hackmastera:
fixes #490

### DIFF
--- a/app/helpers/edit_field_helper.rb
+++ b/app/helpers/edit_field_helper.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: false
 module EditFieldHelper
-  def reorder_languages(languages)
-    front_languages = %w[English Portuguese Spanish]
-    pull_to_front(languages) { |term| front_languages.include? term.label }
+  def reorder_languages(languages, top_languages)
+    pull_to_front(languages) { |term| top_languages.include? term }
   end
 
   private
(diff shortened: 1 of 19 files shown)


app/helpers/edit_field_helper.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: false
+module EditFieldHelper
+  def reorder_languages(languages)
+    front_languages = %w[English Portuguese Spanish]
```
escowles: I expect we'll have a similar request from other ephemera projects who want other languages at the top of the list.  Can we move the list of languages to the ephemera project?

hackmastera: You want a new field on the project and a UI for setting these?

escowles: I'm open to other ways of configuring this if there's something easier, but I definitely expect that different ephemera projects will want different languages to be at the top of the list.

hackmastera: Okay. Letting them put in what they want is probably better than making them ask us. The whole thing is strange to me since I am keyboard-oriented and would just start typing :)

hackmastera:
ready for review

Issue #509, opened 2017-10-31
Save and create another not working


escowles:
When clicking "Save and Create Another", it goes to the show page, so it looks like it's just doing a normal save.

escowles:
I realized the problem: this works when creating a new record, because there is an `after_create_response` method that gets overridden in the EphemeraFoldersController.  But there is no corresponding `after_update_response` to override after editing and saving an existing record.

PR #511, opened 2017-10-31
Extending the EphemeraFolder feature tests for saving and creating subsequent folders

jrgriffiniii:
Ensures that #509 is resolved

### DIFF
--- a/spec/features/ephemera_folder_spec.rb
+++ b/spec/features/ephemera_folder_spec.rb
@@ -23,6 +23,90 @@ RSpec.feature "Ephemera Folders", js: true do
     ephemera_box
   end
 
+  context 'within an existing ephemera project' do
+    scenario 'users can save a new folder' do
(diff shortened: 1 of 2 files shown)


jrgriffiniii:
Please note that this does *not* actually resolve #509, as I could not reproduce this on the `master` branch.  My suspicion is that this was quite recently resolved by https://github.com/pulibrary/figgy/pull/507

escowles:
I just confirmed the problem with Save and Create Another in production just now.  I'm going to deploy the current `master` branch to see if it is resolved.

escowles:
Alas, after doing a deploy to production, I'm still seeing the same error — doing "Save and Create Another" behaves the same as "Save"

jrgriffiniii:
Apologies, I'll need to return to attempting to reproduce this, as I'm getting the following within production:

issues_509_screenshot_prod

PR #512, opened 2017-10-31
Adding ephemera hard drive management

### LINKED ISSUE #495: Hard drive management
How/where do you attach a hard drive? Is there a "Manage Drives" button? The previous administrative system had "Manage Boxes," Manage Folders, "Manage Drives."

escowles:
* Adds a drive barcode field to ephemera boxes
* Displays drive barcode in box list
* Adds buttons for attaching and removing drives

Fixes #495

### DIFF
--- a/app/change_sets/ephemera_box_change_set.rb
+++ b/app/change_sets/ephemera_box_change_set.rb
@@ -9,6 +9,7 @@ class EphemeraBoxChangeSet < Valhalla::ChangeSet
   property :shipped_date, multiple: false, required: false
   property :received_date, multiple: false, required: false
   property :tracking_number, multiple: false, required: false
+  property :drive_barcode, multiple: false, required: false
   property :member_ids, multiple: true, required: false, type: Types::Strict::Array.member(Valkyrie::Types::ID)
   property :member_of_collection_ids, multiple: true, required: false, type: Types::Strict::Array.member(Valkyrie::Types::ID)
   # override the default value defined in VisibilityProperty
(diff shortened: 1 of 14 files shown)


escowles:
This is somewhat different from the PULStore implementation, since hard drives are not their own class.  But this has similar functionality, and generally folds drive management into box management.

app/change_sets/ephemera_box_change_set.rb
```diff
@@ -18,19 +19,26 @@ class EphemeraBoxChangeSet < Valhalla::ChangeSet
...
   validate :barcode_valid?
+  validate :drive_barcode_valid?
 
   def barcode_valid?
     return if Barcode.new(Array.wrap(barcode).first).valid?
     errors.add(:barcode, 'has an invalid checkdigit')
   end
 
+  def drive_barcode_valid?
+    return if drive_barcode.nil? || drive_barcode.empty? || Barcode.new(Array.wrap(drive_barcode).first).valid?
```
jrgriffiniii: Shouldn't `Barcode` return a `false` for `#valid?` if a `nil` or empty value is supplied to the constructor?

Please see https://github.com/pulibrary/figgy/blob/master/app/values/barcode.rb#L9

escowles: Yes, it would — but the nil and empty checks prevent the test from getting there.  This is to allow empty drive barcodes (since there's no requirement for a drive barcode to be supplied).

PR #528, opened 2017-11-03
Introduces JavaScript form functionality for warning users of duplicate source_metadata_identifiers and barcodes

### LINKED ISSUE #521: Detect and alert about duplicate barcodes
This should be pretty similar to the BibID duplication warning for ScannedResources

jrgriffiniii:
Resolves #521

### DIFF
--- a/spec/features/ephemera_folder_spec.rb
+++ b/spec/features/ephemera_folder_spec.rb
@@ -21,6 +21,7 @@ RSpec.feature "Ephemera Folders", js: true do
     sign_in user
     ephemera_project
     ephemera_box
+    ephemera_folder
   end
 
   context 'within an existing ephemera project' do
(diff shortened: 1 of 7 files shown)


spec/features/ephemera_folder_spec.rb
```diff
@@ -67,7 +68,7 @@
 
   context 'within an existing ephemera box' do
     scenario 'users can save a new folder' do
-      visit parent_new_ephemera_box_path(parent_id: ephemera_box.id)
+      visit parent_new_ephemera_box_path(parent_id: ephemera_folder.id)
```
escowles: Shouldn't the parent be an ephemera box here?  The context says "within an existing ephemera box".

spec/features/ephemera_folder_spec.rb
```diff
@@ -87,6 +88,17 @@
       expect(page).to have_content 'EphemeraFolder'
     end
 
+    scenario 'users see a warning if they try to use duplicate barcodes' do
+      visit parent_new_ephemera_box_path(parent_id: ephemera_folder.id)
```
escowles: Same as above: shouldn't the parent be an ephemera box?

PR #537, opened 2017-11-07
Display multi-volume-work volumes

### LINKED ISSUE #533: Diplay volumes on MVW show page

escowles:
Fixes #533


hackmastera:
@escowles I'm confused how this fixes the problem; can you describe a little bit?

escowles:
Renaming the partial to `_members_multi_volume_work.html.erb` fixes the problem, because Blacklight was looking for that instead of `_members_scanned_resource.html.erb`.

PR #539, opened 2017-11-07
Removing unique constraint on user email field

### LINKED ISSUE #460: Guest users for Figgy can't search
https://app.honeybadger.io/projects/53391/faults/35095579#notice-summary

escowles:
Fixes #460


tampakis:
Yay guest users can search now!

hackmastera:
@escowles I'm confused; are there not side effects of this change for registered user accounts?

escowles:
@hackmastera This change brings the index in line with the column definition: the email field wasn't declared unique (which is why AR can create multiple users with the same (empty) email), but then the unique constraint on the index would result in an error.

hackmastera:
@escowles thanks for the clarification!!

PR #543, opened 2017-11-10
Bulk jobs use low-priority queue

### LINKED ISSUE #541: Separate queues for batch operations
Uploads and other background jobs from interactive users shouldn't have to wait for large batch ingest jobs to finish.

escowles:
Fixes #541


hackmastera:
The queue has to be passed in to the persister because the persister kicks off a characterization job. All job invocations now use `.set` (an ActiveJob method) to apply or pass along the relevant queue. `queue_name` is an ActiveJob method that gives you the queue the current job is on.

This all looks good to me in terms of code. I assume we want to test it out on staging before deploying? I'll approve but wait on merge.

tpendragon:
Tested in staging - works like a charm.

PR #548, opened 2017-11-13
Implements an indexer for the EphemeraFolders for cases where the folders should be hidden

### LINKED ISSUE #494: Folders which are not "Complete" and whose box is not marked "all in production" should not show up in LAE front-end and should be unavailable to users without permission.
Visibility: It looks like all folders are "Open to the world." We can manually make them "Private" until the images arrive. After that, would it be possible to identify all the folders designated as private without having to look for them folder by folder? If not, is there a feature we could add?

jrgriffiniii:
Resolves #494 by implementing an indexer for the EphemeraFolders supporting the following:

- [x] Incomplete Folders (all EphemeraFolders in a state other than `complete`) within incomplete EphemeraBoxes (this assumes that a Box exists for the Folder, and that it is in a state other than `all_in_production`) are privately visible
- [x] Incomplete Folders without Boxes are `private`, but restored to their current visibility once in a state of `complete`
- [x] Incomplete Folders are not indexed for discovery within [pulibrary/lae-blacklight](https://github.com/pulibrary/lae-blacklight)

### DIFF
--- a/app/indexers/ephemera_folder_indexer.rb
+++ b/app/indexers/ephemera_folder_indexer.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+class EphemeraFolderIndexer
+  delegate :query_service, to: :metadata_adapter
+  attr_reader :resource
+  def initialize(resource:)
+    @resource = resource
+  end
+
+  def to_solr
+    return {} unless resource.is_a?(::EphemeraFolder)
+    {
+      Hydra.config[:permissions][:read].group => read_groups
+    }
+  end
+
+  private
+
+    def read_groups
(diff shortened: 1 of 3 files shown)


jrgriffiniii:
As I believe that `PaginatingJSONLDResponse` is used to access the JSON-LD for resources in Figgy by clients in Spotlight applications (i. e. [pulibrary/lae-blacklight](https://github.com/pulibrary/lae-blacklight)), these resources should not be accessible for indexing to such clients:

https://github.com/pulibrary/lae-blacklight/blob/de307df61ebaac488155c09dbe29c55b487187ba/app/indexers/reindexer.rb#L16

PR #558, opened 2017-11-16
WIP: Adding ImageMagick-based characterization

### LINKED ISSUE #556: Use identify instead of Tika?
I think we could get all of the properties we currently generate for TIFFs (width, height, file size, mime type) using ImageMagick's `identify` command.  During some informal testing, `identify` could extract these for TIFFs in < 1 sec, where Tika took 5+ seconds.

escowles:
Fixes #556

### DIFF
--- a/app/characterization_services/identify_file_characterization_service.rb
+++ b/app/characterization_services/identify_file_characterization_service.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# Class for ImageMagick identify based file characterization service
+class IdentifyFileCharacterizationService < TikaFileCharacterizationService
+  attr_reader :mime_types, :file_node, :persister
+
+  def initialize(file_node:, persister:)
+    @file_node = file_node
+    @persister = persister
+    @mime_types = {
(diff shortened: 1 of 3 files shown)


app/characterization_services/identify_file_characterization_service.rb
```diff
@@ -0,0 +1,26 @@
...
+
+# Class for ImageMagick identify based file characterization service
+class IdentifyFileCharacterizationService < TikaFileCharacterizationService
+  attr_reader :mime_types, :file_node, :persister
+
+  def initialize(file_node:, persister:)
+    @file_node = file_node
+    @persister = persister
+    @mime_types = {
+      "TIFF": "image/tiff"
```
escowles: This isn't great, but that is the only mime type we really support right now.  We could add others as we add support for them?

tpendragon: You can get a real mime type from the verbose output of identify. See #559.

escowles:
Closing in favor of https://github.com/pulibrary/figgy/pull/559

PR #559, opened 2017-11-16
Use ImageMagick to characterize.

### LINKED ISSUE #556: Use identify instead of Tika?

tpendragon:
Closes #556.

### DIFF
--- a/app/characterization_services/imagemagick_characterization_service.rb
+++ b/app/characterization_services/imagemagick_characterization_service.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+# Class for Apache Tika based file characterization service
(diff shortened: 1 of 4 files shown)


app/characterization_services/imagemagick_characterization_service.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+# Class for Apache Tika based file characterization service
```
hackmastera: Update comments referencing Tika

PR #566, opened 2017-11-16
Add comments to IngestableFile.

### LINKED ISSUE #500: Add comments to container_attributes and node_attributes
What types of attributes are expected in `container_attributes` and/vs `node_attributes`? I understood file node to be the metadata about a file. Seems like most of these `mime_type` `file_path` etc would go there. Some comments would be helpful https://github.com/pulibrary/figgy/blob/d1a2e6df11155c5abbcfa92c0e7a2c97d296535e/app/models/ingestable_file.rb#L7

tpendragon:
Closes #500.


tpendragon:
@hackmastera Can you take a look at this and see if it clears things up?

hackmastera:
yes. thank you!

PR #572, opened 2017-11-20
Add Benchmarking & Background Checksums

tpendragon:
Lets us look at the log to see how fast Valkyrie is performing, and speeds up PlumImport


escowles:
Looks good to me — my only question is whether we want to get `storage_adapter_logging` merged upstream before we merge this.

tpendragon:
I wonder if I should just make that a local thing like the `BenchmarkingMetadataAdapter`

PR #595, opened 2017-11-27
Vue filemanager

sdellis:
This is feature complete, but requires more work that's documented in github issues. Some feature tests are commented out for the moment while I work on Jest-based tests for these.


valhalla/app/javascript/packs/application.js
```diff
@@ -0,0 +1,10 @@
+/* eslint no-console:0 */
+// This file is automatically compiled by Webpack, along with any other files
+// present in this directory. You're encouraged to place your actual application logic in
+// a relevant structure within app/javascript and only use these pack files to reference
+// that code so it'll be compiled.
+//
+// To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
+// layout file, like app/views/layouts/application.html.erb
+
+console.log('Hello World from Webpacker')
```
eliotjordan: Is this line needed?

valhalla/app/javascript/packs/filemanager.js
```diff
@@ -0,0 +1,21 @@
+import 'babel-polyfill'
+import Vue from 'vue/dist/vue.esm'
+import App from '../app.vue'
+import store from '../store'
+
+document.addEventListener('DOMContentLoaded', () => {
+// document.addEventListener('turbolinks:load', () => {
```
eliotjordan: Looks like this can be deleted.

valhalla/app/javascript/store/index.js
```diff
@@ -0,0 +1,130 @@
...
+    state.selected = imgArray
+  },
+  SET_STATE (state, ImageCollection) {
+    state.id = ImageCollection.id
+    state.startPage = ImageCollection.startpage
+    state.thumbnail = ImageCollection.thumbnail
+    state.images = ImageCollection.images
+    state.ogImages = ImageCollection.images
+    // state.images = [ ...ImageCollection.images ]
+    // state.ogImages = [ ...ImageCollection.images ]
```
eliotjordan: Delete?

valhalla/app/javascript/test/unit/coverage/clover.xml
```diff
@@ -0,0 +1,143 @@
...
+        <line num="93" count="0" type="stmt"/>
+        <line num="94" count="0" type="stmt"/>
+        <line num="187" count="0" type="cond" truecount="0" falsecount="2"/>
+        <line num="188" count="0" type="cond" truecount="0" falsecount="2"/>
+        <line num="189" count="0" type="cond" truecount="0" falsecount="68"/>
+        <line num="190" count="0" type="cond" truecount="0" falsecount="2"/>
+      </file>
+    </metrics>
+  </project>
+</coverage>
```
eliotjordan: Everything in `valhalla/app/javascript/test/unit/coverage` looks like a coverage report for a run done on a local machine. If this is true, this directory should be deleted in the PR and added to .gitignore.

valhalla/app/javascript/test/unit/specs/sum.spec.js
```diff
@@ -0,0 +1,3 @@
+test('1 + 1 equals 2', () => {
+  expect(1 + 1).toBe(2);
+});
```
eliotjordan: Example spec?

Procfile
```diff
@@ -0,0 +1,3 @@
+# Procfile
+backend: bin/rails s -p 3000
+frontend: bin/webpack-dev-server
```
jrgriffiniii: Is this going to remain in place for the production deployment?  I only ask given the more NPM-bound approaches outlined using:
https://webpack.js.org/guides/production/#cli-alternatives

valhalla/app/javascript/app.vue
```diff
@@ -0,0 +1,96 @@
...
+  <div id="filemanager">
+    <gallery></gallery>
+    <detail></detail>
+    <panel></panel>
+    <controls></controls>
+  </div>
+</template>
+
+<script>
+// import {Gallery, Detail, Panel, Controls} from 'filemanager'
```
jrgriffiniii: Should this be removed?

valhalla/app/javascript/components/Panel.vue
```diff
@@ -0,0 +1,189 @@
+<template>
+  <div class="formPanel actions">
+    <p v-if="selectedTotal === 0" id="noneSelected" class="formContent">No items selected.</p>
+    <!-- Multiple Selected Form-->
+    <div v-if="selectedTotal > 1" id="multiSelected" >
+      <h2>Generate Labels <small>for selected items</small></h2>
+      <form class="formContent form-inline">
+        <!-- <div id="preview" class="row"><p class="text-muted">Example: <em>{{ example }}</em></p></div> -->
```
jrgriffiniii: Should this be removed also?

valhalla/app/javascript/mixins/manifesto-filemanager-mixins.js
```diff
@@ -0,0 +1,58 @@
...
+  mainSequence: function () {
+    // mainSequence is the one without an id (not ideal since it could have an id)
+    const s = this.getSequences()
+    const main_sequence = s.filter((seq) => seq.id !== 'undefined')
+    return main_sequence[0]
+  },
+
+  getCanvasMainThumb: function (canvas) {
+    const images = canvas.getImages()
+    var thumb = "http://via.placeholder.com/300x400"
```
jrgriffiniii: Should this be a protocol-relative URL specified here? (Then again, please see https://jeremywagner.me/blog/stop-using-the-protocol-relative-url/).  Also, could this be webpacked or compiled as a static asset (as in http://clarkdave.net/2015/01/how-to-use-webpack-with-rails/#including-precompiled-assets-in-views)?

README.md
```diff
@@ -38,7 +40,7 @@ bundle install
    - ``export SECRET_KEY_BASE=`rake secret` ``
    - `rake db:setup`
    - `rake figgy:development`
-   - In a separate terminal: `rails s`
+   - In a separate terminal: `foreman start`
```
escowles: I get `foreman: command not found` — is there more to add to the setup instructions?

config/deploy/staging.rb
```diff
@@ -9,6 +9,19 @@
 # server 'db.example.com', user: 'deploy', roles: %w{db}
 server 'figgy-staging1', user: 'deploy', roles: %w[app db web worker]
 
+before "deploy:assets:precompile", "deploy:npm_install"
```
escowles: Do we need to do this in production? If so, probably better to move this to `config/deploy.rb`.

config/routes.rb
```diff
@@ -4,6 +4,8 @@
   default_url_options Rails.application.config.action_mailer.default_url_options
   concern :exportable, Blacklight::Routes::Exportable.new
 
+  get 'filemanager/index'
```
escowles: What is this for? If I try to access this path, I get an error about `FilemanagerController` not being defined.

spec/controllers/ephemera_folders_controller_spec.rb
```diff
@@ -103,7 +103,7 @@
 
         get :manifest, params: { id: resource.id.to_s, format: :json }
 
-        expect(response).to be_not_found
+        expect(response).to be_forbidden
```
escowles: Is this change required? As a general rule, it is preferable to avoid giving information about resources a user is not allowed to access, including whether they exist or not.

valhalla/app/controllers/concerns/valhalla/resource_controller.rb
```diff
@@ -67,7 +67,10 @@ def update
         change_set_persister.buffer_into_index do |persist|
           obj = persist.save(change_set: @change_set)
         end
-        after_update_success(obj, @change_set)
+        respond_to do |format|
```
escowles: `after_update_success` already does the format negotiation.  I think it's better to have that happen in there instead of directly in the controller method.

valhalla/app/javascript/mixins/manifesto-filemanager-mixins.js
```diff
@@ -0,0 +1,80 @@
+const ManifestoFilemanagerMixins = {
+
+  mainSequence: function () {
+    // mainSequence is the one without an id (not ideal since it could have an id)
```
escowles: I'm not sure how to read this comment.  When would it have an id? Is this an unresolved issue where we need some other way to identify it?

valhalla/app/javascript/mixins/manifesto-filemanager-mixins.js
```diff
@@ -0,0 +1,80 @@
...
+    if (services.length) {
+      thumb = services[0].id + '/full/400,/0/default.jpg'
+    }
+    return thumb
+  },
+
+  getResourceId: function (canvas) {
+    const images = canvas.getImages()
+    const r = images[0].getResource()
+    // this seems to be a bug in manifesto ...
```
escowles: Can you open a ticket about this? It would be good to get some clarification on whether this is the expected behavior or not.

sdellis: It's actually a bug in our code. Tom Crane confirmed that it is an invalid manifest. I opened ticket #598. I will update this code comment to reflect that.

valhalla/app/javascript/mixins/manifesto-filemanager-mixins.js
sdellis: Sorry, this an old note to self. However, we need to determine which is the default sequence if there are multiple. The spec says it would be the first in the list, so I will just note that and remove the cruft.

Procfile
sdellis: Yes, this is a foreman config for developer convenience. Production uses the compiled code that doesn't require npm to start a separate webpack dev server.

valhalla/app/javascript/mixins/manifesto-filemanager-mixins.js
sdellis: Yes, created a separate issue for this: https://github.com/pulibrary/figgy/issues/613

spec/controllers/ephemera_folders_controller_spec.rb
sdellis: The response was returning a 403. I changed it to forbidden to get the test to pass.

valhalla/app/javascript/test/unit/coverage/lcov-report/Panel.vue.html
```diff
@@ -0,0 +1,632 @@
+<!doctype html>
```
tpendragon: We shouldn't have the coverage reports in the repository.

tpendragon:
Can we do a flash update or something instead of the javascript alert popups when something's saved?

README.md
sdellis: We added the foreman gem to the dev group in the Gemfile

sdellis:
@tpendragon Yes, I think that would be best. That was meant to be convenient and temporary. I we should also have a spinner while it's waiting. I'll add tickets for those.

tpendragon:
Drag/Drop reordering doesn't seem to be working ATM.

escowles:
Closing in favor of #700

PR #608, opened 2017-11-30
Introduces a SolrFacadeService for querying Solr for Collection Members and renders these on the View partial

### LINKED ISSUE #592: Improve collection display/navigation
- [x] Collections should display a paginated list of their members
 - [x] Collections on resource show pages should link to the collection show page

jrgriffiniii:
Resolves #592 by introducing a SolrFacadeService for querying Solr for Collection members and extending the Resource view partials for rendering these members (and linking to Collections from member Resource templates)


app/services/solr_facade_service.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
```
escowles: I'm not sure this service object adds that much to the embedded SolrFacade class.  Could be just have the SolrFacade object without this wrapper?

Issue #549, opened 2017-11-14
Cleanup stray files and filesets


escowles:
Due to ingest and deletion errors, we have a significant number of unattached FileSets, with TIFFs and JP2s on disk.  We should identify all files and FileSets that aren't attached to works and delete them.

Rake task or Thor command to check every file on disk and see if it's linked to by a resource. 

Iterate over every file, see if a fileset exists. If no, delete. `StorageAdapter` doesn't have `#all` so probably just use `Dir`.

Iterate over every file_set, delete if it has no parent.

An option for safety is to move all the files to a new location, then validate that all the objects have files before actually deleting.

This seems both low-priority and high-risk. Delay until we have a mirror of the data.

escowles:
Also, we're queueing up CleanupDerivativesJobs in PlumChangeSetPersister (https://github.com/pulibrary/figgy/blob/master/app/change_set_persisters/plum_change_set_persister.rb#L28), but then deleting the metadata it needs before it can execute.

We either need to delete derivatives inline, or we need to modify the derivatives cleanup to use file paths instead of using the metadata.

tpendragon:
Closing as I can't see us prioritizing this as a project.

PR #611, opened 2017-12-01
Handle errors when trying to cleanup derivatives for non-existent FileSets

escowles:


app/jobs/cleanup_derivatives_job.rb
```diff
@@ -13,6 +13,8 @@ class CleanupDerivativesJob < ApplicationJob
   def perform(file_set_id)
     file_set = query_service.find_by(id: Valkyrie::ID.new(file_set_id))
     Valkyrie::DerivativeService.for(FileSetChangeSet.new(file_set)).cleanup_derivatives
+  rescue Valkyrie::Persistence::ObjectNotFoundError
+    Rails.logger.error "Unable to find FileSet #{file_set_id}"
```
hackmastera: I think I would feel confused reading this error message. I can't tell how serious it is; does it mean we've lost data? Or does it just mean the files couldn't be deleted because they were gone already? Since there won't be a stack trace maybe some extra context would be helpful.

escowles: I think this is happening when we delete FileSets before the CleanupDerivativesJob has a chance to run — see https://github.com/pulibrary/figgy/issues/549#issuecomment-346938857

The ultimate solution to this problem is probably to cleanup derivatives and other files inline before the metadata is deleted.  Or maybe to make the CleanupDerivativesJob not rely on looking up the metadata (by passing the file path instead of the FileSet id?).

hackmastera: Cool, okay, maybe we could make it "Unable to find FileSet #{file_set_id} for deletion"? A little extra context would make me feel less nervous if I happened on this message in a log.

escowles:
@hackmastera I've updated the error message that gets logged.

PR #698, opened 2017-12-15
Multiple title support for DPUL

tpendragon:


app/services/manifest_builder.rb
```diff
@@ -47,7 +47,7 @@ def initialize(resource)
     # Stringify the Resource by delegating to the header within the Decorator
     # @return [String]
     def to_s
-      decorate.header
+      resource.title.map(&:to_s)
```
hackmastera: looks like the yard docs are now out of sync with the method here.

tpendragon: Yup! Fixed.

hackmastera:
When I looked at this not too long ago I didn't use the array because the iiif spec seemed to imply that label should be a string. Did I read the spec wrong? http://iiif.io/api/presentation/2.1/#label

escowles:
@hackmastera I'm seeing a lot of "at least one label" type of language that implies multiple labels are OK.

hackmastera:
Ah, great, thanks @escowles

PR #704, opened 2017-12-26
Update Webpacker/MathJS

### LINKED ISSUE #701: Upgrade mathjs
See vulnerability in https://github.com/pulibrary/figgy/network/dependencies

tpendragon:
Closes #701.


sdellis:
This looks good. The only thing I wonder about is that we should probably keep the Gem and the npm package versions of webpacker in sync. This branch still has 3.0.2 in the Gemfile: https://github.com/pulibrary/figgy/blob/mathjs/Gemfile#L24

It's kind of annoying that it has to be maintained in both package managers.

tpendragon:
@sdellis :+1: Fixed.

PR #761, opened 2018-01-24
Refactor schemas

### LINKED ISSUE #339: destroy schemas in favor of attribute lists per-model
the indirection is confusing; not clear what we gain

hackmastera:
This is my proposal to close #339. It's not drastic at all. It pretty much keeps the prior design, but combines a lot of the schemas and deletes others that were not in use. The reason I kept some schemas I did:
- common is used from decorators to delegate attributes to the imported metadata object.
- geo and iiif are all referenced from decorators to include / exclude their properties in various attribute lists.
- there are SO MANY marc relators and I think the usual use case would be to use them all at once; encapsulating them this way will reduce noise.

So in these cases there were compelling reasons to keep them separate / DRY. Please argue if you disagree.


hackmastera:
Re: the failing test, it looks like the hash of metadata values (see example below) may be in a different order. I hope order is not important, since it's a hash? I'm going to look at rewriting the spec to not be order-dependent, but please let me know if we're relying on the order for some ungodly reason.
```
[{"label"=>"Created At", "value"=>["01/24/18 02:59:48 PM UTC"]},
 {"label"=>"Updated At", "value"=>["01/24/18 02:59:49 PM UTC"]},
 {"label"=>"Actor", "value"=>["test person"]},
 {"label"=>"Label", "value"=>["test label"]},
 {"label"=>"Identifier", "value"=>["ark:/88435/abc1234de"]},
 {"label"=>"Title", "value"=>["test title1"]},
 {"label"=>"Description", "value"=>["Test Description"]},
 {"label"=>"Portion Note", "value"=>["test value1"]},
 {"label"=>"Edition", "value"=>["test edition"]},
 {"label"=>"Call Number", "value"=>["test value2"]},
 {"label"=>"PDF Type", "value"=>["gray"]}]
```

hackmastera:
oops it's an array of hashes so i guess it wouldn't be crazy for it to require a certain order I guess the reordering must have been a side effect of the way the attributes are ordered in the models now? that seems crazy?

PR #775, opened 2018-01-31
Make this test more specific so that other kinds of saves are still okay

hackmastera:


tpendragon:
Mentioned this in Slack, but putting it here too:

`save` is never meant to be called on the indexing persister here, because no indexing should happen until the very end of the process. I'm not sure why fixity breaks that - maybe it's an extra job that runs?

hackmastera:
I'm gonna try to get fixity out of solr.

Issue #142, opened 2017-08-29
Background fixity checking of files in Figgy


escowles:
We should have a background job to do fixity checking, so we could devote a worker to auditing checksums

hackmastera:
maybe create a 'super-low' priority queue instead of assigning a whole worker

hackmastera:
Question: What should happen if a fixity check fails? Should it email someone? If so is there a generic / alias email address that could be used? And/or should there be some kind of fixity event dashboard?

escowles:
👍 to a super-low priority queue for background fixity checking

I like the idea of a fixity dashboard, where you could go and see all failures, and check the status of any item.  Can we do that first, and then figure out if email or some other notification is needed?

hackmastera:
- a job that runs a single fixity check
  - how is this job triggered?
  - a job that loops through all files
    - then starts over again
    - then stops; is started again by cron
      - what happens if/when the repo gets large enough that the first job is still running when the next one starts?
    - How often does any given file need to be checked? what is our backup schedule like?
- a job that runs a fixity check and then kicks off a new instance of itself on the next object
  - how does it know where to start its fixity check?
    - run through an enumerator and check against fixity object table for either
      - this object has never been run
      - this object has not been run in X days
  - how does it know what the next object is?
    - probably just does the above again
  - this design is simple and idempotent. If the job dies it can just be started up again. Could we even have the fixity dashboard look for the job and if it's not found, provide a button to start it?
- fixity object
  - file id
  - first run
  - last run
  - expected checksum (is there a way to store multiple checksums in one field or do we need a field per checksum?)

hackmastera:
Jobs required:

- [ ] Create a new fixity check (to be invoked by changeset persister on resource creation)
- [ ] Update least-recently-updated fixity check, then run self (to be invoked from a rake task)

Other stuff required:

- [ ] FixityCheck object
- [ ] Delete fixity check object when file is deleted
- [ ] Fixity dashboard (show failures, most / least recent checks)

hackmastera:
The only problem is that if you have two copies of the job running they will likely hit a race condition on checking the same item. maybe if we decide we need to do that we can add some kind of odds/evens switch on file IDs or something.

escowles:
@hackmastera Maybe it makes more sense to have one job that selects objects to check, and then a separate job for actually doing the fixity check?

hackmastera:
[...]

PR #779, opened 2018-02-02
Add FixityCheck, create it after characterization, delete it with object

hackmastera:
refs #142
for the rest of the plan, see https://github.com/pulibrary/figgy/issues/142#issuecomment-361742266

### DIFF
--- a/app/change_set_persisters/plum_change_set_persister/delete_fixity_check.rb
+++ b/app/change_set_persisters/plum_change_set_persister/delete_fixity_check.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+class PlumChangeSetPersister
+  class DeleteFixityCheck
+    attr_reader :change_set_persister, :change_set
+    delegate :resource, to: :change_set
+    delegate :query_service, to: :change_set_persister
+
+    def initialize(change_set_persister:, change_set:)
+      @change_set = change_set
+      @change_set_persister = change_set_persister
+    end
+
+    def run
+      return unless resource.respond_to? :original_file
+      resource.original_file.file_identifiers.each do |file_id| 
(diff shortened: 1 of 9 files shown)


app/change_set_persisters/plum_change_set_persister/delete_fixity_check.rb
```diff
@@ -0,0 +1,23 @@
...
+    def initialize(change_set_persister:, change_set:)
+      @change_set = change_set
+      @change_set_persister = change_set_persister
+    end
+
+    def run
+      return unless resource.respond_to? :original_file
+      resource.original_file.file_identifiers.each do |file_id| 
+        query_service.custom_queries.find_by_string_property(property: :file_id, value: file_id).
+          select{ |obj| obj.class == FixityCheck }.each do |fc|
```
tpendragon: Should we add an optional `model:` parameter to this query?

app/change_set_persisters/plum_change_set_persister/delete_fixity_check.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+class PlumChangeSetPersister
+  class DeleteFixityCheck
```
tpendragon: Maybe we should just store the fixity properties on the FileSet - maybe in a nested object to keep them nicely contained, if that matters? That way we don't have to worry about this - the fixity check will just be attached to the life of the fileset.

hackmastera: if it's nested doesn't that means it doesn't get its own created / updated dates? we are going to need those.

hackmastera: well updated, at least. I guess that could be managed manually.

hackmastera:
superceded by #799

PR #799, opened 2018-02-07
Add fixity fields to file_metadata, fixity check job

### LINKED ISSUE #142: Background fixity checking of files in Figgy

hackmastera:
resolves #142

### DIFF
--- a/app/jobs/check_fixity_recursive_job.rb
+++ b/app/jobs/check_fixity_recursive_job.rb
@@ -13,7 +13,7 @@ class CheckFixityRecursiveJob < ApplicationJob
   private
 
     def find_next_file_to_check
-      query_service.find_all_of_model(model: FileSet).map(&:decorate).sort_by(&:fixity_sort_date).first
+      query_service.find_all_of_model(model: FileSet).sort_by(&:updated_at).first
     end
 
     def metadata_adapter
--- a/app/jobs/check_fixity_recursive_job.rb
+++ b/app/jobs/check_fixity_recursive_job.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+class CheckFixityRecursiveJob < ApplicationJob
+  delegate :query_service, to: :metadata_adapter
+
+  def perform
+    file_set = find_next_file_to_check
+    original_file_metadata = file_set.run_fixity
+    file_set.file_metadata = file_set.file_metadata.select { |x| !x.original_file? } + Array.wrap(original_file_metadata)
+    metadata_adapter.persister.save(resource: file_set)
+    CheckFixityRecursiveJob.set(queue: queue_name).perform_later
+  end
+
+  private
+
+    def find_next_file_to_check
+      query_service.find_all_of_model(model: FileSet).map(&:decorate).sort_by(&:fixity_sort_date).first
+    end
+
+    def metadata_adapter
+      Valkyrie::MetadataAdapter.find(:indexing_persister)
+    end
+end
--- a/app/jobs/check_fixity_recursive_job.rb
+++ b/app/jobs/check_fixity_recursive_job.rb
@@ -3,7 +3,7 @@ class CheckFixityRecursiveJob < ApplicationJob
   delegate :query_service, to: :metadata_adapter
 
   def perform
-    file_set = find_next_file_to_check
+    file_set = query_service.custom_queries.least_recently_updated_file_set
     original_file_metadata = file_set.run_fixity
     file_set.file_metadata = file_set.file_metadata.select { |x| !x.original_file? } + Array.wrap(original_file_metadata)
     metadata_adapter.persister.save(resource: file_set)
@@ -11,10 +11,6 @@ class CheckFixityRecursiveJob < ApplicationJob
   end
 
   private
-
-    def find_next_file_to_check
-      query_service.find_all_of_model(model: FileSet).sort_by(&:updated_at).first
-    end
 
     def metadata_adapter
       Valkyrie::MetadataAdapter.find(:indexing_persister)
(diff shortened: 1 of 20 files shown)


app/jobs/check_fixity_recursive_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    original_file_metadata = file_set.run_fixity
+    file_set.file_metadata = file_set.file_metadata.select { |x| !x.original_file? } + Array.wrap(original_file_metadata)
+    metadata_adapter.persister.save(resource: file_set)
+    CheckFixityRecursiveJob.set(queue: queue_name).perform_later
+  end
+
+  private
+
+    def find_next_file_to_check
+      query_service.find_all_of_model(model: FileSet).map(&:decorate).sort_by(&:fixity_sort_date).first
```
escowles: Can we avoid loading and decorating all the FileSets?  We have ~3M FileSets now, and that will keep growing, so I'm worried about how well this will perform.

hackmastera: I don't really know what sorts of methods will load them and what won't. some guidance here would be appreciated.

tpendragon: We'll probably need a custom query for this that runs some SQL. We might need a custom index to make it sortable though..

hackmastera: hm. okay that `fixity_sort_date` method holds some logic that draws on multiple attributes; not sure whether that will translate to an SQL query. maybe? geez, it's been a long time since I did any SQL.

escowles: I believe this is the SQL we need to get the oldest fixity_last_run_date (never checked at the top), sub-sorted by created date:

```
select * from orm_resources
order by metadata->>'fixity_last_run_date' nulls first, metadata->>'created_at'
```

hackmastera: nice, thanks @escowles! it's slightly different logic but I don't have any compelling argument for the way I did it in light of performance concerns. I'll try this out.

hackmastera: Seems like I could add `limit 1` to the end of that, also, since i just need one at a time

tpendragon: We'll need some indexes on fixity_last_run_date and created_at to make that perform, I think. Running it on prod now to see..

hackmastera:
New plan:
- Just use updated_at instead of `fixity_last_run_date`
- Kick off a fixity check each time a `FileSet` is persisted.
- create an index for `updated_at`

hackmastera:
@tpendragon @escowles I assume we'll still need the custom query because finding / sorting would load everything?

escowles:
@hackmastera Yes, I think we'll still need a custom query — though maybe a "find the least-recently-updated-FileSet" query will be useful for other things too?

hackmastera:
👍 it's definitely more general, at least!

hackmastera:
This is one side-effect spec failure and one small PR to valkyrie short of being done.

hackmastera:
here's the relevant valkyrie PR: https://github.com/samvera-labs/valkyrie/pull/358

app/jobs/check_fixity_recursive_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+class CheckFixityRecursiveJob < ApplicationJob
+  delegate :query_service, to: :metadata_adapter
+
+  def perform
+    file_set = query_service.custom_queries.least_recently_updated_file_set
```
tpendragon: To reduce duplication, maybe we should just do `CheckFixityJob.perform_now(file_set.id)`?

app/jobs/check_fixity_recursive_job.rb
```diff
@@ -0,0 +1,18 @@
...
+    original_file_metadata = file_set.run_fixity
+    file_set.file_metadata = file_set.file_metadata.select { |x| !x.original_file? } + Array.wrap(original_file_metadata)
+    metadata_adapter.persister.save(resource: file_set)
+    CheckFixityRecursiveJob.set(queue: queue_name).perform_later
```
tpendragon: Should we catch everything above this line, so that if the job fails it's not over forever?

app/jobs/check_fixity_recursive_job.rb
hackmastera: I think I was just avoiding loading the file_set object twice. but if that doesn't seem like an issue to you i'm glad to do it!

app/jobs/check_fixity_recursive_job.rb
hackmastera: good idea

app/jobs/check_fixity_recursive_job.rb
tpendragon: Yeah, it only takes a couple dozen milliseconds to load - I don't mind doing it twice.

app/jobs/check_fixity_recursive_job.rb
hackmastera: it will probably just keep trying again on the same file set? is that bad?

tpendragon: Um. I'm not sure? It also strikes me that if it fails Sidekiq will try again - maybe it's better to just leave it that way.

hackmastera: yeah; rubocop no likey and sidekiq should take care of it. i'll revert that change.

PR #803, opened 2018-02-08
Initial Bag Exporter/Importer

tpendragon:
I think the only thing this is missing is:

1. Evaluating if we want to generate proper JSON-LD.
2. Payload-Oxum and Bag-Size in bag-info.txt


app/adapters/bagit/query_service.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def find_references_by(resource:, property:)
+      ids = (resource.try(property) || []).select { |id| id.is_a?(Valkyrie::ID) }
+      ids.lazy.map do |id|
+        find_by(id: id)
+      end
+    end
+
+    def find_all_of_model(model:)
+      Valkyrie.logger.warn("Bagit Query Service has been asked to find all resources of a specific type. This will require iterating over the metadata of every bag - AVOID.")
```
escowles: I like these warnings — I wonder if we should add them to the Fedora query service methods that use `find_all` too?

config/config.yml
```diff
@@ -6,6 +6,7 @@ defaults: &defaults
   plum_solr_url: <%= "http://127.0.0.1:#{ENV['TEST_JETTY_PORT'] || 8984}/solr/blacklight-core-test" %>
   plum_binary_path: "/mnt/diglibdata/hydra_binaries"
   plum_derivative_path: "/mnt/libimages1/data/jp2s/plum_prod"
+  bag_path: <%= Rails.root.join("tmp", "bags") %>
```
escowles: I assume we'll setup an ENV variable in production to put these somewhere outside of `tmp`

app/adapters/bagit/storage_adapter.rb
```diff
@@ -0,0 +1,81 @@
...
+        new_path = data_path.join("#{generate_id}-#{original_filename}")
+        old_path = file.try(:disk_path) || file.path
+        FileUtils.cp(old_path, new_path)
+        output_file = find_by(id: Valkyrie::ID.new("bag://#{new_path.relative_path_from(base_path)}"))
+        create_manifests(output_file)
+        output_file
+      end
+
+      def create_manifests(file)
+        checksums = file.checksum(digests: [Digest::SHA1.new, Digest::MD5.new, Digest::SHA256.new])
```
escowles: We talked about this on Slack, but I'm putting a note here so I don't forget: it would be better to use the existing checksums from the file metadata — both because it's faster and because it would catch any corruption that might happen when copying the files.

config/config.yml
tpendragon: Done.

app/adapters/bagit/query_service.rb
tpendragon: Yeah, probably not a bad idea.

PR #817, opened 2018-02-14
Ensures that `BulkIngestService#attach_dir` can only parse directory content using an absolute path

### LINKED ISSUE #741: IngestFolderJob should use an absolute path instead of the release path
If we deploy before an old job starts the directory no longer exists.

jrgriffiniii:
Resolves #741

### DIFF
--- a/app/controllers/scanned_resources_controller.rb
+++ b/app/controllers/scanned_resources_controller.rb
@@ -71,48 +71,4 @@ class ScannedResourcesController < BaseResourceController
       volume_count: locator.volume_count
     }
   end
-
-  class IngestFolderLocator
-    attr_reader :id
-    def initialize(id:)
-      @id = id
-    end
-
-    def root_path
-      Pathname.new(BrowseEverything.config["file_system"][:home]).join("studio_new")
-    end
-
-    def exists?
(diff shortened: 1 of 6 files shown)


escowles:
@jrgriffiniii The ticket isn't clear on this, but the problem we've seen isn't with relative paths, but with paths that are prefixed with a release directory (which might not exist on all machines, or may have disappeared by the time a job runs if there have been a lot of releases).  E.g., when we spawn an ingest job, it currently has the path prefixed in the release (e.g., `/opt/figgy/releases/20180123142369/staged_files/hydra_sources/studio_new`) instead of the shared version (`/opt/figgy/shared/staged_files/hydra_sources/studio_new`) which is really a symlink to the real location (`/mnt/hydra_sources/studio_new`).

I'm not sure what the best approach is here — whether there's an easy way to get the shared directory name, or just replacing `releases/[timestamp]` with `current`, etc.

jrgriffiniii:
In order for #741 to be fully resolved, a separate issue will need to be created in [pulibrary/princeton_ansible](https://github.com/pulibrary/princeton_ansible) where the necessary environment variable is set and passed in the Playbook.

PR #820, opened 2018-02-14
Poster Importer

tpendragon:
Still needs tested on staging.


tpendragon:
Creates objects like this: https://figgy-staging.princeton.edu/catalog/parent/a7441740-343a-428b-aff8-f06cf4fd4a92/369cf4af-6a2c-4f9b-9470-168e5640f6d8

PR #821, opened 2018-02-15
Import GeoBlacklight document generation

### LINKED ISSUE #230: Migrate GeoBlacklight document generation
Migrate services for GeoBlacklight document generation.

eliotjordan:
GeoBlacklight documentation imported from GeoWorks.

Closes #230


app/services/geo_resources/discovery/document_builder/date_builder.rb
```diff
@@ -0,0 +1,51 @@
...
+        private
+
+          # Returns the date the layer was issued.
+          # @return [String] date in XMLSchema format.
+          def issued
+            datetime = resource_decorator.issued.first
+            datetime = DateTime.parse(datetime).utc
+            datetime.utc.xmlschema
+          rescue
+            ''
```
jrgriffiniii: Should a warning be logged for cases where the datestamp cannot be normalized and formatted?

app/services/geo_resources/discovery/document_builder/date_builder.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          # Returns a year associated with the layer. Taken from first
+          # value in temporal or from resource created date.
+          # @return [Integer] year
+          def layer_year
+            date = resource_decorator.temporal.first || resource_decorator.model.created_at
+            year = date.match(/(?<=\D|^)(\d{4})(?=\D|$)/)
+            year ? year[0].to_i : nil
+          rescue
+            ''
```
jrgriffiniii: Handling errors for empty `String` and `nil` values seem to be of greater importance regarding consistency in discovery (I'm assuming that this is the value populating `solr_year_i`: https://maps.princeton.edu/catalog/princeton-ms35tb05h.json

app/services/geo_resources/discovery/document_builder/layer_info_builder.rb
```diff
@@ -0,0 +1,72 @@
...
+            return unless file_set
+            file_set.mime_type.first
+          end
+
+          # Uses parent work class to determine file geometry type.
+          # These geom types are used in geoblacklight documents.
+          # @return [String] file geometry type
+          def geom_type
+            case resource_decorator.model_name.to_s
+            when /ScannedMap/
```
jrgriffiniii: Why are `RegExp`s used here?  This is not a criticism, but I'm curious as to to why more precise Class-based matching isn't used (using `case resource_decorator.model` and `when ScannedMap`...)

app/services/geo_resources/discovery/document_builder/date_builder.rb
eliotjordan: I'm not terribly worried about a missing issued property because of a malformed date. We don't even display the field in Pulmap. Probably good to add it to our docs, if we can, but no big deal if we can't.

app/services/geo_resources/discovery/document_builder/date_builder.rb
eliotjordan: It populates `solr_year_i`. This method _should_ always return a year of some kind because it pulls from created_at.

eliotjordan: I suppose we could test if temporal is valid first and then use `created_at`.

app/services/geo_resources/discovery/document_builder/layer_info_builder.rb
eliotjordan: This was imported from GeoWorks, so I'm not sure exactly why the class based matching wasn't done. I left it as is because we haven't implemented a RasterWork resource yet, and I want to keep this as placeholder. When we do RasterWorks, it would be a good idea to update this method as you suggest.

PR #834, opened 2018-02-21
Gives unique names to figgy solr cores

### LINKED ISSUE #796: Give Solr cores names that are unique to figgy
Most of our apps use the default naming convention when it comes to our solr cores. This can lead to confusion and errors. something like "figgy-core-development" and "figgy-core-test" would probably be fine.

sdellis:
Fixes #796

### DIFF
--- a/config/blacklight.yml
+++ b/config/blacklight.yml
@@ -12,13 +12,13 @@
 
 development:
   adapter: solr
-  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/blacklight-core" %>
+  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/figgy-core-dev" %>
 test: &test
   adapter: solr
-  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:#{ENV['TEST_JETTY_PORT'] || 8984}/solr/blacklight-core-test" %>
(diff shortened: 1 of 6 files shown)


config/blacklight.yml
```diff
@@ -12,13 +12,13 @@
 
 development:
   adapter: solr
-  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/blacklight-core" %>
+  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/figgy-core-dev" %>
 test: &test
   adapter: solr
-  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:#{ENV['TEST_JETTY_PORT'] || 8984}/solr/blacklight-core-test" %>
+  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:#{ENV['TEST_JETTY_PORT'] || 8983  }/solr/figgy-core-test" %>
```
escowles: Changing the core name is great, but please don't change the port (which can be overridden with an environment variable anyway)

sdellis: If these are two different cores, why do they also need two different ports? At any rate, for some reason, that ENV variable has not worked for me in the past. Let me set up the cores manually and try it again.

escowles: The typical setup is to run `solr_wrapper` with different configurations for dev and test, on different ports.  The main advantage of this is that they are completely independent, and can be run at the same time, restarted, etc. without conflicting or impacting the other one.

eliotjordan:
This looks good to me. 👍

PR #847, opened 2018-02-22
Adds scanned map importer

### LINKED ISSUE #846: Migrate scanned maps and map sets from Plum

eliotjordan:
- Duplicates existing plum importer with some modification to import scanned maps and map sets.
- Sets file set mime_type with Plum geo_mime_type.
- Accounts for records with no bibid whose metadata comes from FGDC files.
- Makes subject property multivalued.

Closes #846


app/services/plum_scanned_map_importer.rb
```diff
@@ -0,0 +1,521 @@
+# frozen_string_literal: true
+class PlumScannedMapImporter
+  attr_reader :id, :change_set_persister
+  attr_writer :all_collections, :id_cache
```
jrgriffiniii: This is perhaps following https://github.com/pulibrary/figgy/blob/88b4bafeeb7bfacc67af276e2238887058749157/app/services/plum_importer.rb#L4, but I does `id_cache` need to be a public mutator method?

PR #856, opened 2018-02-27
Scanned map thumbnail_id points to a file set

### LINKED ISSUE #855: Generating geoblacklight document for MapSets throws an error
https://app.honeybadger.io/projects/53391/faults/36878078

An issue with FileSet vs. ScannedMap members.

eliotjordan:
GeoBlacklight document builders assumed that a map set's thumbnail id pointed to a ScannedMap. The thumbnail id should be the id for a file set instead.

Closes #855

### DIFF
--- a/app/services/geo_resources/discovery/document_builder/document_path.rb
+++ b/app/services/geo_resources/discovery/document_builder/document_path.rb
@@ -80,9 +80,8 @@ module GeoResources
           def map_set_file_set
             member_id = Array.wrap(resource_decorator.thumbnail_id).first
             return unless member_id
-            member = query_service.find_by(id: member_id)
-            member_file_sets = member.decorate.geo_members
-            member_file_sets.first.decorate unless member_file_sets.empty?
+            file_set = query_service.find_by(id: member_id)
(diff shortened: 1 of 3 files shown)


app/services/geo_resources/discovery/document_builder/document_path.rb
```diff
@@ -80,9 +80,8 @@ def host
           def map_set_file_set
             member_id = Array.wrap(resource_decorator.thumbnail_id).first
             return unless member_id
-            member = query_service.find_by(id: member_id)
-            member_file_sets = member.decorate.geo_members
```
jrgriffiniii: I did not realize that `MapSet` members are `FileSets` retrieved using `#thumbnail_id`

eliotjordan: This is for building the GeoBlacklight document. GBL docs, especially those that are restricted, need a thumbnail url reference for display in Pulmap. The method above fetches the map set's thumbnail id, which should reference a file set contained in one of the child scanned map documents.

PR #863, opened 2018-03-02
Add raster resource model, change set, decorator, controller, and views

### LINKED ISSUE #210: Migrate RasterWork
- [ ] Model / metadata
- [ ] ChangeSet
- [ ] Controller
- [ ] Decorator

eliotjordan:
Closes #210


app/services/geo_resources/discovery/document_builder/layer_info_builder.rb
```diff
@@ -50,14 +50,17 @@ def geo_mime_type
...
-              'Image'
-            when /RasterWork/
-              'Raster'
-            when /VectorWork/
-              vector_geom_type
-            end
+            return 'Image' if resource_decorator.model.is_a?(ScannedMap)
+            return 'Raster' if resource_decorator.model.is_a?(RasterResource)
+            return vector_geom_type if resource_decorator.model.is_a?(VectorWork)
+            # case resource_decorator.model_name.to_s
```
escowles: Delete the commented-out previous code?

app/models/raster_resource.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  attribute :id, Valkyrie::Types::ID.optional
+  attribute :member_ids, Valkyrie::Types::Array
+  attribute :member_of_collection_ids
+  attribute :state
+  attribute :workflow_note, Valkyrie::Types::Array.member(WorkflowNote).optional
+  attribute :file_metadata, Valkyrie::Types::Set.member(FileMetadata.optional)
+  attribute :pending_uploads, Valkyrie::Types::Array.member(PendingUpload)
+
+  # How is this used?
```
escowles: I suspect this is used when a property links to a resource, and it uses the resource's `.to_s` method for display:

https://github.com/pulibrary/figgy/blob/master/app/views/catalog/_resource_attributes_default.html.erb#L23

eliotjordan: This was a comment carried over from Vector Works. Thanks for linking to that partial. I'll remove the comment in both places.

app/services/geo_resources/discovery/document_builder/layer_info_builder.rb
eliotjordan: Ouch. Yep.

PR #877, opened 2018-03-06
Change title at top to say Figgy

### LINKED ISSUE #3: Branding
Plum/Figgy should probably have its own brand, and we should use that instead of DPUL (see https://github.com/pulibrary/figgy/pull/2/files#diff-37c9b835927e3c363588f172de9bb6faR24).

That brand might be "Plum" or it might be something else, but it shouldn't be DPUL.

tpendragon:
Closes #3


tpendragon:
screen shot 2018-03-06 at 10 25 40 am

PR #885, opened 2018-03-09
Adds progress bar for loading images

### LINKED ISSUE #762: VUE File Manager should show "loading" icon or similar while it's waiting for IIIF Manifest
It's not super long, but it takes a couple seconds to load a 1300 page manifest.

sdellis:
fixes #762

### DIFF
--- a/valhalla/app/javascript/components/Thumbnails.vue
+++ b/valhalla/app/javascript/components/Thumbnails.vue
@@ -21,22 +21,31 @@
         <i class="fa fa-image fa-lg"></i>
       </div>
     </div>
-    <draggable v-model="thumbnails" tag="div" name="list-complete" class="img_gallery">
-        <div @click.capture="select(thumbnail.id, $event)"
-              v-bind:style="{'max-width': thumbPixelWidth + 'px' }"
-              class="thumbnail"
-              v-bind:class="{ hasChanged: hasChanged(thumbnail.id), selected: isSelected(thumbnail) }"
-              v-for="thumbnail in thumbnails" :key="thumbnail.id">
-          <img :src="thumbnail.url" class="thumb">
-          <div v-bind:style="{'padding': captionPixelPadding + 'px' }" class="caption">
-            {{thumbnail.label}}
+    <transition name="fade">
+      <div v-if="show" class="progress">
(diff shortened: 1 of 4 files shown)


valhalla/app/javascript/components/Thumbnails.vue
```diff
@@ -84,6 +101,17 @@ export default {
     }
   },
   methods: {
+    imageProgress: function (instance, image ) {
+      const result = image.isLoaded ? 'loaded' : 'broken';
+      var total = this.thumbnails.length
+      this.numLoaded = this.numLoaded + 1
+      this.pctLoaded = this.numLoaded * 100 / total
+      console.log( 'image is ' + result + ' for ' + image.img.src )
```
tpendragon: Do we want to leave the console logs?

sdellis: No, nice catch.

sdellis:
I am going to remove the overlay, but keep the progress bar in the gallery (just make it thin and out of the way). I still think it's helpful to see when all the images have loaded, but it will not prevent one from working with the filemanager before all the images are loaded.

Let's keep this PR open while I work on it. The new focus will be on a loading indicator for large manifests, not the images. The original issue was worded clearly, but since I've only been working with small manifests in dev, I mis-focused on the images which is what were taking a while to load for me.

sdellis:
In order to test a large manifest, I tried seeding works with lots of images by upping the number of files in this line to 1000:
https://github.com/pulibrary/figgy/blob/master/db/seeds.rb#L13

It took a long time, but I only got 7 files with 1 image each.

I'm going to need some guidance on how to do this come Monday.
(Also getting some Valkyrie failed tests in CircleCI, unrelated to this work.)

sdellis:
@tpendragon looking at this more closely and it seems that the manifest generation (over 10 seconds in my example case) is actually blocking the Rails view rendering, so a loader/spinner does not appear while the user is waiting for the view to render. If we can render the manifest asynchronously, or after the view renders, we may be able to address this with a loader.

tpendragon:
@sdellis I'm confused. The file manager requests the manifest right? The only reason I can see that it would stop the rendering is if the file manager either isn't using ajax or the web server has one thread.

Issue #811, opened 2018-02-12
fixity dashboard


hackmastera:
show failures, most / least recent checks

PR #890, opened 2018-03-12
Fixity dashboard

hackmastera:
Query / controller work that refs #811


app/queries/most_recently_updated_file_sets.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def most_recently_updated_file_sets
+    run_query(query)
+  end
+
+  def query
+    <<-SQL
+      select * FROM orm_resources WHERE
+      internal_resource='FileSet'
+      order by updated_at desc
```
escowles: I think it's probably worth combining the impl here with `LeastRecentlyUpdatedFileSets` — and maybe asking whether they should be a single class with a parameter to flip the ordering.

hackmastera: Can I do that like `order by updated_at ?` and a sql substitution? I tried doing something like that with a `limit` and it didn't seem to accept the `?`.

app/controllers/dashboard_controller.rb
```diff
@@ -0,0 +1,12 @@
+class DashboardController < ApplicationController
```
escowles: Are we eventually going to have fixity and other things on a single dashboard?  If so, I'm fine with `DashboardController`.  But if this is only going to be for fixity, it might be better to call it `FixityController` instead.

app/queries/most_recently_updated_file_sets.rb
escowles: I don't know if you can do it with SQL interpolation, but you couldn't you put the order statement in a separate method and concatenate them?

app/controllers/dashboard_controller.rb
hackmastera: Yeah, my thinking was that it could be built on as we found more things we needed to put on a dashboard, with some sort of tabbed UI. Happy to change it, though, if you don't think there will be such use cases, or if you just prefer to wait until we actually implement them. I just thought it would be a helpful flag for remembering to do so if/when the time comes, down the line.

escowles: That makes sense.  Though it makes me think the recently-created ReportsController should be folded into this, too.

app/queries/most_recently_updated_file_sets.rb
hackmastera: ah, good point; it's just a string isn't it :D okay i can work something out here

app/controllers/dashboard_controller.rb
hackmastera: I considered that pretty seriously. In the end i sort of thought reports and dashboards are different things. but maybe that doesn't matter because they have similar use cased scenarios? superuser-type functionality, etc. Shall I fold the report in and call it all "dashboard"?

escowles: 👍 Sounds good to me.

hackmastera:
@escowles I made those changes. I left the report where it is in the UI for now; i can either address it in the views PR or we can create an issue for the future.

Issue #894, opened 2018-03-12
Redesign 'home' as a dashboard


hackartisan:
We're starting to add reports and other data displays. The create icons are taking up a lot of real estate. This should be redesigned as a proper dashboard since that's the use it's shaping up to fill.

PR #895, opened 2018-03-12
Add a fixity dashboard to figgy

### LINKED ISSUE #811: fixity dashboard

hackmastera:
(screenshots to come)
Fixes #811


hackmastera:
Link on the home page:
screen shot 2018-03-12 at 5 20 23 pm

Full fixity dashboard page:
screen shot 2018-03-12 at 5 20 47 pm

hackmastera:
@escowles you said something about a chart or mini-report on the home page. To do that I think I'll need to create a model object or service object or something besides just the `FixityDashboardController`, which I didn't do in the initial design, to avoid adding more calls to the query service in weird places. If that is desired perhaps we should ticket it and link it to #894

PR #903, opened 2018-03-13
Add characterization and derivative generation services for raster and vector resources

### LINKED ISSUE #216: Port geo-derivative generation services
Port geo-derivative generation services from GeoWorks. Much of this will work as is. Could use some re-factoring, if there's time.

eliotjordan:
Closes #216


circle.yml
```diff
@@ -11,8 +11,12 @@ dependencies:
     - curl -L -o google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
     - sudo dpkg -i google-chrome.deb
     - rm google-chrome.deb
+    - sudo apt-get update
```
escowles: Looks like we're doing `apt-get update` twice.

circle.yml
```diff
@@ -11,8 +11,12 @@ dependencies:
...
+    - sudo apt-get update; sudo sudo apt-get install libgdal1h libgdal-dev gdal-bin libcairo2-dev libpango1.0-dev
```
escowles: The list of packages to install here isn't exactly the same as what's in the README — do we want to just include this line in the README?

spec/spec_helper.rb
```diff
@@ -25,6 +25,13 @@
...
   config.example_status_persistence_file_path = "tmp/rspec_examples.txt"
   config.order = :random
+
+  # Cleanup test files and derivatives
+  config.after(:suite) do
+    break unless Rails.env.test?
+    FileUtils.rm_rf(Figgy.config['derivative_path'])
+    FileUtils.rm_rf(Figgy.config['repository_path'])
+  end
 end
```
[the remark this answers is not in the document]
escowles: 👍

circle.yml
eliotjordan: Fixed.

PR #906, opened 2018-03-13
Load folder data asynchronously.

### LINKED ISSUE #838: Project index page loads very slowly when there are many attached ephemera folders.
LAE has ~4,000 posters attached, and it takes a while to load up.

tpendragon:
Closes #838.

### DIFF
--- a/app/queries/count_members.rb
+++ b/app/queries/count_members.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+class CountMembers
+  def self.queries
+    [:count_members]
+  end
+
+  attr_reader :query_service
+  delegate :orm_class, to: :resource_factory
+  delegate :resource_factory, to: :query_service
+  def initialize(query_service:)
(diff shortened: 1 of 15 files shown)


app/queries/count_members.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  attr_reader :query_service
+  delegate :orm_class, to: :resource_factory
+  delegate :resource_factory, to: :query_service
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def count_members(resource:, model: nil)
```
hackmastera: @tpendragon can you add a comment here to clarify what happens if you do / don't provide a model? It would count all members if you don't, or just members of a certain model if you do?

hackmastera: ah, yes I see it in the spec. I still think it would be good to have a comment. I can PR it.

PR #939, opened 2018-03-20
Integrates Google Drive support as a `browse-everything` provider

### LINKED ISSUE #851: Update `browse-everything` for support with Google Drive
https://github.com/samvera/browse-everything/pull/189 ensures that the support for the latest Google Drive API is now supported as a provider

jrgriffiniii:
Resolves #851

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,7 +12,7 @@ AllCops:
     - 'db/schema.rb'
     - 'spec/support/**/*'
     - 'node_modules/**/*'
-    - 'simple-tiles/**/*'
+    - 'tmp/**/*'
 Style/FileName:
   Enabled: true
   Exclude:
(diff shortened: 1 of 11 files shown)


.rubocop.yml
```diff
@@ -20,9 +20,22 @@ Style/FileName:
...
+    - 'simple-tiles/Vagrantfile'
+Style/BarePercentLiterals:
+  Exclude:
+    - 'simple-tiles/Vagrantfile'
+Style/PercentLiteralDelimiters:
+  Exclude:
+    - 'simple-tiles/Vagrantfile'
+Style/UnneededPercentQ:
+  Exclude:
+    - 'simple-tiles/Vagrantfile'
```
eliotjordan: That simple-tiles directory is an artifact of compiling the simple tiles library on circle. I would have thought that AllCops exclusion on line 15 would have taken care of this. Does rubocop fail without these now? Alternately, we can remove that file as part of the circle configuration process. Maybe add a `rm Vagrantfile` here: https://github.com/pulibrary/figgy/blob/master/bin/ci_simple_tiles_install.sh#L11

Gemfile
```diff
@@ -71,7 +71,8 @@ end
 gem 'rsolr', '>= 1.0'
 
 gem 'aasm'
-gem 'browse-everything'
+gem 'browse-everything', github: 'samvera/browse-everything'
+# gem 'extracted_library', :path => './vendor/extracted_library'
```
eliotjordan: Remove?

.rubocop.yml
escowles: 👍 Can we move the simple-tiles directory to tmp and exclude the entire tmp directory?  It seems like Rubocop shouldn't be checking stuff like this that's not our code.

PR #940, opened 2018-03-20
Reconfigure event generation to add geoblacklight events

### LINKED ISSUE #828: Port GeoBlacklight messaging from Plum and GeoWorks
Port RabbitMQ-based GeoBlacklight messaging from Plum and GeoWorks.

eliotjordan:
Adds geoblacklight event generation and message functionality while attempting to minimize changes to existing interfaces and specs. I tried other approaches, such as in the change_set_persister, however this one seemed the least disruptive.

Closes #828


escowles:
This looks reasonable, but I'd like @tpendragon to take a look too.

app/services/event_generator/composite_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+    attr_reader :generators
+
+    def initialize(generators)
+      @generators = generators
+    end
+
+    private
+
+      # Send method and args to generators.
+      def method_missing(m, *args)
```
tpendragon: Rather than depending on which argument gets passed and using method_missing, maybe we should just put the EventGenerator interface as methods on the composite?

app/services/event_generator/geoblacklight_event_generator.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+class EventGenerator
+  class GeoblacklightEventGenerator
```
tpendragon: These are really great. Maybe a shared spec for what the expected event generator interface is?

eliotjordan: Added shared spec.

app/services/event_generator/composite_generator.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def record_updated(record)
+      delegate_to_generator(__method__, record)
+    end
+
+    private
+
+      # Send method with record argument to first valid generator
+      def delegate_to_generator(method_name, record)
+        generators.each do |generator|
```
tpendragon: Can probably just do `generators.find{|generator| generator.valid?(record)}.send(method_name, record)`

hackmastera:
i like it

app/services/event_generator/composite_generator.rb
eliotjordan: I think we still need to check if the generator responds to the method in addition to validity. Alternatively, each generator needs to respond to all of the composite generator methods.  For example, the Geoblacklight generator only responds to `record_updated` and `record_deleted` at the moment. We would have to add an empty method definition:

```
def record_created; end
```

Thoughts? I could go either way.

tpendragon: I'm personally of the opinion that we should have the empty method definition. I'm surprised they pass the shared spec otherwise - I would expect `record_created` to be part of the interface.

PR #959, opened 2018-03-27
CircleCI Workflow

tpendragon:


.circleci/config.yml
```diff
@@ -44,9 +44,31 @@ jobs:
...
+    test:
+      working_directory: ~/figgy
+      docker:
+        - image: circleci/ruby:2.3.6-node-browsers
+          environment:
+            RAILS_ENV: test
+            FIGGY_DB_HOST: localhost
+            FIGGY_DB_USERNAME: figgy
+            FIGGY_DB_PASSWORD:  ""
+        - image: postgres:9.5
```
escowles: Do we want to start using postgres 10 for CI?

tpendragon: Sure, if I could get CircleCI to run - it stopped running my builds. >.>

PR #965, opened 2018-03-28
Add jp2 derivatives for jpeg source files

### LINKED ISSUE #871: Support JPEG as source images for scanned resources

hackmastera:
fixes #871

Note: This implementation uses the Hydra::Derivatives::ImageDerivatives
processor pretty naively. In particular, the jp2 created is larger than
the source jpeg. Possible next steps:

 * Look into passing more specific instructions to minimagick through
 hydra derivatives (likely involves upstream commits).
 * Use minimagick directly, without hydra derivatives.
 * Decide this is good enough.

Some sample output pre- and post-conversion:
```
$ identify /Users/headleya/src/figgy/staged_files/large-jpg-test.jpg
/Users/headleya/src/figgy/staged_files/large-jpg-test.jpg JPEG 1867x2801
1867x2801+0+0 8-bit sRGB 1.78797MiB 0.010u 0:00.009
```

```
$ identify /Users/headleya/src/figgy/tmp/test_files/fe/42/a5/fe42a50fde50483bb9ee0be1cf673a8f/intermediate_file.jp2
/Users/headleya/src/figgy/tmp/test_files/fe/42/a5/fe42a50fde50483bb9ee0be1cf673a8f/intermediate_file.jp2
JP2 1867x2801 1867x2801+0+0 8-bit sRGB 4.48735MiB 0.000u 0:00.009
```


hackmastera:
If we want to try to make these derivatives smaller, the best way to go (suggested by @jrgriffiniii) is probably to derive a class from the shell based processor:
https://github.com/samvera/hydra-derivatives/blob/master/lib/hydra/derivatives/processors/shell_based_processor.rb

We'd likely need to use some jp2-specific options, though I'm not sure exactly which ones. Here is some relevant documentation: https://www.imagemagick.org/script/jp2.php

escowles:
@hackmastera I think this is fine as-is.  Having overly-large JP2s isn't great, but under the circumstances, I think it's acceptable.

But if you want to explore this further, I think one difference that might be important here is that `Hydra::Derivatives::ImageDerivatives` is using OpenJPEG to generate the JP2 instead of Kakadu. So it might be worth trying a two-step process: convert the JPEG to an uncompressed TIFF, and then using `Hydra::Derivatives::Jpeg2kImageDerivatives` to generate the JP2.

hackmastera:
Just trying it on the command line, converting to tiff first and then using kakadu does result in a smaller file, which is slightly smaller than the original jpg. See below; the one created from the intermediate tiff is called `output.jp2`:

```
-rw-r--r--@  1 headleya  staff   4.5M Mar 28 10:00 large-jpg-test.jp2
-rw-r--r--@  1 headleya  staff   1.8M Mar 27 13:10 large-jpg-test.jpg
-rw-r--r--   1 headleya  staff    15M Mar 28 14:26 large-jpg-test.tiff
-rw-r--r--@  1 headleya  staff   1.5M Mar 28 14:29 output.jp2
```

hackmastera:
Okay marking this do not merge as there may be some time this week for @jrgriffiniii to look into this further. There's a WIP commit (https://github.com/pulibrary/figgy/commit/f4fa40355f489da9b400c67d36bf4ad308f1b3f5) on branch https://github.com/pulibrary/figgy/commits/intermediate_tiff; I've briefed him on the status but to summarize:

When you try to clean up the intermediate tiff it looks like figgy was expecting to attach it to the file set. When I ran it without the cleanup (in which case tests pass) I got an error in sidekiq that the tif couldn't be opened. So was unable to see whether multiple derivatives were being added to the resource.

If there ends up not being time to work on this more, and we don't mind the size of the jp2s, this PR can actually be merged.

PR #975, opened 2018-03-28
Allows for simple cut and paste in Vue Filemanager

### LINKED ISSUE #585: Add Cut and Paste for selected filesets in Vue Filemanager
Should be part of right-click context menu as well as standard key commands. Increase transparency to indicate "cut" files.

sdellis:
Fixes #585

### DIFF
--- a/valhalla/app/javascript/components/Thumbnails.vue
+++ b/valhalla/app/javascript/components/Thumbnails.vue
@@ -21,8 +21,9 @@
           <span class="caret"></span>
         </button>
         <ul class="dropdown-menu" aria-labelledby="withSelected">
-          <li><a id="select_all_btn">Cut</a></li>
-          <li class="disabled"><a id="select_none_btn">Paste</a></li>
+          <li v-bind:class="{ disabled: isCutDisabled() }"><a @click="cutSelected()" id="cut_btn">Cut</a></li>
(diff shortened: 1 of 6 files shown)


valhalla/app/javascript/components/Thumbnails.vue
```diff
@@ -105,13 +137,34 @@ export default {
...
+      }
+    },
     isSelected: function (thumbnail) {
       if (this.selected.indexOf(thumbnail) > -1) {
         return true
       } else {
         return false
       }
     },
+    isCutDisabled: function () {
```
jrgriffiniii: Might this be refactored into an arrow function expression?  (It seems to me that this is something similar to just returning `!!this.cut.length`)

sdellis: I will look at it. At the very least, I could reduce line numbers by using ternary operators in all of those types of functions.

sdellis: @jrgriffiniii I was having trouble using arrow functions in this context... the Vue docs use the current style. However, I did reduce the if/then cruft, making the functions much shorter. I also fixed a couple of cut/paste bugs that I didn't notice. This should be ready to merge.

Issue #747, opened 2018-01-17
Sammelband Support in Order Manager


escowles:
We should be able to manage both files and volumes at the same time in the file manager, in order to support sammelbands, which have covers/frontmatter and multiple child volumes.  Maybe we can simply revert this: https://github.com/pulibrary/figgy/blob/master/valhalla/app/javascript/components/Thumbnails.vue#L16 — but there is additional functionality (like correcting sammelbands ingested as single volumes), so this should probably be part of a larger effort to make that work smoothly.

Child of #247

sdellis:
I agree that we should be able to manage files and volumes in the file manager. I looked into this more closely and there is an unrelated bug that is preventing the "flattened" MVW from displaying. I fixed the bug (PR coming any minute), but it doesn't address some of the other sammelband use cases you outline (i.e., ingested as single volumes, or adding single images to MVWs without flattening). 

This will be part of a larger effort as it will require significant changes on both front and backend. For example, if someone adds filesets a la carte to a MVW, we should generate a manifest that represents a sammelband via the IIIF Presentation spec (using Ranges to represent Volumes instead of Manifests) since that's how the viewers will use it as well. I think that the Filemanager could evolve to manage structure too, which would provide the grouping/splitting functionality needed for these use cases.

tpendragon:
@sdellis A lot of time went into figuring out how to represent Sammelbands. In Plum they do what you're saying - in Figgy the flattening happens, but we just haven't written the range builder.

hackmastera:
In the meantime, should we consider detecting sammelbands and displaying a warning?

hackmastera:
oops, mistakenly autoclosed

sdellis:
Sammelbands currently have a `viewingHint=multi-part` although this is [only valid on Collections](http://prezi3.iiif.io/api/presentation/3.0/#behavior). What is the purpose of setting viewingHint to multi-part here?

tpendragon:
It shouldn't have that viewingHint. Made a new issue: #1565

tpendragon:
Closed - all materials work now since moving to GraphQL.

PR #983, opened 2018-03-29
Allows user to set thumbnail for MVW

### LINKED ISSUE #783: Vue Filemanager should allow setting thumbnail for MVW

sdellis:
fixes #783

### DIFF
--- a/valhalla/app/javascript/components/Controls.vue
+++ b/valhalla/app/javascript/components/Controls.vue
@@ -116,6 +116,7 @@ export default {
       body.resource[this.resourceClassName] = {
         member_ids: this.imageIdList,
         viewing_direction: this.viewingDirection,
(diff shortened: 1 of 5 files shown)


jrgriffiniii:
I can set the thumbnail (both locally and in staging) when files are *only* uploaded for attached (child) scanned resources.  This is also the case if I upload and subsequently delete all `FileSets` for the parent scanned resource (i. e. I can then set the thumbnails).

I can create a separate issue for this if my initial case exhibiting the unintended behavior is not intended to be supported.

sdellis:
@jrgriffiniii Yes, the case you are describing not working is a Sammelband. The manifest builder does not build Sammelbands correctly as it's flattening Attached Children canvases into the sequence with other canvases, but not creating the Range data that would model the logical Volume. This is a known issue... see #747

PR #997, opened 2018-04-02
Ensures that only Collection titles are featured within the IIIF Manifest metadata

jrgriffiniii:
Discovered while attempting to resolve https://github.com/pulibrary/pomegranate/issues/336

### DIFF
--- a/app/services/pdf_generator/canvas_downloader.rb
+++ b/app/services/pdf_generator/canvas_downloader.rb
@@ -48,11 +48,6 @@ class PDFGenerator
         [max_dimensions[:width], canvas.width].min
       end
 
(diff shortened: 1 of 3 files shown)


app/services/pdf_generator/canvas_downloader.rb
```diff
@@ -48,11 +48,6 @@ def max_width
         [max_dimensions[:width], canvas.width].min
       end
 
-      def max_height
```
tpendragon: Why was this removed?

tpendragon: Ah, I see the commit message.

PR #1021, opened 2018-04-05
Using an upstream bug resolution in order to ensure that file system paths for `browse-everything` with punctuation can be resolved

jrgriffiniii:
Originally reported by @roelmoon and scoped within https://github.com/samvera/browse-everything/issues/204

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,7 +71,7 @@ end
 gem 'rsolr', '>= 1.0'
 
 gem 'aasm'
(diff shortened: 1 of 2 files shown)


jrgriffiniii:
This doesn't appear to be strictly limited to the `browse-everything` bug, as there are also issues with the samba share on the staging server:
```
$ ls /mnt/[...]/PUAM_Ethiopian\ /
ls: cannot access '/mnt/[...]/PUAM_Ethiopian /': No such file or directory
```

jrgriffiniii:
@kayiwa resolved the aforementioned problem on the Samba server issue using a symbolic link

Issue #752, opened 2018-01-22
Background fixity checking of preservation Bags


escowles:

hackmastera:
* Find the bag on disk
* run bag verification (this is more than just fixity -- presumably there's info on this in the bagit spec)
* record success on resource
* save resource directly to persister (bypass the change set persister so another bag is not exported, plus we don't need any of the other stuff that's in there)
* queue the next check

hackmastera:
@escowles @tpendragon y'all seemed to know about some external validation tool I should be using here. where should I look for that?

escowles:
@hackmastera The main tools I was thinking of were `md5sum` and `shasum`, which you can use to validate the files match the checksums in the manifests:

```
md5sum -c manifest-md5.txt
```

There are also BagIt tools from LoC that can be used to validate:

* https://github.com/LibraryOfCongress/bagit-python
* https://github.com/LibraryOfCongress/bagit-java

hackmastera:
@escowles I just dropped in the python bagit tool and it works well. python is currently installed on production servers but is not a figgy dependency. Would that be something we want to add? If not I'm happy to go back to my own thing; it was getting pretty close.

escowles:
@hackmastera I think it's worth adding python to the Ansible playbooks, and using the python-bagit (which I assume is doing a lot of verification of the Bag structure beyond just checking the checksums).

hackmastera:
The hart of this work is done on https://github.com/pulibrary/figgy/tree/bag_validation_python (PR'd in #1034) but it requires a python to be added via ansible (see princeton_ansible#245) and some additional work to stub responses from the python script in the specs. Because the adoption of that PR is dependent on preservation and backup strategy decisions still in-process, I'm closing the PR for now. We can pick this back up when those decisions are made and the way forward is clarified.

hackmastera:
close in favor of  #2773

PR #1034, opened 2018-04-09
Add bag validations properties, job

hackmastera:
refs #752; this is probably as far as we can go on this ticket until we get some more concrete answers regarding where the bag files will live and whether/how they will be validated.

Note we cannot run this job in staging or prod until https://github.com/LibraryOfCongress/bagit-python is installed.

Actually CI may also fail for that reason; I guess we'll find out!


hackmastera:
closing; see https://github.com/pulibrary/figgy/issues/752#issuecomment-382474729

Issue #1031, opened 2018-04-09
BookPlates should have a simple draft/published workflow


escowles:

hackmastera:
It will be identical to the folder workflow, except for names of things.

PR #1039, opened 2018-04-10
Add a simple draft-publish workflow for use with plates (SimpleResource), refs #1031

hackmastera:
It's nearly identical to the folder workflow; the only difference is the names of the states / transitions.

### DIFF
--- a/app/models/draft_publish_workflow.rb
+++ b/app/models/draft_publish_workflow.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# Simple draft / published state-based workflow: Start as draft (will be
+# visible, but not able to view the manifest.) When published, the
+# manifest is visible.
+class DraftPublishWorkflow
+  include AASM
+
+  def initialize(state)
+    aasm.current_state = state.to_sym unless state.nil?
+  end
+
+  aasm do
+    state :draft, initial: true
+    state :published
+
+    # ingest workflow
+    event :publish do
+      transitions from: :draft, to: :published
+    end
+    event :unpublish do
+      transitions from: :published, to: :draft
+    end
+  end
+
+  # this workflow doesn't have a suppressed state but other code checks for it; make it false
+  def suppressed?
+    false
+  end
+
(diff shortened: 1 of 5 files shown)


app/models/draft_publish_workflow.rb
```diff
@@ -0,0 +1,37 @@
...
+    event :publish do
+      transitions from: :draft, to: :published
+    end
+    event :unpublish do
+      transitions from: :published, to: :draft
+    end
+  end
+
+  def suppressed?
+    false
```
escowles: Shouldn't this check `draft?` ?

hackmastera: Hm, I didn't really see this used so I had tried removing it but it clearly needs more attention.

In folder workflow, which I based this on, it's always false. Do you know how this works @escowles ?
https://github.com/pulibrary/figgy/blob/master/app/models/folder_workflow.rb#L25-L27

hackmastera: here it seems to be an optional other state: https://github.com/pulibrary/figgy/blob/master/app/models/book_workflow.rb#L5

I'm not sure it's' used in indexing...

hackmastera: For access we need something like this https://github.com/pulibrary/figgy/blob/master/app/indexers/ephemera_folder_indexer.rb#L19

hackmastera: I added the false back in with a comment that I think is accurate. Also created #1040.

escowles: I'm not sure if the `suppressed?` method is really being used for anything, so maybe we can keep as-is, and use #1040 to explore whether we need it, or if there is some refactoring to do between indexers and `Ability` to correctly suppress unpublished items?

hackmastera: sounds good. I'll add a note to #1040

hackmastera:
This is ready for review; I think we'll need it in before we merge the new model.

PR #1065, opened 2018-04-16
Push state-based visibility logic into workflow classes and decorators

hackmastera:
Refactor `Ability` and `EphemeraFolderIndexer` to call the new logic instead of duplicating with one another.

refs @1040


hackmastera:
I believe this will pass and is ready for review.

PR #1070, opened 2018-04-17
Ensures that links for editing and deleting resources are only rendered for users with the appropriate permissions

### LINKED ISSUE #795: Hide edit/delete links from show page if you don't have permission

jrgriffiniii:
Also encompasses permissions for:
* file management
* editing the logical structure
* downloading derivative files in FileSets

Resolves #795

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -12,4 +12,32 @@ module CatalogHelper
     val = Array.wrap(presenter(document).heading).map { |v| h(v) }.join("<br />")
     content_tag(tag, val, { itemprop: "name", dir: val.to_s.dir }, false)
   end
+
+  def can_edit?
+    can? :update, resource
+  end
+
+  def can_manage_files?
+    can? :file_manager, resource
+  end
+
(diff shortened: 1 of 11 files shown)


app/helpers/catalog_helper.rb
```diff
@@ -12,4 +12,32 @@ def render_document_heading(*args)
     val = Array.wrap(presenter(document).heading).map { |v| h(v) }.join("<br />")
     content_tag(tag, val, { itemprop: "name", dir: val.to_s.dir }, false)
   end
+
+  def can_edit?
```
tpendragon: What's the benefit of having these in helpers rather than just having `can?(:update, resource)` in the view?

jrgriffiniii: Slightly more complex cases (e. g. `can?(:edit, resource) && can?(:create, EphemeraFolder)`) felt to cluttered for the View layer, but I can shift these into the View template

jrgriffiniii:
This should probably block https://github.com/pulibrary/figgy/pull/1069

PR #1071, opened 2018-04-18
Upgrade chromedriver-helper to resolve circle-ci failure

hackmastera:
`session not created exception: Chrome version must be >= 65.0.3325.0`


hackmastera:
resolved by #11077

Issue #1006, opened 2018-04-03
Create / import screen for ArchivalMediaCollection


hackmastera:
User gives EAD Collection ID and bag location.

A collection is created/accessed.

Metadata for the collection should be imported from the EAD.

Bag import job is called.

It must be possible to repeat this process with multiple bags.

see https://drive.google.com/drive/u/0/folders/1WsYZH7o-Fjoyr6xJ7KDFhqPN01oubCfi

hackmastera:
This will be a form that call initiates a job (tbd in 753). Moving preview screen to another issue.

PR #1089, opened 2018-04-19
Add controller, link to create Archival Media Collection

### LINKED ISSUE #1087: Add ArchivalMediaCollection to front page

hackmastera:
resolves #1087

Currently linked to a form that renders Collection objects; that should
be addressed as part of #1006

### DIFF
--- a/spec/controllers/archival_media_collections_controller_spec.rb
+++ b/spec/controllers/archival_media_collections_controller_spec.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe ArchivalMediaCollectionsController do
+  let(:user) { nil }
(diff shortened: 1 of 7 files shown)


spec/controllers/archival_media_collections_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  before do
+    sign_in user if user
+  end
+
+  context "when an admin" do
+    let(:user) { FactoryBot.create(:admin) }
+
+    describe "GET /collections/new" do
+      render_views
+      # This creates a Collection for now; this should change as part of work for #1006
```
escowles: I'm not sure I understand why this has to create a Collection — can't ArchivalMediaCollectionsController set the resource class to create an ArchivalMediaCollection instead?

hackmastera: Only because I didn't know to do that. i'll upate.

PR #1095, opened 2018-04-19
Do Not Merge: Fixes Filemanager for Geo

### LINKED ISSUE #785: Fix / Implement File Manager for Geo
Bring back the old file manager for geo objects or adapt the existing one.

sdellis:
Fixes #785 by splitting out FileManager and OrderManager into separate functions. Tidies up Filemanager.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -4,7 +4,7 @@ class Ability
   # Define any customized permissions here.
   def custom_permissions
     alias_action :show, :manifest, to: :read
-    alias_action :color_pdf, :pdf, :edit, :browse_everything_files, :structure, :file_manager, to: :modify
+    alias_action :color_pdf, :pdf, :edit, :browse_everything_files, :structure, :file_manager, :order_manager, to: :modify
     roles.each do |role|
       send "#{role}_permissions" if current_user.send "#{role}?"
     end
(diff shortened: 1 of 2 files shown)


escowles:
@sdellis I think it would be helpful to create a table listing the purpose of the file/order/structure managers for each work type.  Because for books, the file and order managers seem like they have a lot of overlap.  So I think it might help us think about whether having all three makes sense for all work types.

sdellis:
@escowles here is the reasoning behind this move:

We have learned recently that one class of users is only concerned with uploading and downloading files (i.e., digital studios, photoduplication staff). They have no need to reorder or relabel, and some of these users should not be allowed to do so.

Another class of users deals primarily with reordering, relabeling, and adding structure. As expressed in the standup, I would like to merge the reordering, relabeling, and structuring into a "Presentation Manager", thus merging the structure and order managers. That is too epic a change at the moment. 

Many work types will require the management of a variety of files, many of which should not be expressed as canvases in a IIIF manifest. This is true for Vector and Raster work types. These filesets contain extra files that can and should be expressed differently in the manifest. They also don't require any ordering or structuring. Relabeling can be done via the Fileset edit screens, and they don't require automated labeling. The point of this PR was to simplify and converge on a model that works for all work types, which this does. 

This PR provides a way to manage the permissions and create workflows that are aligned with user roles. You will notice that in this PR, we have not removed the option to upload or get to the Fileset edit screens in the Order Manager. I believe that should be removed, but wanted to leave it open for now.

Does that answer your question enough to merge? Or do we need to post to staging and get feedback from other stakeholders?

escowles:
@sdellis Yes, that's very helpful.  I do remember now that we discussed merging the Order and Structure managers.  I would like to get feedback on this direction before merging this PR, and will try to schedule a session next week.

escowles:
Closing in favor of #1138

PR #1103, opened 2018-04-20
Fix collection manifests with MVWs.

tpendragon:
When collection manifests have multi-volume works the thumbnail_id of
the MVW points to the scanned resource within it that has the thumbnail
it should use, rather than the FileSet. The thumbnail builder was
assuming it'd get back a FileSet, which caused an error to be thrown
while embedding thumbnails in those manifests. The error can be seen
here: https://app.honeybadger.io/projects/53391/faults/37473184

The change here iterates down thumbnail_ids until it finds a FileSet,
and if it doesn't it just doesn't append a thumbnail to the manifest.

### DIFF
--- a/app/services/manifest_builder/thumbnail_builder.rb
+++ b/app/services/manifest_builder/thumbnail_builder.rb
@@ -33,9 +33,19 @@ class ManifestBuilder
       end
 
       def file_set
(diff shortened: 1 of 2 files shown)


hackmastera:
@tpendragon can you update the commit message to describe the code change itself, and maybe describe what was broken / fixed in the PR description?

PR #1113, opened 2018-04-23
Maintain newlines when saving text version of HOCR

tpendragon:


tpendragon:
@hackmastera I think I took care of it. Had to refactor a bit to make mocking reasonable.

hackmastera:
Awesome, thank you.

PR #1121, opened 2018-04-23
Allow editing of both title and source metadata id on geo resources

### LINKED ISSUE #1120: Use portion note to differentiate between child geo resources that share a bibid
This is mainly to allow easier editing of title for nested work type such as MapSets. For example, a top-level scanned map might look like this:

```
title: 'Christopher and his kind, 1929-1939'
source_metada_id: '1234567'
```

The child sheet of the MapSet will share the same source metadata and  imported metadata, but will have a slightly modified title:

```
title: 'Christopher and his kind, 1929-1939 (Sheet 1)'
source_metada_id: '1234567'
```

- [ ] Add portion note to geo change sets
- [ ] Create / update decorator method that concats portion note value with title for display

eliotjordan:
Closes #1120


eliotjordan:
After discussion with @tpendragon, it appears the correct path is to use part note to differentiate between nested resources that share a bibid. Closing in favor of that approach.

PR #1137, opened 2018-04-24
[WIP] Implements `FindUnrelated` and `FindUnrelatedParents` Solr queries for populating the list of attachable and parent resources

### LINKED ISSUE #1100: Query Solr for relating `ScannedMaps`, `RasterResources`, and `VectorResources`
Discovered when attempting to resolve #1069 using PostgreSQL queries

jrgriffiniii:
Resolves #1100

### DIFF
--- a/app/assets/javascripts/relationships/parent_resources_table.es6
+++ b/app/assets/javascripts/relationships/parent_resources_table.es6
@@ -8,8 +8,14 @@ export default class ParentResourcesTable extends RelatedResourcesTable {
   constructor(element, form) {
     super(element, form)
     this.attribute = 'member_ids'
+    this.members = []
+  }
+
+  bindTable(element) {
+    super.bindTable(element)
     this.parents = this.table.data('parents')
     this.resourceId = this.table.data('resource-id')
+    this.parents = this.table.data('parents')
   }
 
   /**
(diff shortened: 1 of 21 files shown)


eliotjordan:
The code behind this is solid, and the loading time is 10x faster using solr queries. I'm still a little concerned overall about using title to search for scanned map members. For example, the only way to distinguish between members in new map sets is to use the portion note (map sets migrated from plum don't currently use portion note). Also, while the performance is much better, I still wonder about a dropdown that has 20,000 rows. Maybe someone else has comments and would like to weigh in?

escowles:
@eliotjordan I think it's worth talking through the workflow to see if there's some other pattern that makes sense.  The top-down approach used for books is one pattern, but if that doesn't fit, then we can explore other patterns here.  For example, is there a query that would get all of the members that you wanted to add to a MapSet?  If so, then doing that query in the normal search interface, selecting the items to add with checkboxes or something, might be a better way to find the items.  Or maybe using the bookmarks feature to collect resources, and having the dropdown menus populated from the user's bookmarks.

jrgriffiniii:
Pending further discussions with the stakeholders for ingesting and managing geospatial assets, work shall be undertaken which may be substantially different from what is offered on this branch.  Hence, it is no longer necessary that this remain open.

PR #1153, opened 2018-05-01
Ingest archival media bags, fixes #753

### LINKED ISSUE #753: Job to ingest a bag of A/V objects
A MediaResource object is created for each component id, with the appropriate title (drawn from EAD finding aid) and appropriate files attached.

hackmastera:


app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -0,0 +1,227 @@
+# frozen_string_literal: true
+class IngestArchivalMediaBagJob < ApplicationJob
+  BARCODE_WITH_PART_REGEX = /(\d{14}_\d+)_.*/
+
+  def perform(collection_component:, bag_path:, user:)
+    amc = find_or_create_amc(collection_component)
```
tpendragon: Do we want to move this into the buffer_into_index part below? Otherwise if something goes wrong we're left with an AMC object created.

app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -0,0 +1,227 @@
...
+          Valkyrie::Vocab::PCDMUse.PreservationMasterFile
+        elsif intermediate?
+          Valkyrie::Vocab::PCDMUse.IntermediateFile
+        elsif access?
+          Valkyrie::Vocab::PCDMUse.ServiceFile
+        end
+      end
+
+      def master?
+        path.fnmatch "*_pm.wav"
```
tpendragon: I'm not sure if this is important, but I expect `path.to_s.end_with?("_pm.wav")` to be faster.

app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -0,0 +1,227 @@
...
+      attr_reader :collection, :bag, :user, :adapter
+      def initialize(collection:, bag:, user:, adapter:)
+        @collection = collection
+        @bag = bag
+        @user = user
+        @adapter = adapter
+      end
+
+      def ingest
+        component_dict = BarcodeComponentDict.new(collection)
```
tpendragon: Extract to instance method

app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -0,0 +1,227 @@
...
+        component_groups = bag.component_groups(component_dict)
```
tpendragon: Extract to instance method

app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -0,0 +1,227 @@
...
+
+        component_groups.each do |cid, sides|
+          media_resource = find_or_create_media_resource(cid)
+          sides.each do |side|
+            file_set = create_av_file_set(side)
+            media_resource.member_ids += Array.wrap(file_set.id)
+          end
+          media_resource.member_of_collection_ids += Array.wrap(collection.id)
+          persister.save(resource: media_resource)
```
tpendragon: Do we want to save these as ChangeSets through a ChangeSetPersister? Fixity jobs and stuff won't run this way, but I'm not sure if that's a problem

tpendragon: I don't think metadata will get imported either this way.

app/models/media_resource.rb
```diff
@@ -5,7 +5,7 @@ class MediaResource < Valhalla::Resource
   include Schema::Common
   attribute :id, Valkyrie::Types::ID.optional
   attribute :member_ids, Valkyrie::Types::Array
-  attribute :member_of_collection_ids
+  attribute :member_of_collection_ids, Valkyrie::Types::Array
```
tpendragon: Is it important this is an array?

spec/jobs/ingest_archival_media_bag_job_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect(file_set.file_metadata.map { |file| file.use.first.to_s }).to contain_exactly(
+        "http://pcdm.org/use#PreservationMasterFile", "http://pcdm.org/use#ServiceFile", "http://pcdm.org/use#IntermediateFile"
+      )
+    end
+
+    it 'creates one MediaResource per component id' do
+      expect(query_service.find_all_of_model(model: MediaResource).size).to eq 1
+    end
+
+    it 'for each compopnent id-based MediaRsource, puts it on the collection' do
```
tpendragon: Small typo - "component"

tpendragon: And "MediaResource"

app/jobs/ingest_archival_media_bag_job.rb
hackmastera: @tpendragon just like `component_dict = get_component_dict(collection)`?

app/models/media_resource.rb
hackmastera: when it's not I have problems adding to it because it gets initialized as nil. maybe using change sets will help? I'll try taking it out once I've got the changeset stuff in place.

app/jobs/ingest_archival_media_bag_job.rb
tpendragon: Sorry - something like
```
def component_dict
  @component_dict ||= get_component_dict(collection)
end
```

app/models/media_resource.rb
tpendragon: Oh, that's because the default is a `Valkyrie::Types::Set.optional`. `Valkyrie::Types::Set` will also work.

spec/jobs/ingest_archival_media_bag_job_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    # We need to actually save it with persister to get the imported metadata
+    let(:collection) do
+      cs = DynamicChangeSet.new(ArchivalMediaCollection.new)
+      cs.validate(source_metadata_identifier: "C0652", bag_path: bag_path)
+      cs.sync
+      change_set_persister.save(change_set: cs)
+    end
+
+    before do
+      stub_request(:get, "https://findingaids.princeton.edu/collections/C0652.xml?scope=record")
```
tpendragon: Small thing, but can you just use `stub_pulfa` from https://github.com/pulibrary/figgy/blob/master/spec/support/stub_bibdata.rb#L20 ?

PR #1165, opened 2018-05-01
Ingest bag when a bag path is available.

### LINKED ISSUE #1149: Call IngestArchivalMediaBag from ArchivalMediaCollectionController
This job should be called in after_create and after_update in the ArchivalMediaCollectionController

tpendragon:
When something's ingested which responds to bag_path and it's a
non-empty value it will run the bag ingest job.

I went with this way instead of the proposed after_create/after_update hooks so that we could also ingest via the console or a rake task.

Closes #1149


tpendragon:
I tried to the test bag now that this was wired up, but it failed to parse a title from the C0652 record on PULFA.

tpendragon:
I can confirm that this ingests now. There are bugs with displays, but I created tickets for them (#1176, #1177, #1178)

PR #1169, opened 2018-05-02
Update metadata import for ArchivalMediaCollection

tpendragon:
Fixes fixture to use the ?scope=record format, updates
PulMetadataServices to a version that can handle it, and adds a test to
ensure metadata is extracted.


tpendragon:
Closing this PR - the record scope isn't going to work, it doesn't have the barcode information we need to import.

PR #1173, opened 2018-05-02
Update metadata import for ArchivalMediaCollection

tpendragon:
Fixes fixture to use the ?scope=record format, updates
PulMetadataServices to a version that can handle it, and adds a test to
ensure metadata is extracted.


hackmastera:
Yay, thanks @tpendragon! I think the first comment of this PR has a typo; should say "fixes fixture to use the ?scope=source format"

tpendragon:
@hackmastera Really? I think it's using ?scope=record

hackmastera:
@tpendragon I'm confused then. why would the fixture use record scope when you did all that work to allow source scope?

PR #1181, opened 2018-05-03
Display AMC/MediaResources in search

### LINKED ISSUE #1177: Don't hide ArchivalMediaCollections and MediaResources from catalog search
Right now they don't show up when they're ingested and you search. Need to modify SearchBuilder#models_to_solr_clause

tpendragon:
Closes #1177


app/models/search_builder.rb
```diff
@@ -34,7 +34,7 @@ def readable_states
   end
 
   def models_to_solr_clause
-    [ScannedResource, Collection, EphemeraFolder, EphemeraBox, ScannedMap, VectorResource, RasterResource, SimpleResource].join(",")
+    [ScannedResource, Collection, EphemeraFolder, EphemeraBox, ScannedMap, VectorResource, RasterResource, SimpleResource, ArchivalMediaCollection, MediaResource].join(",")
```
escowles: As this gets longer, I wonder if we should reverse the logic and list models we don't want to show up?

tpendragon: Yeah, I think we only want to hide...what, FileSet?

escowles: I think the ones missing from that list are FileSet, EphemeraProject, EphemraVocabulary, and EphemeraTerm.

tpendragon: Do you want me to do that now?

escowles: Yes, I think it would be good to do now.  If it causes a lot of test failures or something, feel free to make it a separate ticket.

tpendragon: Done!

PR #1206, opened 2018-05-08
Update imported ARKs instead of minting duplicates

### LINKED ISSUE #1180: ScannedResource with both imported ARK and newly-minted ARK
Some records with imported metadata have both an ARK imported from Voyager, and also have a newly-minted ARK from EZID.  Example record: https://figgy.princeton.edu/catalog/5cb25d6a-4bf3-4b29-9e60-0482f946afa3

These items do not display a viewer in the Catalog because they are not retrieved by a query for the imported ARK.

escowles:
Fixes #1180

### DIFF
--- a/app/services/identifier_service.rb
+++ b/app/services/identifier_service.rb
@@ -1,16 +1,23 @@
 # frozen_string_literal: true
 class IdentifierService
   def self.mint_or_update(resource:)
-    if resource.identifier.present?
+    if identifier_for(resource).present?
       update_metadata resource
     else
       mint_identifier resource
     end
   end
 
+  private_class_method def self.identifier_for(resource)
+    return resource.identifier if resource.identifier.present?
+    return unless resource.respond_to? :imported_metadata
+    ark_url = resource.imported_metadata&.first&.identifier&.first
+    ark_url&.gsub(/.*ark:/, "ark:")
+  end
+
   private_class_method def self.update_metadata(resource)
     return if minter_user == "apitest"
-    minter.modify(Array.wrap(resource.identifier).first, metadata(resource))
+    minter.modify(Array.wrap(identifier_for(resource)).first, metadata(resource))
   end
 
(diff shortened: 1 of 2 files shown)


app/services/identifier_service.rb
```diff
@@ -1,16 +1,23 @@
 # frozen_string_literal: true
 class IdentifierService
   def self.mint_or_update(resource:)
-    if resource.identifier.present?
+    if identifier_for(resource).present?
```
tpendragon: Something for the future - maybe we should look into making this an instance now, if we're calling private class methods?

app/services/identifier_service.rb
```diff
@@ -1,16 +1,23 @@
...
+    return resource.identifier if resource.identifier.present?
+    return unless resource.respond_to? :imported_metadata
+    ark_url = resource.imported_metadata&.first&.identifier&.first
+    ark_url&.gsub(/.*ark:/, "ark:")
+  end
+
   private_class_method def self.update_metadata(resource)
     return if minter_user == "apitest"
-    minter.modify(Array.wrap(resource.identifier).first, metadata(resource))
+    minter.modify(Array.wrap(identifier_for(resource)).first, metadata(resource))
```
tpendragon: I have some concern that using identifier_for instead of like, overwriting the local identifier with the imported one will mean that we'll have a local identifier property which shows up in the UI but is never actually used for anything and never updated.

escowles: I thought about the precedence here, and decided to have the top-level `identifier` override the imported one made sense to prevent this.  If we mint and identifier, then we use it.  If not, then we fall back on the imported one.

I'd be fine with just copying the imported identifier to the top-level one, though that seemed like it was crossing a boundary between local and imported metadata that I wasn't sure we wanted to cross.

escowles:
Closing in favor of #1207

PR #1207, opened 2018-05-08
Copy imported identifiers to the resource to prevent minting duplicate ARKs

### LINKED ISSUE #1180: ScannedResource with both imported ARK and newly-minted ARK

escowles:
Fixes #1180 
Alternative to #1206


tpendragon:
I'm :+1: to this implementation over #1206, personally.

escowles:
@tpendragon That's fine with me — it does keep the implementation located in the remote metadata code and leaves the identifier service simpler.  Ready to merge?

PR #1213, opened 2018-05-08
Adds static file to serve UV v3 for embedding elsewhere

sdellis:
iframe source would be: 
`/uv.html#?manifest={manifest}`

Specifying start time/canvas parameters can be tacked on as well, via [UV Hash Parameters](https://github.com/UniversalViewer/universalviewer/wiki/Embedding#hash-parameters)


escowles:
Could we move `uv.html` to `uv/index.html` so the URL would be `/uv?manifest=...` ?

escowles:
OK, thanks @sdellis

PR #1216, opened 2018-05-09
Add 'Ephemera box' field to folder edit form, fixes #1143

### LINKED ISSUE #1143: Ephemera: Allow moving boxless folders into boxes
This is to support recording the boxes for the ~4K posters migrated into Figgy as boxless.  Important to make this as efficient as possible, given the volume of manual work.

Ideal workflow:
1. Search on poster title
2. Edit poster
3. Select box on edit form

hackmastera:
screen shot 2018-05-09 at 1 59 48 pm

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -69,6 +69,7 @@ Metrics/ClassLength:
     - 'app/services/manifest_builder/manifest_service_locator.rb'
     - 'app/services/pdf_generator/cover_page_generator.rb'
     - 'app/decorators/valkyrie/resource_decorator.rb'
+    - 'app/controllers/ephemera_folders_controller.rb'
 Metrics/MethodLength:
   Exclude:
     - 'db/migrate/**/*'
--- a/app/controllers/ephemera_folders_controller.rb
+++ b/app/controllers/ephemera_folders_controller.rb
@@ -8,6 +8,7 @@ class EphemeraFoldersController < BaseResourceController
   )
   before_action :load_fields, only: [:new, :edit, :update, :create]
   before_action :cache_box, only: [:destroy]
+  before_action :load_boxes, only: [:edit]
 
   def change_set_class
     @change_set_class ||= parent_resource.is_a?(EphemeraBox) ? EphemeraFolderChangeSet : BoxlessEphemeraFolderChangeSet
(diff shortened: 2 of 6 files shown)


app/controllers/ephemera_folders_controller.rb
```diff
@@ -88,13 +92,40 @@ def load_fields
...
+      []
+    end.unshift(nil_box)
+  end
+
+  def ephemera_box
+    parent_resource.is_a?(EphemeraBox) ? parent_resource : nil
+  end
+
+  def nil_box
+    Struct.new(:title, :id).new('', '')
```
tpendragon: What's the nil_box get us that the empty array doesn't?

hackmastera: I actually don't think I tested the empty array case. When you have a non-empty array you need a nil option to keep the form from filling in the first value of the array, and allowing the user to not fill in the field if they don't want to.

app/controllers/ephemera_folders_controller.rb
```diff
@@ -88,13 +92,40 @@ def load_fields
     end
   end
 
+  # provides instance variables needed to populate collection and selected for append_id
+  # @see app/views/ephemera_folders/edit_fields/_append_id.html.erb
+  def load_boxes
+    instance_variable_set("@available_boxes", available_boxes)
```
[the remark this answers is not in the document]
hackmastera: I must have seen it floating around somewhere else and gotten it stuck in my head; I will change.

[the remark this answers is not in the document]
hackmastera: oh, in the method above.

.rubocop.yml
```diff
@@ -69,6 +69,7 @@ Metrics/ClassLength:
     - 'app/services/manifest_builder/manifest_service_locator.rb'
     - 'app/services/pdf_generator/cover_page_generator.rb'
     - 'app/decorators/valkyrie/resource_decorator.rb'
+    - 'app/controllers/ephemera_folders_controller.rb'
```
escowles: For next week: maybe we should increase the class length? Or look at these classes and see if there's any refactoring we want to do?

PR #1262, opened 2018-05-15
Sync change set with model before saving

### LINKED ISSUE #394: Add change_set.sync to change set persister save method
This would make the behavior more like reform's, wherein saving the change set first syncs to the model. It would provide a convenience; not sure there's a use case for saving without syncing first.

eliotjordan:
Closes #394


eliotjordan:
@hackmastera That's a good idea. I'll push up a commit removing unneeded sync calls.

eliotjordan:
Updated.

tpendragon:
Nice!

PR #1265, opened 2018-05-15
Simplify ability/permissions by replacing a bunch of roles with 'staff'

### LINKED ISSUE #650: Simpler permissions
We should have 4 roles:
* admin: can do anything
* staff: can do anything except
  * delete other people's content
  * delete anything with an ARK
* campus_patron: can see public and princeton-only items
* anonymous: can see public items

hackmastera:
Fixes #650

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -3,7 +3,7 @@ class Ability
   include Hydra::Ability
   # Define any customized permissions here.
 
-  self.ability_logic +=[:manifest_permissions]
+  self.ability_logic += [:manifest_permissions]
 
   def custom_permissions
     alias_action :show, to: :read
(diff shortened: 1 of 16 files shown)


app/models/ability.rb
```diff
@@ -15,28 +15,20 @@ def admin_permissions
...
+    end
+    can [:destroy], curation_concerns do |obj|
+      obj.depositor == [current_user.uid]
+    end
   end
 
-  def curation_concern_read_permissions
+  def anonymous_permissions
+    # do not allow viewing incomplete resources
     cannot [:read], curation_concerns do |resource|
```
escowles: Can you reverse the logic of these two `cannot` blocks to be `can :read readable_concern?` etc.? I think that would be easier to read, and avoid the mixing of `can` and `cannot` which we thought might cause problems.

Also, it looks like `anonymous_permissions` is only called from the `campus_patron_permissions` below — maybe we should just fold this into that method?

hackmastera: `anonymous_permissions` is also called dynamically in `custom_permissions`

I will look into reversing the `cannot` logic.

PR #1272, opened 2018-05-16
Fix original file and derivative deletion

### LINKED ISSUE #981: Derivative files are not deleted when resource is deleted
Derivative files remain on the file system when their parent resource is deleted. Calling `delete` on the storage adapter in the derivative service does not delete the file.

https://github.com/pulibrary/figgy/blob/master/app/derivative-services/jp2_derivative_service.rb#L82

eliotjordan:
Trying this one again.

Closes #981


app/change_set_persisters/change_set_persister/cleanup_files.rb
```diff
@@ -0,0 +1,29 @@
...
+      CleanupFilesJob.perform_later(file_identifiers: identifiers_to_remove)
+    end
+
+    private
+
+      # Returns the file identifiers for the original file and
+      # each derivative attached to a file set.
+      def identifiers_to_remove
+        identifiers = []
+        files = resource.file_metadata
```
tpendragon: Can this just be
```ruby
resource.file_metadata.flat_map do |file|
  file.try(:file_identifiers)
end.compact.map(&:to_s)
```
? If that's more readable, at least. Out of curiosity - why the try? Do we ever have things in file metadata that don't respond to file_identifiers?

If not, then

```ruby
resource.file_metadata.flat_map(&:file_identifiers).compact.map(&:to_s)
```

app/derivative-services/vector_resource_derivative_service.rb
```diff
@@ -51,7 +51,11 @@ def cleanup_derivatives
     deleted_files = []
     vector_derivatives = resource.file_metadata.select { |file| file.derivative? || file.thumbnail_file? }
     vector_derivatives.each do |file|
-      storage_adapter.delete(id: file.id)
+      file_identifier = file.file_identifiers.try(:first)
```
tpendragon: Maybe we should make `file_identifiers` not an optional array so we can just trust calling `.first`?

app/jobs/cleanup_files_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class CleanupFilesJob < ApplicationJob
+  delegate :query_service, to: :metadata_adapter
+
+  def perform(file_identifiers:)
+    file_identifiers.each do |file_identifier|
+      storage_adapter = Valkyrie::StorageAdapter.adapter_for(id: file_identifier)
+      # Delete the entire directory to remove unzipped derivatives like display vectors
+      id = File.dirname(file_identifier.to_s)
```
tpendragon: This works for now, but feels like it relies on the delete implementation of the disk storage adapter. I'm not sure what the alternative is...maybe something in Valkyrie?

eliotjordan: I had the same thought, but I couldn't think of a method for handling the more general case. I think in the future we'll load vector data into Postgres/PostGIS or use a single file format (like GeoPackage) and this won't be an issue.

eliotjordan:
Updated with requested changes.

PR #1276, opened 2018-05-17
Integrating support for propagating state to member resources within `ChangeSetPersister::PropagateVisibilityAndState`

### LINKED ISSUE #1240: Refactor ephemera folder publication
Propose: Instead of logic which treats folders as published either if it's complete or its parent is complete, actually propagate the workflow state down to child objects. This would allow us to simplify indexing and ability logic, and also unblock https://github.com/pulibrary/figgy/issues/565

jrgriffiniii:
Resolves #1240

### DIFF
--- a/app/change_set_persisters/change_set_persister/propagate_visibility_and_state.rb
+++ b/app/change_set_persisters/change_set_persister/propagate_visibility_and_state.rb
@@ -12,23 +12,37 @@ class ChangeSetPersister
       return if !change_set.changed?(:visibility) && !change_set.changed?(:state)
       members.each do |member|
         member.read_groups = change_set.read_groups if change_set.read_groups
-        propagate_state(member)
+        propagate_state_for_related(member)
         persister.save(resource: member)
       end
     end
 
-    def state_for_related(related)
-      value = DynamicChangeSet.new(related).state_for_related change_set.model, change_set.state
-      value.to_s
+    def related_workflow(related_resource)
+      related_resource_state = Array.wrap(related_resource.state).first
+      DynamicChangeSet.new(related_resource).workflow_class.new(related_resource_state)
     end
 
-    def should_set_state_for_related?(resource)
-      change_set.state && resource.respond_to?(:state) && valid_states(resource).include?(state_for_related(resource))
+    def valid_states(member)
+      related_workflow(member).valid_states
+    end
+
+    def changed_workflow
+      changed_state = Array.wrap(change_set.state).first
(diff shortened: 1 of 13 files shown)


spec/change_sets/scanned_resource_change_set_spec.rb
```diff
@@ -98,4 +98,20 @@
       expect(change_set.logical_structure[0].label).to eq ["Logical"]
     end
   end
+  context "when an EphemeraBox has EphemeraFolder members" do
```
eliotjordan: Should this be "when a ScannedResource has Scanned Resource members"?

app/models/folder_workflow.rb
```diff
@@ -22,12 +16,25 @@ def initialize(state)
...
+        received: :needs_qa,
+        all_in_production: :complete
+      }
+    }
   end
 
-  def valid_transitions
-    aasm.states(permitted: true).map(&:name).map(&:to_s)
+  # Retrieve the state for the resource
+  # @param related_resource [Valhalla::Resource] the resource related to the current workflow
```
eliotjordan: Parameter is the resource class not the resource itself.

jrgriffiniii:
I feel that I've addressed some of my concerns by decoupling Workflow state translations from Resource Classes (I overlooked that).

jrgriffiniii:
a9058e2 attempts to implement a simplified approach proposed by @tpendragon when discussing how best to refactor this further.

valhalla/app/change_sets/valhalla/change_set.rb
```diff
@@ -7,6 +7,13 @@ def self.apply_workflow(workflow)
       include(Valhalla::ChangeSetWorkflow)
     end
 
+    def state_for_related(related_resource, related_state)
```
hackmastera: where is this used?

app/change_set_persisters/change_set_persister/propagate_visibility_and_state.rb
```diff
@@ -12,13 +12,37 @@ def run
       return if !change_set.changed?(:visibility) && !change_set.changed?(:state)
       members.each do |member|
```
hackmastera: can you add a class-level comment explaining how the logic works?

jrgriffiniii:
This currently blocks the resolution for #1279 (please see https://github.com/pulibrary/figgy/tree/issues-1279-jrgriffiniii-folders-migration).  Once this is acceptable for merging, I can rebase that branch and issue a separate Pull Request.

PR #1277, opened 2018-05-17
refactor ark

christinach:


app/values/ark.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+module Ark
```
tpendragon: Instead of doing a module include, maybe a value object? Something like

```ruby
class Ark
  attr_reader :ark
  def initialize(ark:)
    @ark = ark
  end

  # Should return ark://bla
  def identifier
  end

  # Should return https://arks.princeton.edu/bla
  def uri
  end
end
```

christinach: @tpendragon thank you for the suggestion.

PR #1278, opened 2018-05-17
Documenting the storage adapter integration from Valkyrie

### LINKED ISSUE #355: Document StorageAdapters
Documenting what lies within https://github.com/pulibrary/figgy/blob/master/config/initializers/valkyrie.rb#L5, including the use cases outlining precisely why these have been included (e. g. `disk` moves files).

jrgriffiniii:
Resolves #355

### DIFF
--- a/config/initializers/valkyrie.rb
+++ b/config/initializers/valkyrie.rb
@@ -1,13 +1,18 @@
 # frozen_string_literal: true
 require_relative "figgy"
+
 Rails.application.config.to_prepare do
+  # Registers a storage adapter for a *NIX file system
+  # Binaries are persisted by invoking "mv" with access limited to read/write for owning users, and read-only for all others
+  # NOTE: "mv" may preserve the inode for the file system
+  # file_mover should be a lambda or Proc which performs an operation on the file using its path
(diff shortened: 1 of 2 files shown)


config/initializers/valkyrie.rb
```diff
@@ -1,20 +1,30 @@
 # frozen_string_literal: true
 require_relative "figgy"
+
 Rails.application.config.to_prepare do
+  # Registers a storage adapter for a *NIX file system
+  # Binaries are persisted by invoking "mv" with access limited to read/write for owning users, and read-only for all others
+  # NOTE: "mv" may preserve the inode for the file system
```
tpendragon: Can we add the use case for why `mv`? (Which is that this is typically used for pulling uploads from a temporary location such as a file upload from a form)

tpendragon: Also why the chmod is important (if I remember, it was moving files to the derivative location and preserving the permissions and the image server couldn't read them.)

tpendragon:
@jrgriffiniii Looks good - I think it just needs a rebase.

PR #1306, opened 2018-05-25
Extract bibid from directory name during bulk ingest

### LINKED ISSUE #1305: Bulk ingest service should attempt to grab bibid from directory name
It would be useful if the bulk ingest service would attempt to grab a bibid from the name of the directory it is ingesting; unless the `source_metadata_id` is set in the attribute parameters.

eliotjordan:
Closes #1305


tpendragon:
I'm concerned this will assign a bib id to child volumes if it runs into a MVW. Will it?

eliotjordan:
@tpendragon No. That is prevented here: https://github.com/pulibrary/figgy/blob/ca5fb5599937818b35f2339609a8afaa6204209c/app/services/bulk_ingest_service.rb#L41

PR #1311, opened 2018-05-25
Updates active sort fields in a collection based on the current document

### LINKED ISSUE #1218: On collection members list, sort_by "refresh" button leads to error

christinach:
fixes #1218

### DIFF
--- a/app/views/catalog/_members_archival_media_collection.html.erb
+++ b/app/views/catalog/_members_archival_media_collection.html.erb
@@ -6,7 +6,7 @@
       <fieldset class="col-xs-12">
         <legend class="sr-only"><%= t('sort_label') %></legend>
         <%= label_tag(:sort, "<span>Sort By:</span>".html_safe) %>
-        <%= select_tag(:sort, options_for_select(active_sort_fields, h(params[:sort]))) %>
+        <%= select_tag(:sort, options_for_select(active_sort_fields.keys, h(params[:sort]))) %>
         <%= label_tag(:per_page) do %>
           Show <%= select_tag(:per_page, options_for_select(['10', '20', '50', '100'], h(params[:per_page])), title: "Number of results to display per page") %>
           per page
--- a/app/views/catalog/_members_collection.html.erb
+++ b/app/views/catalog/_members_collection.html.erb
@@ -5,7 +5,7 @@
       <fieldset class="col-xs-12">
         <legend class="sr-only"><%= t('sort_label') %></legend>
         <%= label_tag(:sort, "<span>Sort By:</span>".html_safe) %>
-        <%= select_tag(:sort, options_for_select(active_sort_fields, h(params[:sort]))) %>
+        <%= select_tag(:sort, options_for_select(active_sort_fields.keys, h(params[:sort]))) %>
         <%= label_tag(:per_page) do %>
           Show <%= select_tag(:per_page, options_for_select(['10', '20', '50', '100'], h(params[:per_page])), title: "Number of results to display per page") %>
           per page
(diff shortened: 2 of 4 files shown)


christinach:
I didn't test it with more than one members to check the sorting.

tpendragon:
You have a test that's failing - I think you just need to rebase.

christinach:
It failed once but then it went through. I will rebase.

tpendragon:
Can you write a feature test for this so it doesn't happen again?

christinach:
Yeah, @tpendragon I will. Thanks

app/helpers/catalog_helper.rb
```diff
@@ -12,4 +12,8 @@ def render_document_heading(*args)
     val = Array.wrap(presenter(document).heading).map { |v| h(v) }.join("<br />")
     content_tag(tag, val, { itemprop: "name", dir: val.to_s.dir }, false)
   end
+
+  def active_sort_fields_figgy
```
tpendragon: Can you add a comment about why this is necessary or at least what it's doing? I get the idea reading it, but it takes some mental parsing.

christinach: These are the options in the select_tag for the sort field: 
["score desc, updated_at_dtsi desc", "title_ssort asc", "title_ssort desc", "created_at_dtsi desc", "created_at_dtsi asc", "updated_at_dtsi desc", "updated_at_dtsi asc"]

in catalog/_members_collection.html.erb the method active_sort_fields will return:
#<Blacklight::Configuration::SortField label=\"title (A-Z)\", key=\"title_ssort asc\", field=\"title_ssort asc\", if=true, unless=false, sort=\"title_ssort asc\">","rows":"10","f.display_language_ssim.facet.limit":"6","f.state_ssim.facet.limit":"6","qf":["identifier_tesim","title_ssim","title_tesim","transliterated_title_ssim","transliterated_title_tesim","source_metadata_identifier_ssim","local_identifier_ssim","barcode_ssim"],"facet":"true","wt":"json","f.human_readable_type_ssim.facet.limit":"6"}},"error":{"metadata":["error-class","org.apache.solr.common.SolrException","root-error-class","org.apache.solr.common.SolrException"],"msg":"Can't determine a Sort Order (asc or desc) in sort spec '#<Blacklight::Configuration::SortField label=\"title (A-Z)\", key=\"title_ssort asc\", field=\"title_ssort asc\", if=true, unless=false, sort=\"title_ssort asc\">
This is why it errors.

I replaced the method with a new one 'active_sort_fields_figgy' which will return the keys we want. I build the array so that the result would be what we want in the sort field.

After the checkin meeting I looked into the catalog_controller and there is a configuration for the sort_field. We use it in the index view.

christinach: https://github.com/projectblacklight/blacklight/blob/240013f81fa04a2663373d38a66273edfe4712bc/app/helpers/blacklight/configuration_helper_behavior.rb#L12

https://github.com/projectblacklight/blacklight/blob/7d9ff392dd6fec3c9e87b4b5b0950c77fc92aa95/spec/helpers/blacklight/configuration_helper_behavior_spec.rb#L20

tpendragon: So there's a problem with title_ssort?

christinach:
Thanks @hackmastera. :) I'll remove them.

christinach:
@tpendragon The PR is updated

tpendragon:
@christinach Looks great. Just needs an approval from @hackmastera. Good job figuring out the issue!

PR #1312, opened 2018-05-26
Use X-Accel-Redirect for NGINX download

tpendragon:
Greatly improves the efficiency of downloading of full-res images by delegating down to nginx. The request comes in, authorizes in Figgy, and then passes to an internal path in nginx for handling the request.

This enables file resumption on server restart, HTTP Range requests, and download estimates when downloading in a browser.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -10,6 +10,11 @@ class DownloadsController < ApplicationController
     end
   end
 
+  def send_content
+    prepare_file_headers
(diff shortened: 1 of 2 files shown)


tpendragon:
`ActionDispatch::Static` uses `Rack::SendFile` which handles `X-Accel-Redirect`, so it was passing things off to nginx to handle UV content and it couldn't. Adjusted the configuration so it can handle it, and added in a fix here so it switches the mapping to the one necessary for derivatives/full res tiffs.

PR #1336, opened 2018-05-31
 Send updated message after extracting external geo metadata

### LINKED ISSUE #1335: External metadata derivative service should send record updated message
External metadata derivative service should send a record updated message after the metadata has been extracted and saved in the parent resource. Because the GeoMetadataExtractor doesn't use the change_set_persister, messages aren't fired after saving.

eliotjordan:
- Sends a `record_updated` message after extracting external geo metadata.
- Sends UPDATED geoblacklight event message when a member is updated instead of MEMBER_UPDATED

Closes #1335

### DIFF
--- a/app/derivative-services/external_metadata_derivative_service.rb
+++ b/app/derivative-services/external_metadata_derivative_service.rb
@@ -17,8 +17,6 @@ class ExternalMetadataDerivativeService
 
   attr_reader :change_set, :change_set_persister, :original_file
   delegate :mime_type, to: :original_file
-  delegate :metadata_adapter, to: :change_set_persister
-  delegate :persister, to: :metadata_adapter
   def initialize(change_set:, change_set_persister:, original_file:)
     @change_set = change_set
     @change_set_persister = change_set_persister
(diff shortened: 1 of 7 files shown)


app/services/event_generator/geoblacklight_event_generator.rb
```diff
@@ -36,7 +36,7 @@ def record_member_updated(record)
         record_deleted(record)
       elsif state == "complete"
         publish_message(
-          message("MEMBER_UPDATED", record)
+          message("UPDATED", record)
```
tpendragon: Why not just call `record_updated(record)` in this method? Or get geoblacklight to read MEMBER_UPDATED events.

app/derivative-services/external_metadata_derivative_service.rb
```diff
@@ -30,6 +30,12 @@ def cleanup_derivatives; end
   # Extract external geo metadata into parent vector or raster resource.
   def create_derivatives
     GeoMetadataExtractor.new(change_set: parent_change_set, file_node: change_set.resource, persister: persister).extract
+    # Send record updated message for parent resource
+    messenger.record_updated(parent)
```
tpendragon: Maybe GeoMetadataExtractor should use a ChangeSet/ChangeSetPersister for the parent? Then it would fire that event.

eliotjordan: 👍 That's definitely a better approach.

eliotjordan:
@tpendragon Updated PR.

tpendragon:
@eliotjordan The code looks good, but do we need a new test that would fail without this?

eliotjordan:
@tpendragon I added a test that confirms that the GeoblacklightEventGenerator is called after the ExternalMetadataDerivativeService has run.

PR #1339, opened 2018-05-31
Implementing the `EphemeraFolderMigration` Rake Task (for ensuring that completed Boxes have fully completed Folders)

### LINKED ISSUE #1279: Implement a migration for `EphemeraFolders` which should be `complete`
Identified while discussing #1276, as `IngestEphemeraJob` does not address cases where `EphemeraBoxes` which were marked as prepared for production after they were ingested

jrgriffiniii:
Resolves #1279


db/migrate/20180529160537_update_completed_ephemera_folders.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+class UpdateCompletedEphemeraFolders < ActiveRecord::Migration[5.1]
+  # Perform the one-way migration for updating EphemeraFolders
+  def change
+    EphemeraFolderMigrator.call
```
eliotjordan: If we bake this service and method into the DB migration, do we then have to keep it around forever? Unless we are going to use the EphemeraFolderMigrator more than once, I wonder if it makes sense to create a migrate job that we can later remove when it's no longer needed.

jrgriffiniii: That's a very fair point.  This approach was structured following the implementation for the migration to Valkyrie 1.0:
https://github.com/pulibrary/figgy/blob/master/db/migrate/20180306193020_migrate_to_valkyrie1.rb#L2

I can move the functionality directly into the migration and remove the service Class if that is preferred.

eliotjordan: Maybe this is a good point for discussion about best practices related to data migration. I have the same concern about that valkyrie1 service. Over time, these will pile up in the code base. We should collectively decide if we are good with that.

jrgriffiniii:
The `EphemeraFolderMigrator` Service Class has been removed, and its functionality has been structured strictly within the Migration itself.

hackmastera:
I'm confused; I thought we were talking about using a rake task instead of a db migration

app/services/valkyrie1_migrator.rb
```diff
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+# Service Class for migrating from Valkyrie 0.x releases to 1.0
+# Please note that this approach should not be undertaken for future migrations
```
hackmastera: I don't think we need this comment on the migration service; just on the db migrate script

spec/rails_helper.rb
```diff
@@ -16,7 +16,7 @@
 
 require "rspec/rails"
 require Rails.root.join("spec", "shared_specs.rb")
-Dir[Rails.root.join("spec", "support", "**", "*.rb")].sort.each { |file| require file }
+Dir[Rails.root.join("spec", "support", "**", "*.rb"), Rails.root.join("db", "migrate", "**", "*.rb")].sort.each { |file| require file }
```
hackmastera: Is this leftover from testing the migrate script without the service class? delete?

PR #1341, opened 2018-05-31
Stubs the Logger in order to suppress Bagit errors while executing tests

### LINKED ISSUE #1243: bagit error needs to be caught in specs
`[2018-05-14T16:03:31.117601 #45171] ERROR -- : /Users/headleya/src/figgy/spec/fixtures/bags/invalid_bag/data/data.txt is present but not manifested`

jrgriffiniii:
Resolves #1243


eliotjordan:
💯

PR #1351, opened 2018-06-01
Support compressed TIFFs

tpendragon:


app/derivative-services/jp2_derivative_service.rb
```diff
@@ -114,7 +114,30 @@ def use
...
+      rescue MiniMagick::Error
+        filename
+      end
+  end
+
+  # Remove compression from given TIFF file, just in case.
+  def cleaned_file
+    @cleaned_file ||=
+      begin
+        t = Tempfile.new
```
escowles: Do we need to explicitly clean up this tempfile after the derivative processing happens?

tpendragon: No, Tempfile means it should get deleted after the process closes.

app/derivative-services/jp2_derivative_service.rb
```diff
@@ -69,7 +69,7 @@ def run_derivatives
 
   def run_tiff_derivatives
     Hydra::Derivatives::Jpeg2kImageDerivatives.create(
-      filename,
+      clean_filename,
```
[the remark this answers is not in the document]
tpendragon: Yeah, I can change that on Monday.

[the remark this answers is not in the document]
tpendragon: @escowles Fixed.

app/derivative-services/jp2_derivative_service.rb
```diff
@@ -114,7 +120,28 @@ def use
...
+  end
+
+  # Remove compression from given TIFF file, just in case.
+  def cleaned_file
+    @cleaned_file ||=
+      begin
+        t = Tempfile.new
+        MiniMagick::Tool::Convert.new do |convert|
+          convert << file_object.disk_path.to_s
+          convert.compress.+
```
hackmastera: what does this do? Is it some kind of DSL for the command line option?

app/derivative-services/jp2_derivative_service.rb
```diff
@@ -114,7 +120,28 @@ def use
   end
 
   def filename
-    return Pathname.new(file_object.io.path) if file_object.io.respond_to?(:path) && File.exist?(file_object.io.path)
+    Pathname.new(file_object.disk_path)
```
hackmastera: I wonder why these error checks were in here. your change is definitely an improvement, but it makes me wonder if there were errors? Not sure how there could be an io without a path.

app/derivative-services/jp2_derivative_service.rb
tpendragon: Yeah. Makes the command line `compress+`

app/derivative-services/jp2_derivative_service.rb
tpendragon: This code is from the Valkyrie MVP and existed before `disk_path` was a required property in the Valkyrie API.

tpendragon:
> How did you get around the tika stub problem?

I didn't - it's a tiff, so the stubbed characterization is fine.

PR #1355, opened 2018-06-01
Insert download url into FGDC file after vector derivatives are created

### LINKED ISSUE #1283: Insert download url into FGDC metadata after completion of vector resource
Should the link be to a direct download or to the download page in pulmap?

eliotjordan:
Closes #1283

### DIFF
--- a/app/services/fgdc_update_service.rb
+++ b/app/services/fgdc_update_service.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+class FgdcUpdateService
+  def initialize(file_set:)
+    @file_set = file_set
+  end
+
+  def insert_onlink(url:)
+    onlink = find_or_create_node("//idinfo/citation/citeinfo/onlink")
+    onlink.content = url
+    save_changes
+  end
+
+  private
+
+    def doc
+      @doc ||= Nokogiri::XML(file_object.read)
+    end
+
+    def file_object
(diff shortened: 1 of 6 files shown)


app/services/fgdc_update_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    onlink = find_or_create_node("//idinfo/citation/citeinfo/onlink")
+    onlink.content = url
+    save_changes
+  end
+
+  private
+
+    def doc
+      @doc ||= Nokogiri::XML(file_object.read)
+    end
```
jrgriffiniii: Should errors in XML parsing be caught here?

eliotjordan: If an FGDC document makes to this point, it should not have parsing errors, but it's probably good practice to rescue them.

jrgriffiniii: @eliotjordan indeed addressed this thoroughly, my concern was not warranted (due to the nature of the ingestion process for XML files in this context)

eliotjordan: After talking through this with James and trying to mock this up in specs, it really seems that this isn't necessary. Malformed FGDC XML won't make it to this point, because Figgy won't properly characterize / ingest the file. Does that sound right @jrgriffiniii?

PR #1370, opened 2018-06-05
Import transfer notes, pbcore xml file during av bag ingest, fixes #1151

### LINKED ISSUE #1151: Add technical metadata import to ImportArchivalMediaBagJob
technical metadata is found in the `.xml` file in `data/`

hackmastera:


app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -66,22 +66,41 @@ def ingest
...
-          media_resource_change_set.member_of_collection_ids += Array.wrap(collection.id)
-          changeset_persister.save(change_set: media_resource_change_set)
         end
-      end
 
-      private
+        def add_pbcore(media_resource_change_set, sides)
+          sides.map { |side| side.split("_").first }.uniq.each do |barcode|
+            file_set = create_pbcore_file_set(barcode)
+            media_resource_change_set.member_ids += Array.wrap(file_set.id)
```
tpendragon: Why not just do `+= [file_set.id]`?

app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -66,22 +66,41 @@ def ingest
...
             media_resource_change_set.member_ids += Array.wrap(file_set.id)
             media_resource_change_set.sync
           end
-          media_resource_change_set.member_of_collection_ids += Array.wrap(collection.id)
-          changeset_persister.save(change_set: media_resource_change_set)
         end
-      end
 
-      private
+        def add_pbcore(media_resource_change_set, sides)
```
tpendragon: I considered recommending adding the pbcore file_set to a different property, but this is how we do it with Geo so seems fine.

app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -66,22 +66,41 @@ def ingest
...
-        # Retrieve a Hash of EAD Component IDs/Barcodes for file barcodes specified in a given Bag
-        # @return [Hash] map of EAD component IDs to file barcodes
-        def component_groups
-          @component_groups ||= bag.component_groups
+        # Creates and persists a FileSet for a pbcore xml file
+        # @param barcode [String] barcode_side for a given media object
+        # @return [FileSet] the persisted FileSet containing the binary and file metadata
+        def create_pbcore_file_set(barcode)
+          file_set = FileSet.new(title: barcode)
+          pbcore = bag.pbcore_parsers.find { |pbcore_parser| pbcore_parser.barcode == barcode }
```
tpendragon: I wonder if it'd be cleaner to have `bag.pbcore_parser_for_barcode`? Not super necessary, just a thought.

app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -66,22 +66,41 @@ def ingest
...
+          file = IngestableFile.new(file_path: pbcore.path, mime_type: "application/xml; schema=pbcore", original_filename: pbcore.path.basename)
```
tpendragon: To avoid demeter violations can we have `pbcore.original_filename` or something similar?

app/jobs/ingest_archival_media_bag_job.rb
tpendragon: It strikes me this is all very similar to what would happen if you did something like

```
file = IngestableFile.new(file_path: pbcore.path, mime_type: "application/xml; schema=pbcore", original_filename: pbcore.path.basename, container_attributes: {transfer_notes: my_transfer_notes})
media_resource.files = [file]
```
and then save that, relying on FileAppender.

app/jobs/ingest_archival_media_bag_job.rb
hackmastera: we do get some useful behavior for free this way (listings, downloads)

app/jobs/ingest_archival_media_bag_job.rb
hackmastera: Hm, you're right this is how it's done for FileAppender. I modeled the a/v filesets on FileAppender initially (couldn't use it because the multiple files needed to be on one fileset with special pcdm:use and everything). And then based this on that. I'm not sure whether it's better to do it this way or use the FileAppender. It might be a trade off between DRY and readability.

hackmastera: I'll try it with the FileAppender and add some comments.

PR #1383, opened 2018-06-06
Add an HTML5 viewer for audio files.

tpendragon:


spec/factories/file_set.rb
```diff
@@ -27,5 +27,13 @@
     factory :geo_vector_file_set do
       file_metadata FileMetadata.new(mime_type: ControlledVocabulary.for(:geo_vector_format).all.first.value, use: Valkyrie::Vocab::PCDMUse.OriginalFile)
     end
+
+    factory :audio_file_set do
+      file_metadata [
+        FileMetadata.new(mime_type: "audio/wav", use: Valkyrie::Vocab::PCDMUse.OriginalFile, id: "original"),
```
hackmastera: the av ingests are coming in without `OriginalFile`; they have `PreservationMaster` instead.

app/views/catalog/_media_html5_player_default.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if resource.is_a?(MediaResource) && decorated_resource.audio_file_sets.present?%>
+  <% audio_file = decorated_resource.audio_file_sets.first %>
```
hackmastera: So this player will be able to play just one of the file sets on the resource?

tpendragon: Yeah. I can add a few players, one on top of the other, if it's preferable?

hackmastera: I wonder if we should do them as part of the member list if we're going to show all of them. Or we might just want to put them on the fileset page?

tpendragon:
@hackmastera Liked the idea of adding it to the members list. Looks like this:

screen shot 2018-06-07 at 12 49 48 pm

christinach:
I like this minimalistic viewer for audio files.

PR #1386, opened 2018-06-08
Use faster Valkyrie ORM Converter

tpendragon:
Should provide significantly faster read performance for large result
sets. Needs a decent amount of testing in staging.

This makes the Identifiers to Reconcile report page render in 1.1 minutes vs 2.9 minutes.


tpendragon:
This seems to be working in staging and I can't get it to break - ready for review.

PR #1368, opened 2018-06-04
Integrates MediaInfo for media resource characterization

jrgriffiniii:
Resolves #1349 by:
- Extending `FileSet` and `FileMetadata` for the media resource technical metadata
- Implementing the `MediainfoCharacterizationService` for extracting the technical metadata when a media resource is persisted

PR #1394, opened 2018-06-12
Uses TikaFileCharacterizationService as the default for characterization

### LINKED ISSUE #1379: Refactor characterization service to default to tika
Tika should be the service of last resort. ImageMagick should explicitly check for a tiff or jpeg mime_type in `valid?`.

jrgriffiniii:
Resolves #1379 but is blocked by the merging of #1368


spec/characterization_services/tika_file_characterization_service_spec.rb
```diff
@@ -5,7 +5,6 @@
 include ActionDispatch::TestProcess
 
 RSpec.describe TikaFileCharacterizationService do
-  # it_behaves_like 'a Valkyrie::Derivatives::FileCharacterizationService'
```
hackmastera: Why don't we want this anymore?

PR #1396, opened 2018-06-12
Adds LUX (design system) to Figgy and replaces UV2 with UV3 viewer.

sdellis:


hackmastera:
@sdellis It seems like deleting package-lock.json would result in some loss of predictability between dev / staging / production versions of js packages?

What was the yarn conflict again? Was it a manifesto thing? If so I wonder if switching to graphql will enable us to move to yarn, which I believe has better tools around this type of version predictability stuff.

sdellis:
@hackmastera This is just temporarily deleted to get CI to pass. @axamei had to removed the file for her local setup to work too. I'm not certain, but I think it has to do with the fact that I was using a local file dependency of `lux-design-system` while we got a npm package ownership bug sorted out. The package-lock gets regenerated on `npm install`, so it will get checked in again during my next PR, which should be soon since I noticed a few vulnerabilities in some of the packages that I'll be updating this morning.

sdellis:
In case I wasn't clear, this can be merged unless there are still objections. I'll deal with the vulnerabilities in a separate PR.

hackmastera:
Seems like the package-lock situation is resolved.

This can't be merged until OL is deployed with https://github.com/pulibrary/orangelight/pull/1264; waiting on feedback from @tampakis for that.

In the meantime @sdellis will test on staging.

hackmastera:
tested on staging and got the okay from OL

Issue #1380, opened 2018-06-06
Refactor ingest_archival_media_bag_job to use file appender for the xml file


hackmastera:
note this is complete pending some comments in my stash; there's a mime_type problem that will require #1379 in order to resolve.

PR #1398, opened 2018-06-13
Implementing IngestArchivalMediaBagJob#add_thumbnail and IngestArchivalMediaBagJob#create_image_file_set for ingesting JPEGs in AV bags

### LINKED ISSUE #1390: Attach jpeg images from AV bags to their MediaResources

jrgriffiniii:
Resolves #1390

### DIFF
--- a/app/models/archival_media_bag_parser.rb
+++ b/app/models/archival_media_bag_parser.rb
@@ -36,7 +36,44 @@ class ArchivalMediaBagParser
     pbcore_parsers.find { |pbcore| pbcore.barcode == barcode }
   end
 
+  def image_file(barcode:)
+    image_files.find { |image| image.barcode == barcode }
+  end
+
   private
+
+    # Class modeling asset images
+    class ImageFile
+      attr_reader :path, :original_filename, :barcode
+
+      # Provide the MIME type used for all image files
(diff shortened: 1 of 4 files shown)


app/models/archival_media_bag_parser.rb
```diff
@@ -36,8 +36,45 @@ def pbcore_parser_for_barcode(barcode)
     pbcore_parsers.find { |pbcore| pbcore.barcode == barcode }
   end
 
+  def image_file(barcode:)
```
hackmastera: Can you please add a unit test to the bag parser spec since this is a public method

app/models/archival_media_bag_parser.rb
```diff
@@ -36,8 +36,45 @@ def pbcore_parser_for_barcode(barcode)
...
+    image_files.find { |image| image.barcode == barcode }
+  end
+
   private
 
+    # Class modeling asset images
+    class ImageFile
```
hackmastera: This could maybe be replaced with just an `IngestableFile`?

hackmastera: This and `IngestArchivalMediaBag#add_images` can be refactored as part of #1380

PR #1402, opened 2018-06-13
Move bag validation from bag parser initializer to public method

### LINKED ISSUE #1393: Speed up ingest of an ArchivalMediaCollection
Right now it validates the bag after you hit the button - that should just happen in the background.

hackmastera:
And use it in the ingest job, but not the barcode uniqueness validator;
fixes #1393


hackmastera:
@escowles okay it's actually for real fixed now. sorry about that!

PR #1404, opened 2018-06-14
Uses the `FileAppender` to ingest PBCore XML and JPEG files within `IngestArchivalMediaBagJob`

### LINKED ISSUE #1380: Refactor ingest_archival_media_bag_job to use file appender for the xml file

jrgriffiniii:
Resolves #1380


app/jobs/ingest_archival_media_bag_job.rb
```diff
@@ -69,6 +69,7 @@ def ingest
         component_groups.each do |cid, sides|
           media_resource_change_set = find_or_create_media_resource(cid)
           add_av(media_resource_change_set, sides)
+          media_resource_change_set.files = [] if media_resource_change_set.files.nil?
```
hackmastera: calling `prepopulate!` on the changeset should achieve this.

jrgriffiniii: When attempting to change this using `pry-byebug`, I find the following:
```
    68: def ingest
    69:   component_groups.each do |cid, sides|
    70:     media_resource_change_set = find_or_create_media_resource(cid)
    71:     add_av(media_resource_change_set, sides)
    72:     binding.pry
    73:     media_resource_change_set.prepopulate!
 => 74:     add_pbcore(media_resource_change_set, sides)
    75:     add_images(media_resource_change_set, sides)
    76:     media_resource_change_set.member_of_collection_ids += [collection.id]
    77:     changeset_persister.save(change_set: media_resource_change_set)
    78:   end
    79: end

[1] pry(#<IngestArchivalMediaBagJob::Ingester>)> media_resource_change_set.files
=> nil
```

Is this expected, and should I be doing anything before or after the call to `prepopulate!`?

hackmastera: Hm. I guess I was wrong. Maybe it would work if the property had `default: []`? But I'm not sure if nil has a meaning elsewhere in the code.

jrgriffiniii: Thank you identifying this, I've still found a way to improve this by ensuring that the `MediaResourceChangeSet` has `[]` explicitly set for the `files` property

hackmastera: 👍 looks good!

PR #1407, opened 2018-06-14
Add GraphQL API

tpendragon:
Currently supports one query (`resource()`) and one mutation (`updateResource()`).

There are some missing fields for what the order manager will need, but this should be a good starting point.


spec/graphql/types/scanned_resource_type_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe Types::ScannedResourceType do
+  subject(:type) { described_class.new(scanned_resource, {}) }
+  let(:scanned_resource) { FactoryBot.create_for_repository(:scanned_resource, viewing_hint: "individuals", title: ["I'm a little teapot", "short and stout"]) }
+  describe "class methods" do
+    subject { described_class }
+
+    # Note! These field names use a javascript-y camel-case variable style
```
jrgriffiniii: Thank you for this comment, I would not have expected this behavior

PR #1409, opened 2018-06-14
Pins to earlier manifesto version so thumbs now show in Order Manager

### LINKED ISSUE #1408: Thumbnails not displaying in Ordermanager

sdellis:
fixes #1408


sdellis:
Actually, hold off. 2.2.23 works too. It's 2.2.24 that breaks it.

PR #1425, opened 2018-06-18
Ensures that the thumbnails for member resources within Multi-Volume Works are used as thumbnail URIs for Manifests within ManifestBuilder::ThumbnailBuilder

jrgriffiniii:
Resolves https://github.com/pulibrary/pomegranate/issues/366

### DIFF
--- a/app/services/manifest_builder/thumbnail_builder.rb
+++ b/app/services/manifest_builder/thumbnail_builder.rb
@@ -27,19 +27,34 @@ class ManifestBuilder
         @helper ||= ManifestHelper.new
       end
 
-      # Generate the value Hash modeling the thumbnail resource for the Manifest
+      # Retrieve the member resources (for multi-volume works)
+      # @return [Array<Valkyrie::Resource>]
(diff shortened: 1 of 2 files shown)


app/services/manifest_builder/thumbnail_builder.rb
```diff
@@ -27,21 +27,36 @@ def helper
...
+            "@id" => helper.manifest_image_path(member),
             "profile" => "http;//iiiif.io/api/image/2/level2.json"
           }
         }
       end
 
+      # Generate the value Hash modeling the thumbnail resource for the Manifest
+      # @return [Hash, nil]
+      def thumbnail
+        member = thumbnail_id && file_set && file_set.derivative_file ? file_set : nearest_member_thumbnail_uri
```
eliotjordan: After some study I get what's going on here, but this is quite a gnarly ternary operator. I wonder if extracting the `thumbnail_id && file_set && file_set.derivative_file` term into a seperate method might make the intent more clear.

PR #1430, opened 2018-06-19
Extends ManifestBuilder::CollectionNode in order to hide MVW members

jrgriffiniii:
Resolves https://github.com/pulibrary/pomegranate/issues/362 by extending ManifestBuilder::CollectionNode in order to hide member resources when both the parent and members belong to the same Collection

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -13,7 +13,6 @@ namespace :migrate do
 
   desc "Migrates Ephemera Folders in Ephemera Boxes published in production to a completed workflow state"
   task ephemera_folders: :environment do
-    # Ensures that all member EphemeraFolders have their state properly updated
     resources(model: EphemeraFolder).each do |resource|
       cs = DynamicChangeSet.new(resource)
       cs.prepopulate!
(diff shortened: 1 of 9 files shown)


tpendragon:
I'm pretty concerned about the performance of this - this is several thousand extra queries for large collections. I wonder if we can either A) Find a way to run one query to get all the non-parented members or B) Just remove the "collection" field from nested members and fix the data.

jrgriffiniii:
Given the time already consumed by my attempts, the first approach proposed:
> Find a way to run one query to get all the non-parented members
...is something I cannot resolve as a part of this PR.

I can look to separate this into a separate issue, and take the second proposed approach.

lib/tasks/migrate.rake
```diff
@@ -11,17 +11,41 @@ namespace :migrate do
...
       cs = DynamicChangeSet.new(resource)
       cs.prepopulate!
       logger.info "Migrating folders within the box #{resource.id}..."
       change_set_persister.save(change_set: cs)
     end
   end
 
+  desc "Migrates Collection members with children who have values in the member_of_collection_ids attribute"
+  task collection_members_with_children: :environment do
+    # Ensures that all member EphemeraFolders have their state properly updated
```
hackmastera: I think you want to delete this comment

lib/tasks/migrate.rake
```diff
@@ -11,17 +11,41 @@ namespace :migrate do
...
+      change_set_persister.buffer_into_index do |buffered_change_set_persister|
+        collection.decorate.decorated_members.each do |collection_member|
+          collection_member.members.each do |child|
+            next if child.member_of_collection_ids.empty?
+
+            logger.info "Migrating the collections for member resource #{child.id}..."
+
+            child_change_set = DynamicChangeSet.new(child)
+            child_change_set.prepopulate!
+            child_change_set.member_of_collection_ids = []
```
hackmastera: Did you test this out? I would have expected it to be more like `child_change_set.validate(member_of_collection_ids: [])`

hackmastera:
It looks good to me but I'm hesitant to merge it without a test. Should we give it a trial run on staging?

jrgriffiniii:
The Rake task was tested successfully on staging.  However, an issue was encountered when MVW children which were not persisted could still have collections.  This was fixed.

hackmastera:
great, thanks for testing it!

PR #1437, opened 2018-06-19
Fix visibility to read_group translation logic

### LINKED ISSUE #1434: Fix bug in https://github.com/pulibrary/figgy/pull/1433
permission text and visibiliy text are different sometimes (e.g. "public" vs "open)

hackmastera:
resolves #1434


hackmastera:
Note this PR includes 2 commits: spec changes are a bit easier to understand if you view them separately. But still sort of messy there because of an indentation change.

PR #1471, opened 2018-06-26
Encode periods when sending to the save and ingest handler

### LINKED ISSUE #1451: Save and ingest does not find directory  “RBD1.1_c1”
trying to ingest some FA items (Derrida, RBD1.1). I have a dir full of items named such as “RBD1.1_c1”. Save and ingest won’t find the directory, so I have to go the long way around. I have 71 of these to do with another 400 pending. (report from @roelmoon)

hackmastera:
fixes #1451


hackmastera:
tested on staging
screen shot 2018-06-26 at 10 23 15 am

PR #1472, opened 2018-06-26
Ensures that `<form>` markup is generated properly when creating new resources with duplicated metadata

### LINKED ISSUE #1457: Ephemera: Save and Duplicate Metadata not working

jrgriffiniii:
Resolves #1457 by ensuring that `<form>` markup is generated properly when creating new resources with duplicated metadata (ensures that `Resource#persisted?` checks if the "id" attribute is `nil`)

### DIFF
--- a/app/controllers/ephemera_folders_controller.rb
+++ b/app/controllers/ephemera_folders_controller.rb
@@ -44,8 +44,9 @@ class EphemeraFoldersController < BaseResourceController
       template.nested_properties.first
     elsif params[:create_another]
       resource = find_resource(params[:create_another])
-      # Setting new_record to false ensures that this is not treated as a persisted Resource
-      # https://github.com/samvera-labs/valkyrie/blob/master/lib/valkyrie/resource.rb#L83
+      # Setting new_record to true ensures that this is not treated as a persisted Resource
+      # @see Valkyrie::Resource#persisted?
+      # @see https://github.com/samvera-labs/valkyrie/blob/master/lib/valkyrie/resource.rb#L83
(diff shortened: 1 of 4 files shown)


app/models/resource.rb
```diff
@@ -34,4 +34,10 @@ def geo_resource?
   def media_resource?
     false
   end
+
+  # Ensures that the state of the model is properly returned
+  # @return [TrueClass, FalseClass]
+  def persisted?
```
tpendragon: You should be able to avoid overriding this if, when we duplicate the metadata, you just make sure you set `new_record: false`

tpendragon: You should be able to do it here: https://github.com/pulibrary/figgy/blob/master/app/controllers/ephemera_folders_controller.rb#L47

app/controllers/ephemera_folders_controller.rb
```diff
@@ -44,7 +44,9 @@ def new_resource
       template.nested_properties.first
     elsif params[:create_another]
       resource = find_resource(params[:create_another])
-      resource.new(id: nil, created_at: nil, updated_at: nil, barcode: nil, folder_number: nil)
+      # Setting new_record to false ensures that this is not treated as a persisted Resource
+      # https://github.com/samvera-labs/valkyrie/blob/master/lib/valkyrie/resource.rb#L83
+      resource.new(id: nil, new_record: true, created_at: nil, updated_at: nil, barcode: nil, folder_number: nil)
```
hackmastera: the comment says `false` but the code says `true` it's sort of confusing!

hackmastera: (thank you for the comment by the way it's nice to have)

Issue #1331, opened 2018-05-30
Collection owners should get a weekly email about items that need attention


escowles:
Requires #1330 

Each week, collection owners should get an email about any items in their collection that are in need of attention (e.g., state = "pending", "metadata_review" or "final_review").

tpendragon:
What's a "collection owner"?

escowles:
@tpendragon See #1330 — but so far it's basically the person who should get these emails.

hackmastera:
depends on https://github.com/pulibrary/princeton_ansible/issues/286

hackmastera:
address https://github.com/pulibrary/pul-the-hard-way/issues/30 as part of this ticket. orangelight settings at https://github.com/pulibrary/orangelight/blob/master/config/environments/production.rb#L82.

hackmastera:
We've agreed to use this as an opportunity to try out the whenever gem

hackmastera:
This is mostly done in #1489. Final steps needed:

- [ ] figgy box/es need to be whitelisted to use the smtp server. John K. has a ticket for this.
- [ ] John K. has created a ticket to see if helpdesk can whitelist the emails for proofpoint.
-  [ ] The task can be run manually if desired. Testing was successful on staging (except for the proofpoint issue)
- [ ] PR needed for adding this task to cron via whenever.

PR #1489, opened 2018-06-28
Add collections owner report mailer

hackmastera:
refs #1331; see the issue for the remaining steps necessary to close.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -8,14 +8,16 @@ Rails.application.configure do
   config.assets.js_compressor = :uglifier
   config.assets.compile = false
   config.log_level = :info
-  config.action_mailer.perform_caching = false
   config.i18n.fallbacks = true
   config.active_support.deprecation = :notify
   config.active_record.dump_schema_after_migration = false
   config.public_file_server.headers = {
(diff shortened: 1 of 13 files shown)


config/environments/production.rb
```diff
@@ -8,14 +8,16 @@
...
   config.log_level = :info
-  config.action_mailer.perform_caching = false
   config.i18n.fallbacks = true
   config.active_support.deprecation = :notify
   config.active_record.dump_schema_after_migration = false
   config.public_file_server.headers = {
     "Cache-Control" => "public, max-age=31557600"
   }
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.smtp_settings = { address: "lib-ponyexpr.princeton.edu" }
```
jrgriffiniii: Should this value be managed within https://github.com/pulibrary/princeton_ansible?

hackmastera: I'm not sure; what are our criteria for that decision? Seems like would be fine in either place. It's in rails config for orangelight.

jrgriffiniii: Oh, sorry then, please disregard that comment.

PR #1494, opened 2018-06-29
Migrating the ProcessedEvent Class from Plum in order to ensure that Voyager updates are only processed once

### LINKED ISSUE #1493: Nightly Voyager Update shouldn't process old diffs
The old code would create something in the database for events which it had processed, I think we just need to move over something similar:

https://github.com/pulibrary/plum/blob/master/lib/voyager_updater/event.rb#L11-L13

https://github.com/pulibrary/plum/blob/master/lib/voyager_updater/event.rb#L26

jrgriffiniii:
Resolves #1493

### DIFF
--- a/spec/services/voyager_updater/event_spec.rb
+++ b/spec/services/voyager_updater/event_spec.rb
@@ -27,10 +27,6 @@ describe VoyagerUpdater::Event do
     }
   end
 
-  after do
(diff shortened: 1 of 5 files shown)


spec/services/voyager_updater/event_spec.rb
```diff
@@ -112,7 +108,7 @@
   context "when the job has been enqueued" do
```
hackmastera: Is this still the right context?

PR #1504, opened 2018-07-05
Add Viewing Direction/Thumbnails to GraphQL

### LINKED ISSUE #1492: Add remaining Resource/Fileset fields to GraphQL response
I think the only remaining pieces are bibid (if it exists), startPage, thumbnail, viewingDirection.

tpendragon:
Closes #1492

### DIFF
--- a/app/graphql/types/resource.rb
+++ b/app/graphql/types/resource.rb
@@ -26,16 +26,27 @@ module Types::Resource
     @url ||= helper.show_url(object)
   end
 
+  # We need to centralize logic for navigating a MVW's members to find a
+  # thumbnail file set. This is a hack to use the helper's logic for doing that.
+  # @TODO: Fix this.
   def helper
-    @helper ||= ManifestBuilder::ManifestHelper.new
+    @helper ||= ManifestBuilder::ManifestHelper.new.tap do |helper|
+      helper.singleton_class.include(ThumbnailHelper)
+      helper.define_singleton_method(:image_tag) do |url, _opts|
+        url
+      end
+      helper.define_singleton_method(:image_path) do |url|
+        url
+      end
+    end
   end
 
   def thumbnail
     return if object.try(:thumbnail_id).blank? || thumbnail_resource.blank?
     {
       id: thumbnail_resource.id.to_s,
-      thumbnail_url: helper.manifest_image_thumbnail_path(thumbnail_resource.id.to_s),
-      iiif_service_url: helper.manifest_image_path(thumbnail_resource)
+      thumbnail_url: helper.figgy_thumbnail_path(thumbnail_resource),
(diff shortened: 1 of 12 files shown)


sdellis:
When I try to get viewingDirection, I get this error: `Field 'viewingDirection' doesn't exist on type 'Resource'`

sdellis:
Also, perhaps the server should provide reasonable defaults rather than `null` for certain properties (i.e., startPage should be set to the first fileset, viewingDirection as LEFTTORIGHT -- or possibly derived from the language of the content, etc.) We had agreed this would be the approach for the last OrderManager and it would certainly simplify the logic on the client-side.

tpendragon:
@sdellis I think viewingDirection only exists on ScannedResource, so you probably have to do

```
... on ScannedResource {
  viewingDirection
}

```

sdellis:
No, you're right, the client should be able to handle them.

sdellis:
@tpendragon, I see `memberIds` as a String field on scanned resource. I am confused by this as I would think it would be for changing the order, labels, and viewingHints of the members and would be an array of objects (not a String). 

What I would like to do is simply send the newly updated resource back like this: 

```
{
"input": {
    "id":"4f9e91e1-2e9c-404d-a8ca-30b8c9d01d0d",
    "viewingDirection":"TOPTOBOTTOM",
    "viewingHint":"individuals",
    "startPage":"",
    "thumbnailId":"",
    "members":[
       {"id":"c8376fd6-306c-4aed-b6a8-4eacbd2ca1ab",
        "label":"p. iii","viewingHint":null},
       {"id":"acb1c188-57c4-41cb-88e0-f44aca12e565"
        ,"label":"p. ii","viewingHint":null},
       {"id":"291b467b-36af-4d1e-80f1-dc76e8e250b9",
        "label":"p. i","viewingHint":null}
    ]
  }
}
```
Is that possible?

tpendragon:
@sdellis We had decided to do multiple mutations instead of that nested mutation method.

sdellis:
I don't recall making that decision (if I was part of it), but it's unfortunate since it kind of defeats the purpose of using GraphQL. Let's discuss more tomorrow.

tpendragon:
@sdellis I don't think it does - you can do multiple mutations in one request.

app/graphql/types/resource.rb
```diff
@@ -25,7 +26,37 @@ def url
     @url ||= helper.show_url(object)
   end
 
+  # We need to centralize logic for navigating a MVW's members to find a
+  # thumbnail file set. This is a hack to use the helper's logic for doing that.
+  # @TODO: Fix this.
```
hackmastera: Try moving to wayfinders, as per slack discussion

hackmastera:
@tpendragon perhaps we should ticket this wayfinder work and merge what we have here so far?

tpendragon:
@hackmastera Created https://github.com/pulibrary/figgy/issues/1708

PR #1508, opened 2018-07-06
Capture and display derivative generation errors

### LINKED ISSUE #1411: [Figgy/production] RuntimeError: Unable to execute command "env SHAPE_ENCODING= ogr2ogr -q -nln 59290b93-1376-4dc8-bdd8-f550b69f276b -f 'ESRI Shapefile' -t_srs EPSG:3857 -preserve_fid '/tmp/sufia20180607-37699-8nctqz_1528351990535' '/tmp/sufia20180607-37699-8nctqz_out'". Exit code: pid 24215 exit 1
Error message: Can't transform coordinates, source layer has no
coordinate system.  Use -s_srs to set one.
ERROR 1: Terminating translation prematurely after failed
translation of layer ussr_coalbnd (use -skip...
## Backtrace

    line 110 of [PROJECT_ROOT]/app/derivative-services/vector_resource_derivative_service.rb: run_derivatives
    line 63 of [PROJECT_ROOT]/app/derivative-services/vector_resource_derivative_service.rb: create_derivatives
    line 8 of [PROJECT_ROOT]/app/jobs/create_derivatives_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/38127117)

eliotjordan:
- Saves error messages encountered during derivative generation in file metadata.
- Adds configuration to allow for filtering certain error types and messages in honeybadger.

Closes #1411 

screenshot 2018-07-06 13 44 21

screenshot 2018-07-06 13 46 05


app/derivative-services/jp2_derivative_service.rb
```diff
@@ -180,4 +180,17 @@ def cleanup_derivative_metadata(derivatives:)
         buffered_persister.save(change_set: updated_change_set)
       end
     end
+
+    def storage_adapter
```
tpendragon: Any reason you moved this?

eliotjordan: A bit of OCD. Putting methods in alphabetical order.

eliotjordan: Change these back?

tpendragon: Nah, it's fine.

app/derivative-services/jp2_derivative_service.rb
```diff
@@ -180,4 +180,17 @@ def cleanup_derivative_metadata(derivatives:)
...
+      @storage_adapter ||= Valkyrie::StorageAdapter.find(:derivatives)
+    end
+
+    # Updates error message property on the original file.
+    def update_error_message(message:)
```
tpendragon: Is there somewhere we can standardize this method? It's copied/pasted in every derivative service yeah?

eliotjordan: I agree, it needs some DRYing out. I was conflicted about the best approach -  mixin module probably. However, it's not included in every derivative service (JP2, Image, Vector, Raster only). There are a handful of other methods with the same issue. `cleanup_derivative_metadata`, `file_object`, and `storage_adapter` for example.

hackmastera: In the event that we do decide to DRY here, I would hope to explore inheritance. I don't think a mixin would be the right choice here; it seems like these are specific to derivatives creation.

PR #1652, opened 2018-08-10
Update Valkyrie to RC2

tpendragon:


tpendragon:
The github status says it's at 99.989%, but it's pointing at the wrong report, this is the latest: https://coveralls.io/builds/18434722

PR #1660, opened 2018-08-15
Prefix geo derivative names and ids

### LINKED ISSUE #1659: Update ids passed to GeoServer
We need to change the scheme for IDs that are passed to GeoServer and are included in the GeoBlacklight references because of a bug in GeoServer that is unlikely to be fixed in the near future. From a slack message on the #geo channel:

> So that issue that Wangyal brought up is a problem with GeoServer (and probably with an XML library that GeoServer uses). When you request features using WFS, GeoServer generates an XML-based GML document. Each featureMember begins with a fully qualified name. Something like `<test:4q77fv84n>`, which is the workspace (test), a colon, and then the layer identifier (4q77fv84n). If the identifier begins with a letter, everything is fine. If the identifier begins with a number then the XML validator chokes and throws an exception: `Failed to parse QName`. I’m not sure where the best place to solve this is. There are numerous references to similar problems with WFS layer names in GeoServer tickets with the recommended resolution to change the layer name. I imagine the issue is upstream of GeoServer. I was thinking that we might fix this by simplifying our layer names by using the ark name prepended with a letter of series of letters. Something like: `princeton-4q77fv84m`

eliotjordan:
- Sets the output projection for vector derivatives to EPSG:4326 to avoid shapefile projection issues in GeoServer. The modest gains in performance when using 3857 are mitigated by increased complexity in setting up new layers.
- Prefixes geo derivative names and ids with a letter to avoid XML validation errors with GeoServer WFS generated GML.

Closes #1659

### DIFF
--- a/app/derivative-services/raster_resource_derivative_service.rb
+++ b/app/derivative-services/raster_resource_derivative_service.rb
@@ -81,7 +81,7 @@ class RasterResourceDerivativeService
     {
       input_format: original_file.mime_type.first,
       label: :display_raster,
-      id: resource.id,
+      id: prefixed_id,
       format: "tif",
       srid: "EPSG:3857",
       url: URI("file://#{temporary_display_output.path}")
@@ -102,6 +102,12 @@ class RasterResourceDerivativeService
   def parent
     decorator = FileSetDecorator.new(change_set)
     decorator.parent
+  end
+
+  # Resource id prefixed with letter to avoid restrictions on
+  # numbers in QNames from GeoServer generated WFS GML.
+  def prefixed_id
+    "p-#{resource.id}"
   end
 
   def run_derivatives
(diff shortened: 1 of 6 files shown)


app/derivative-services/raster_resource_derivative_service.rb
```diff
@@ -81,7 +81,7 @@ def instructions_for_display
     {
       input_format: original_file.mime_type.first,
       label: :display_raster,
-      id: resource.id,
+      id: prefixed_id,
       format: "tif",
       srid: "EPSG:3857",
```
jrgriffiniii: Using EPSG:3857 for raster resources is fine here?

eliotjordan: Yeah. It's preferred. The issue is because GeoServer can't autogenerate a shapefile layer in 3857 very well.

Issue #1583, opened 2018-07-27
Ingest "19th c. sheet music" (pudl0020)


tpendragon:
Note: 3 or 4 vols in PUDL; many more images available. Migrate manually?

Details: https://docs.google.com/spreadsheets/d/1f-c3goE0CC2QJ0Cbk_-JN67nk3fpI9J2Dl37AhGIzos/edit#gid=17

tpendragon:
http://pudl.princeton.edu/collections/pudl0020

jrgriffiniii:
I suspect that the items in this collection which have been published to PUDL do have Voyager records, but they are not linked by ARK from the MARC record (please see http://pudl.princeton.edu/objects/xd07gt10f and https://catalog.princeton.edu/catalog/10738958)

jrgriffiniii:
Without proper Bib. IDs, `BulkIngestService` fails to create new resources during ingestion (as the attributes provided for the `ScannedResourceChangeSet` are invalid)

jrgriffiniii:
Directory names for these are not valid bib. IDs (e. g. `smb://diglibdata1.princeton.edu/pudl/pudl0020/vol02/001/`), hence I have introduced the following: https://github.com/pulibrary/figgy/compare/issues-1583-jrgriffiniii-pudl0020.

I can proceed by issuing this as a pull request, and then looking to ingest these into staging using a local ID containing `pudl0020/vol02/001`

hackmastera:
@jrgriffiniii do we want to expand the script for use with pulfa objects? Or should we keep the `bibdata?` check in, as well?

jrgriffiniii:
I'm sorry, I misunderstood, where are these referenced from within PULFA?  I cannot find metadata for these.  For example using "Alpine songs" from http://pudl.princeton.edu/objects/xg94hq91z, I cannot retrieve any records: https://findingaids.princeton.edu/?v1=Alpine+songs&f1=kw&b1=AND&v2=&f2=kw&b2=AND&v3=&f3=kw&year=before&ed=&ld=&rpp=10&start=0

hackmastera:
@jrgriffiniii it might be helpful if you create the PR and i can comment there

jrgriffiniii:
@hackmastera please see https://github.com/pulibrary/figgy/pull/1668

jrgriffiniii:
[...]

PR #1668, opened 2018-08-20
Extending the validation for bib. ID directory names within BulkIngestService#attach_dir

jrgriffiniii:
Currently if a directory with the name `001` is encountered, it is treated as a valid bib. ID by `BulkIngestService`, but invalidated by the `ScannedResourceChangeSet`.  Discovered while addressing #1583


app/services/bulk_ingest_service.rb
```diff
@@ -21,6 +21,14 @@ def attach_each_dir(base_directory:, property: nil, file_filter: nil, **attribut
     end
   end
 
+  # Determines whether or not the string encodes a bib. ID
+  # See SourceMetadataIdentifierValidator#validate
+  # @param [String] value
+  # @return [Boolean]
+  def bibdata_identifier?(value)
+    RemoteRecord.retrieve(value).success?
```
hackmastera: I believe this will succeed whether it's a bibdata id or a pulfa id.

There are some pulfa resources on our list of things to ingest, so maybe this test is what we want, but just with a different method name like `valid_identifier?` or something?

the `bibdata?` implementation is rather naive
https://github.com/pulibrary/pul_metadata_services/blob/master/lib/pul_metadata_services/client.rb#L19-L21
I guess another option would be to change it there or implement something more specific here.

jrgriffiniii: I have created a separate issue for pul_metadata_services: https://github.com/pulibrary/pul_metadata_services/issues/32

jrgriffiniii:
Blocks https://github.com/pulibrary/figgy/issues/1583

PR #1682, opened 2018-08-21
Changing the DraftPublishWorkflow to DraftCompleteWorkflow

### LINKED ISSUE #1671: Change Draft/Publish workflow to Draft/Complete
Match end state name for consistency.

jrgriffiniii:
Resolves #1671


app/models/draft_complete_workflow.rb
```diff
@@ -1,37 +1,37 @@
...
 
     # ingest workflow
-    event :publish do
-      transitions from: :draft, to: :published
+    event :complete do
+      transitions from: :draft, to: :completed
     end
-    event :unpublish do
-      transitions from: :published, to: :draft
+    event :draft do
```
hackmastera: Is this what the event is called in other workflows? It seems confusing to have an event and state with the same name.

jrgriffiniii: Understood, I can adjust this; I was basing it off of :https://github.com/pulibrary/figgy/blob/master/app/models/folder_workflow.rb#L11

app/models/draft_complete_workflow.rb
```diff
@@ -1,37 +1,37 @@
...
 
     # ingest workflow
-    event :publish do
-      transitions from: :draft, to: :published
+    event :make_complete do
+      transitions from: :draft, to: :complete
     end
-    event :unpublish do
-      transitions from: :published, to: :draft
+    event :draft do
```
hackmastera: `make_draft`?

PR #1704, opened 2018-08-21
Improve GeoServer delete message behavior

### LINKED ISSUE #1669: Ensure that FileSet parent visibility is accessible when they are being deleted
## Backtrace

    line 61 of [PROJECT_ROOT]/app/services/geoserver_message_generator.rb: parent
    line 96 of [PROJECT_ROOT]/app/services/geoserver_message_generator.rb: vector_resource_parent?
    line 54 of [PROJECT_ROOT]/app/services/geoserver_message_generator.rb: layer_type

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/39147191)

eliotjordan:
- Better handles conditions where a vector file set's parent resource no longer exists or when the member files are missing.
- Sends two messages to GeoServer on every delete. One for the public workspace and one for the restricted workspace. This will help to reduce manual cleanup on GeoServer because of anomalous events in Figgy. 

Tested this in staging, and was able to delete Vector Resources without error as well as a FileSet that had become separated from it's parent.

Closes #1669

### DIFF
--- a/app/services/event_generator/geoserver_event_generator.rb
+++ b/app/services/event_generator/geoserver_event_generator.rb
@@ -14,8 +14,14 @@ class EventGenerator
     end
 
     def derivatives_deleted(record)
+      # Attempt to delete from both public and restricted
+      # workspaces to make sure all traces of the file
+      # are cleaned up on GeoServer.
       publish_message(
-        message("DELETED", record)
+        message("DELETED", record, public_workspace)
+      )
+      publish_message(
+        message("DELETED", record, authenticated_workspace)
       )
     end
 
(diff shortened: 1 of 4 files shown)


tpendragon:
That's a genius workaround.

app/services/event_generator/geoserver_event_generator.rb
```diff
@@ -14,8 +14,14 @@ def derivatives_created(record)
...
 
     def derivatives_deleted(record)
+      # Attempt to delete from both public and restricted
+      # workspaces to make sure all traces of the file
+      # are cleaned up on GeoServer.
       publish_message(
-        message("DELETED", record)
+        message("DELETED", record, public_workspace)
+      )
+      publish_message(
```
jrgriffiniii: 💯 To this approach, I definitely could see how changing `#authenticated_workspace`  or `Figgy.config["geoserver"]["authenticated"]["workspace"]` in the config. would provide adopters with the ability to have separate GeoServer instances if needed

PR #1729, opened 2018-08-23
Fix geoblacklight iiif references for map sets

### LINKED ISSUE #1726: GeoBlacklight documents don't include IIIF references for MapSets
IIIF references for map sets are not generated in the GeoBlacklight document. Existing map sets seem to be fine, but newly created map sets are not.

eliotjordan:
It appears that a map set's thumbnail id _should_ be set to the id of a child scanned map. Changes here detect if the thumbnail id is a FileSet or a child member and returns the correct FileSet.

Closes #1726

### DIFF
--- a/app/services/geo_discovery/document_builder/document_path.rb
+++ b/app/services/geo_discovery/document_builder/document_path.rb
@@ -70,10 +70,13 @@ module GeoDiscovery
         # Returns a map set's thumbnail file set decorator.
         # @return [FileSetDecorator] thumbnail file set decorator
         def map_set_file_set
-          member_id = Array.wrap(resource_decorator.thumbnail_id).first
-          return unless member_id
-          file_set = query_service.find_by(id: member_id)
-          file_set.decorate if file_set && file_set.is_a?(FileSet)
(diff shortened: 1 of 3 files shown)


app/services/geo_discovery/document_builder/document_path.rb
```diff
@@ -70,10 +70,13 @@ def host
...
-          member_id = Array.wrap(resource_decorator.thumbnail_id).first
-          return unless member_id
-          file_set = query_service.find_by(id: member_id)
-          file_set.decorate if file_set && file_set.is_a?(FileSet)
+          @map_set_file_set ||= begin
+            member_id = Array.wrap(resource_decorator.thumbnail_id).first
+            return unless member_id
+            member = query_service.find_by(id: member_id)
+            return member.decorate if member && member.is_a?(FileSet)
+            member.decorate.geo_members.try(:first)
```
jrgriffiniii: My understanding of this is that `ScannedMaps` may either have `FileSets` or `ScannedMaps` as members when considered map sets.

PR #1747, opened 2018-08-27
Improve `copyable` documentation in IngestableFile

### LINKED ISSUE #1744: Clarify comments in ingestablefile
Please rephrase in terms of true / false. What does true mean and what will happen if true and when should it be true? what does false mean and what will happen if false and when should it be false?
https://github.com/pulibrary/figgy/blob/d2cfe2eebfe36c5cc2a3356882747036e34541c3/app/models/ingestable_file.rb#L12-L13

tpendragon:
Closes #1744.

### DIFF
--- a/app/models/ingestable_file.rb
+++ b/app/models/ingestable_file.rb
@@ -9,8 +9,18 @@ class IngestableFile < Valkyrie::Resource
   # file.
   attribute :node_attributes, Valkyrie::Types::Hash
   attribute :use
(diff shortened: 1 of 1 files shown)


hackmastera:
what is the default behavior?

Do we use different storage adapter in figgy? in generally, should I be setting this one way or another?

tpendragon:
@hackmastera That better?

hackmastera:
yes thank you!!

PR #1752, opened 2018-08-27
Move feature specs to JS-free when possible.

tpendragon:
Locally this shaves off ~ 50 seconds from the test suite.

### DIFF
--- a/spec/features/ephemera_folder_spec.rb
+++ b/spec/features/ephemera_folder_spec.rb
@@ -2,7 +2,7 @@
 require "rails_helper"
 include ActionDispatch::TestProcess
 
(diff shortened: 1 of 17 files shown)


spec/features/ephemera_folder_spec.rb
```diff
@@ -180,17 +176,6 @@
           expect(page.find(:css, ".alert-info")).to have_content "Deleted EphemeraFolder"
         end
 
-        scenario "users see a warning if they try to use duplicate barcodes" do
```
tpendragon: Note for reviewers: This got removed because it's a duplicate test in this file.

Issue #1727, opened 2018-08-23
ARKs should never get rewritten if they point to a finding aid


tpendragon:
Right now if the marc record's first ARK is to a finding aid that'll be one that gets imported, and when the record's marked complete it'll change the endpoint for that ARK to be orangelight instead of pulfa, which is wrong.

tpendragon:
https://figgy.princeton.edu/catalog/72c78c38-5432-418c-92c9-c6dc9574aebf was the record this happened with.

Metadata Identifier: 1101494
ARK that gets imported: http://arks.princeton.edu/ark:/88435/47429918s
Proper ARK (in METS): ark:/88435/7p88ch283

Another potential solution is to just set the identifier if it's in the mets - the imported ark shouldn't override it then.

tpendragon:
When I imported "Brigham Young Collection" it copied metadata from the MARC record, which copies the first ARK that's in the 856 there, which is http://arks.princeton.edu/ark:/88435/47429918s (seen here: https://bibdata.princeton.edu/bibliographic/1101494).

When that happens the ARK is copied from the imported metadata to the local metadata (https://github.com/pulibrary/figgy/blob/master/app/change_set_persisters/change_set_persister/apply_remote_metadata.rb#L45)

Unfortunately in this case the first ARK in the metadata record isn't the one that points to PUDL, it's the one that points to the finding aid. The second ARK is the one we want but we have no way to know that. If I were to then complete the newly imported Scanned Resource, it would overwrite the redirect to point to Orangelight instead of the finding aid (because normally it would have the PUDL ark and fix -that- redirect). That happens here: https://github.com/pulibrary/figgy/blob/master/app/change_set_persisters/change_set_persister/mint_identifier.rb#L20 and https://github.com/pulibrary/figgy/blob/master/app/services/identifier_service.rb#L31.

The METS file has a statement at the top that has the PUDL ark (ark:/88435/7p88ch283). So I think we have two things we can do:

1. Update the identifier when ingesting from METS so it's for sure always the PUDL one (do that here-ish: https://github.com/pulibrary/figgy/blob/master/app/jobs/ingest_mets_job.rb#L63)

2. When we go to update the identifier, resolve the ARK first and see if it points to finding aids. If it does, then don't update metadata. Do that here-ish: https://github.com/pulibrary/figgy/blob/master/app/services/identifier_service.rb#L3

Is that clearer?

hackmastera:
[...]

PR #1758, opened 2018-08-28
Throw an error if you try to overwrite a pulfa ARK

tpendragon:
References #1727.


spec/support/stub_ezid.rb
```diff
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 module EzidStubbing
-  def stub_ezid(shoulder:, blade:)
+  def stub_ezid(shoulder:, blade:, location: "http://blankplace.com")
```
hackmastera: should we use example.com out of convention?

tpendragon:
@hackmastera Fixed.

PR #1761, opened 2018-08-28
Update RubyTika

### LINKED ISSUE #1753: [Figgy/staging] RubyTikaApp::CommandFailedError: execution failed with status WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font not found: Verdana
WARN  Font ...
## Backtrace

    line 29 of [PROJECT_ROOT]/app/characterization_services/tika_file_characterization_service.rb: json_output
    line 33 of [PROJECT_ROOT]/app/characterization_services/tika_file_characterization_service.rb: file_characterization_attributes
    line 22 of [PROJECT_ROOT]/app/characterization_services/tika_file_characterization_service.rb: characterize

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/38699091)

eliotjordan:
- Swaps RubyTikaApp for a forked version updated to Tika 1.8.
- Brings tika-config file into Figgy.
- Tests example of zipped shapefile with documentation that was causing issues.

Closes #1753


tpendragon:
@eliotjordan Can you expand on why we had to bring in the tika-config to Figgy? For specs? Do we need to update Ansible?

eliotjordan:
@tpendragon For specs, yes. However, I also discovered inconsistent results using the TIKA_CONFIG variable to set the configuration. It worked to for some of the configuration settings (parsers and detectors) but not this service loader config which quiets a torrent of warnings that RubyTika can't handle. `<service-loader initializableProblemHandler="ignore"/>`. I added a new optional param to RubyTikaApp to pass a config file path. If this is merged, we should update Ansible. I can work on a PR for that.

eliotjordan:
PR to remove tika config from Ansible created.

PR #1762, opened 2018-08-28
WIP - Graphql-powered OrderManager

sdellis:


sdellis:
Currently removing all the old ordermanager files...

PR #1765, opened 2018-08-29
IngestMETSJob should ingest SimpleResources

### LINKED ISSUE #1681: Implement a Rake task for ingesting files as SimpleResources
Currently `IngestFolderJob` and `IngestFoldersJob` are invoked from within Rake tasks using just `ScannedResource` as the model:

- https://github.com/pulibrary/figgy/blob/master/app/jobs/ingest_folder_job.rb#L3
- https://github.com/pulibrary/figgy/blob/master/app/jobs/ingest_folders_job.rb#L3
- https://github.com/pulibrary/figgy/blob/master/lib/tasks/bulk.rake#L22
- https://github.com/pulibrary/figgy/blob/master/lib/tasks/bulk.rake#L123

tpendragon:
Closes #1681.

### DIFF
--- a/app/jobs/ingest_mets_job.rb
+++ b/app/jobs/ingest_mets_job.rb
@@ -62,7 +62,7 @@ class IngestMETSJob < ApplicationJob
     def ingest
       # Only import the MODS metadata from the METS Document if a MARC record is
       #   not provided
-      change_set.source_metadata_identifier = mets.bib_id unless mets.bib_id.blank?
+      change_set.source_metadata_identifier = mets.bib_id if mets.bib_id.present?
       change_set.resource.identifier = mets.ark_id unless mets.ark_id.blank?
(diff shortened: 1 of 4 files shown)


app/jobs/ingest_mets_job.rb
```diff
@@ -120,7 +120,11 @@ def slug_to_id(slug)
     # Provide the resource class used for ingesting Resources using a METS Document
     # @return [Class]
     def resource_klass
-      ScannedResource
+      if import_mods?
```
hackmastera: So this allows us to have a source_metadata_identifier in a simple resource, but not import metadata from there? like a pulfa object?

tpendragon: This is assuming we don't have MODS resources with a source_metadata_identifier in the METS.

PR #1767, opened 2018-08-29
Graphql-powered OrderManager

eliotjordan:
Rebased version (hopefully complete without errors) of https://github.com/pulibrary/figgy/pull/1762. This required some extreme git surgery which necessitated recreation on a new branch. The `resolve merge conflict` commits on the original branch were particularly nasty, which is why it made sense to start over.


sdellis:
@eliotjordan 🥇 Yay! You are the best. Thank you for helping with that epic rebase!

app/javascript/app.vue
```diff
@@ -1,5 +1,7 @@
 <template>
   <div id="order-manager">
+    <heading level="h1">The quick brown fox</heading>
```
hackmastera: what does this heading do?

eliotjordan: Good catch. Looks unnecessary to me.

.gitignore
```diff
@@ -19,3 +19,4 @@
 /app/javascript/test/unit/coverage/*
 /yarn.lock
 dump.rdb
+package-lock.json
```
escowles: Are we ignoring `package-lock.json` now? In general, I thought it was good practice to check that in to control what versions were running in production.

app/javascript/helpers/apolloClient.js
```diff
@@ -0,0 +1,35 @@
...
+const fragmentMatcher = new IntrospectionFragmentMatcher({
+  introspectionQueryResultData: {
+    __schema: {
+      types: [
+        {
+          kind: "INTERFACE",
+          name: "Resource",
+          possibleTypes: [
+            { name: "FileSet" },
+            { name: "ScannedResource" },
```
escowles: TODO: add the just-merged ScannedMap here too — I created https://github.com/pulibrary/figgy/issues/1777 for this.

escowles:
This looks good to me — the only question I have is whether we want to ignore `package-lock.json` or keep checking it in.

.gitignore
sdellis: I think that it causes more problems than it's worth. It's the source of constant conflicts that can't be manually resolved.

app/javascript/helpers/apolloClient.js
sdellis: I can do this on this PR if you want. Need to bump the lux version anyway.

sdellis: Actually, never mind. I would rather fix both in a separate PR.

PR #1769, opened 2018-08-29
Implements the `bulk:link_invalid_thumbnails` Rake task for linking resources with invalid thumbnail IDs

### LINKED ISSUE #1732: thumbnail display inconsistent
The thumbnails in figgy aren’t all displaying right now.  See for example Ethiopic where only the new or newly-updated objects in the list view have a thumbnail showing.

jrgriffiniii:
Resolves #1732 by addressing the following:
- Implements the `bulk:link_invalid_thumbnails` Rake task for linking resource thumbnails (where the IDs are corrupted)
- Implements the `FindInvalidThumbnailResources` query in order to find resources with these invalid thumbnail IDs

### DIFF
--- a/app/queries/find_invalid_thumbnail_resources.rb
+++ b/app/queries/find_invalid_thumbnail_resources.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+class FindInvalidThumbnailResources
+  # Access the list of methods exposed for the query
+  # @return [Array<Symbol>] query method signatures
+  def self.queries
+    [:find_invalid_thumbnail_resources]
+  end
+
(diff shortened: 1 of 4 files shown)


eliotjordan:
Not sure if this is the correct place for this, but another class of invalid thumbnail ids are records whose thumbnail id points to a resource that no longer exists. Lately, I've seen this a lot for Vector Resources where the shapefile is replaced because errors were discovered during derivative generation. Usually this is noticed before completion, so I'm not sure how useful a rake task that a regular user can't run would be.

eliotjordan:
I'll add a ticket for the comment above. This PR is ready to go.

PR #1778, opened 2018-08-30
Replace SimpleResource with just a ChangeSet

### LINKED ISSUE #1681: Implement a Rake task for ingesting files as SimpleResources

### LINKED ISSUE #1766: Remove SimpleResources
An alternative implementation would be that we provide forms for ScannedResources using a different ChangeSet that allows them to input metadata.

That way we can use all the same views for everything but the form. Maybe the route would be something like `/concern/scanned_resources/metadata/new` ?

tpendragon:
Closes #1766
Closes #1681


app/models/workflow_registry.rb
```diff
@@ -4,32 +4,17 @@
...
   end
 
   # @return array of strings
   def self.public_read_states
-    hash.values.uniq.map(&:public_read_states).flatten.uniq
+    workflows.map(&:public_read_states).flatten.uniq
+  end
+
+  def self.workflows
+    [BoxWorkflow, FolderWorkflow, GeoWorkflow, BookWorkflow, DraftCompleteWorkflow]
```
hackmastera: this seems like a good solution but I wonder if there's a way to dynamically generate this array. i'm worried it will be easy to miss this step when creating a new workflow, and there wouldn't be a test or obvious bug to catch the omission.

hackmastera: `BaseWorkflow.descendants`?

hackmastera: actually we could use that in a test just to make sure everything is registered if we don't want to use it in runtime.

hackmastera:
Oh this doesn't seem to close #1681 because it doesn't handle the ingest_mets rake task; perhaps that should be a separate PR after all.

hackmastera:
disregard previous comment; i remembered it determines change set dynamically in the job.

hackmastera:
I ingested a couple of mods items with the `ingest_mets_job` and they look great. My only question is whether we want some indication on the show page that these are local-metadata Scanned Resources as opposed to imported metadata?

spec/features/root_spec.rb
```diff
@@ -18,6 +18,5 @@
     expect(page).to have_content "Test Project"
     expect(page).to have_link "View Boxes"
     expect(page).to have_link "Add Box"
-    expect(page).to have_link "New Simple Resource"
```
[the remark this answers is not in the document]
tpendragon: Created #1783

PR #1788, opened 2018-08-31
Multi-volume map sets

### LINKED ISSUE #1786: Allow multi-volume map sets
Need to update IIIF Manifest generator to allow multivolume map sets. This might also fix the sammelband display issues for map sets.

eliotjordan:
Closes #1786


eliotjordan:
Sammelbands are not tested now. Working on it...

Issue #1612, opened 2018-07-27
Ingest "Arabic movie posters" (pudl0100)


tpendragon:
Notes: MODS

Details: https://docs.google.com/spreadsheets/d/1f-c3goE0CC2QJ0Cbk_-JN67nk3fpI9J2Dl37AhGIzos/edit#gid=33
Depends on #1674
Depends on #1716

tpendragon:
Files: https://drive.google.com/drive/u/0/folders/0B4Wo5hgOEFY3ZlZGUU5OS3FIbDA

tpendragon:
http://pudl.princeton.edu/collections/pudl0100

jrgriffiniii:
There are three different filename prefixes for the METS files within the directory `posters`:
- `le1_`
- `eg1_`
- `iq1_`

These seem to be linked to the countries of origin (Lebanon, Egypt, and Iraq), but for Lebanon no `placeTerm` is provided.

Within the directory `lc`, (for "Lobby cards" as opposed to "posters") the files continue for Lebanon, Egypt, and Syria with some additional prefixes:
- `eg0_`
- `eg1_`
- `eg2_`
- `egx_`
- `le0_`
- `le1_`
- `lex_`
- `sy1_`
- `syx_`

jrgriffiniii:
The MODS metadata within the METS files follows this structure:

Field | Element | XPath | Example | Authorities/Encodings
-- | -- | -- | -- | --
Title | titleInfo | mods:mods/mods:titleInfo | The Golden Age of Comedy |  
Resource Type | typeOfResource | mods:mods/mods:typeOfResource | still image |  
Genre | genre | mods:mods/mods:genre | Posters | AAT
Date Created | dateCreated | mods:mods/mods:originInfo/mods:dateCreated | 1901-2000 |  
Language | languageTerm | mods:mods/mods:language/mods:languageTerm | ara | ISO639-2b
Extent | extent | mods:mods/mods:physicalDescription/mods:extent | 1 piece. |  
Collection | collection | mods:mods/mods:relatedItem[@type="host"]/mods:titleInfo/mods:title | Middle Eastern Film Posters Digitization Initiative |  
Identifier | identifier | mods:mods/mods:identifier | le1-0143 |  
Use Rights | accessCondition | mods:mods/mods:accessCondition[@type="useAndReproduction"] | http://www.princeton.edu/~rbsc/research/rights.html |  
[...]

PR #1796, opened 2018-09-06
Changes for PUDL0100

tpendragon:
Allows for transliterated actors and does the rendering for it. Due to a
bug in solr's conversion of RDF::Literals and nested objects together (tracked at https://github.com/samvera-labs/valkyrie/issues/596),
this PR also converts the catalog show page to use a record out of
Postgres instead of Solr. I doubt there'll be any noticeable performance implications, but we should keep an eye on it.

References #1612

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -19,7 +19,7 @@ class SolrDocument
   use_extension(LinkedData)
 
   def resource
-    @resource ||= Valkyrie::MetadataAdapter.find(:index_solr).resource_factory.to_resource(object: to_h)
+    @resource ||= Valkyrie::MetadataAdapter.find(:indexing_persister).query_service.find_by(id: id)
   end
 
   def decorated_resource
--- a/spec/models/mets_document/mods_document_spec.rb
+++ b/spec/models/mets_document/mods_document_spec.rb
@@ -27,11 +27,53 @@ RSpec.describe METSDocument::MODSDocument do
         expect(mods_document.series).to eq ["The Subway Sun. Volume 3. Number 6"]
         expect(mods_document.access_condition).not_to be_blank
         expect(mods_document.holding_simple_sublocation).to eq ["Mudd"]
-        expect(mods_document.shelf_locator).to eq ["MC085. Box 135"]
+        expect(mods_document.shelf_locator).to eq ["Mudd, MC085. Box 135"]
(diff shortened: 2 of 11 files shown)


tpendragon:
screen shot 2018-09-05 at 4 54 39 pm
This is what Actors looks like right now.

app/models/solr_document.rb
```diff
@@ -19,7 +19,7 @@ class SolrDocument
   use_extension(LinkedData)
 
   def resource
-    @resource ||= Valkyrie::MetadataAdapter.find(:index_solr).resource_factory.to_resource(object: to_h)
+    @resource ||= Valkyrie::MetadataAdapter.find(:indexing_persister).query_service.find_by(id: id)
```
hackmastera: is it just this line that converts the catalog show page to use a record out of
Postgres instead of Solr? I'd rather see that change in its own commit (fine with me to be in this PR).

spec/models/mets_document/mods_document_spec.rb
```diff
@@ -200,7 +242,7 @@
     let(:mets_file) { Rails.root.join("spec", "fixtures", "mets", "pudl0038-7350.mets") }
 
     it "accesses the shelfLocator field" do
-      expect(mods_document.shelf_locator).to contain_exactly "Box AD01, Item 7350"
+      expect(mods_document.shelf_locator).to contain_exactly "Mudd, Box AD01, Item 7350"
```
hackmastera: Could we run this by Alexis? It seems redundant to me with the holding location.

tpendragon: It might be redundant, but sometimes that holding location isn't "Mudd", and it's like, "Curator's Office"

tpendragon:
@hackmastera I split the commits up - better?

spec/models/mets_document/mods_document_spec.rb
tpendragon:
```
Trey Pendragon [11:41 AM]
@Alexis A. I just added a change for ingest of Russian movie posters which displays the location as "Curator's Office, Cabinet 11/06". Unfortunately the MODS in PUDL also has a subLocation for Mudd so things like grounds and buildings would show the location as "Mudd, MC085. Box 135." Is that alright?


Alexis A. [11:42 AM]
@Trey Pendragon i don't see any problem with that
```

Issue #1588, opened 2018-07-27
Ingest "Grounds and Buildings" (pudl0038)


tpendragon:
Note: MODS
Depends on #1675
Depends on #1681

tpendragon:
Files: https://drive.google.com/drive/u/0/folders/0B4Wo5hgOEFY3ZlZGUU5OS3FIbDA

tpendragon:
http://pudl.princeton.edu/collections/pudl0038

tpendragon:
These are MODS, but based on a finding aid. Marking as a dependent of #1675.

tpendragon:
The finding aid these are pulled from is here: https://findingaids.princeton.edu/collections/AC111.xml

The METS files are laid out such that the first folder is the "box" identifier, and the second folder is the "folder" number. So `pudl0038/ad001/001/7350.mets` is from AD01, folder 1. There is only metadata at the folder level in the finding aid, there are no component IDs for each photo.

I'll work on a metadata table for what's in the MODS below.

tpendragon:
| Field | Example |
| --- | --- |
| title | Aaron Burr Hall |
| dateCreated (approximate) | 1891 |
| extent | 1 photograph |
| genre (aat) | photographic prints |
| subject (lcsh) (corporate) | Aaron Burr Hall (Princeton, N.J.) |
| note | Affixed to cardboard backing, written on back "179" |
| location | Princeton University Library. Department of Rare Books and Special Collections. Seeley G. Mudd Manuscript Library. |
| holdingLocation | Mudd |
| shelfLocator | Box AD01, Item 7350 |
| accessCondition - useAndReproduction | Single photocopies may be made for research purposes.... |
| accessCondition - restrictionOnAccess | The collection is open for research |

tpendragon:
This seems simple to the point that we could add some fields and ingest these now.

tpendragon:
Make sure when we migrate these we maintain the box number, folder number, and item number.

tpendragon:
We're going to ingest this one first.

hackmastera:
We can get box / item from the mets. I'm checking whether the folder number is necessary (if so, i'm not sure why it wouldn't have been included in the mets).

Filed in a field mapping in table above. We won't need any new fields in figgy; i'll remote that dependency note.

hackmastera:
ah, for the fields that don't seem to be working: i think they actually are but since the script only ingests to scanned resource the fields don't come through the change set.

[...]

PR #1810, opened 2018-09-07
Mods subtitle, coverage_point to allow ingest of pudl0017

hackmastera:
refs #1588


hackmastera:
yes, 38, sorry about that! updated.

PR #1813, opened 2018-09-07
Adds title with nonsort tag, publisher and published_date

christinach:


app/models/mets_document/mods_document.rb
```diff
@@ -29,6 +29,21 @@ def title
...
+      title_value
+    end
+
+    def publisher
+      values = value_from xpath: "mods:originInfo/mods:publisher"
+      values.join(" - ")
+    end
+
+    def published_date
+      date_array = normalize_whitespace(value_from(xpath: "mods:originInfo/mods:dateOther") +
```
tpendragon: I think these should be separate fields, rather than all combined into `published_date`

app/models/mets_document/mods_document.rb
```diff
@@ -20,7 +20,7 @@ def initialize(element)
 
     def title
```
tpendragon: This is good, but we need a `sort_title` field too which is the title without the nonSort bit.

app/models/mets_document/mods_document.rb
christinach: I will add them as separate fields and display them separately. correct?

tpendragon: :+1:

app/models/mets_document/mods_document.rb
christinach: I will add this as well. Thanks!

christinach:
@tpendragon can you please check the changes?

app/models/mets_document/mods_document.rb
```diff
@@ -29,6 +29,30 @@ def title
           title
         end
       end
+      non_sort = value_from xpath: "mods:titleInfo/mods:nonSort"
+      return "#{non_sort} #{title_value}".gsub(/["\[\]]/, "") unless non_sort.empty?
+      title_value
+    end
+
+    def sort_title
+      normalize_whitespace(value_from(xpath: "mods:titleInfo/mods:title")).join(",")
```
tpendragon: Can you make this only return if there's a nonSort element?

app/models/mets_document/mods_document.rb
```diff
@@ -29,6 +29,30 @@ def title
...
+    end
+
+    def publisher
+      values = value_from(xpath: "mods:originInfo/mods:publisher")
+      values.join(" - ")
+    end
+
+    def date_other
```
tpendragon: I think you should keep this `date_published`

app/models/mets_document/mods_document.rb
christinach: ok

app/models/mets_document/mods_document.rb
christinach: @tpendragon, Why do we want to display date_published with an array of: date_other, date_copyright, date_issued and also display the date_other, date_copyright, date_issued? Isn't this redundant?

tpendragon: No, I'm saying I think `date_other` IS `date_published` - we don't want an array.

christinach: oh ok, I didn't understand that date_other IS date_published. thanks

christinach:
@tpendragon

PR #1843, opened 2018-09-12
Add new title index field and use it in catalog controller

### LINKED ISSUE #1837: Title with subtitle is displaying as an ID in search results
It also doesn't appear to be searchable. see, e.g., the first object in this result: 

https://figgy.princeton.edu/?f%5Bmember_of_collection_titles_ssim%5D%5B%5D=pudl0038

hackmastera:
fixes #1837

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -5,7 +5,15 @@ class CatalogController < ApplicationController
   layout "application"
   def self.search_config
     {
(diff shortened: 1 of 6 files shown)


tpendragon:
@hackmastera When we deploy this we'll have to reindex everything, yeah?

hackmastera:
@tpendragon yes. I will split this up to 2 PRs; in the first add the new index field and in the 2nd switch catalog controller. that way we can deploy 1, run the reindex, then deploy the 2nd one.

PR #1851, opened 2018-09-13
Checking model instead of decorator for can_have_manifests?

escowles:


hackmastera:
closing in favor of #1852

PR #1863, opened 2018-09-14
Rename bulk:update_state task to update_attrs and add rights_statement option

### LINKED ISSUE #1859: Add rake task to set rights statement for all records in a collection
model it on https://github.com/pulibrary/figgy/blob/a4b86987306ec6ec502f2ea0b28d785172f6cf6d/lib/tasks/bulk.rake#L211-L219

child of #697

hackmastera:
fixes #1859


escowles:
This looks good to me.  Do you want to remove the UpdateState service class and its test here, too?

hackmastera:
Oh yes I totally do thanks for catching that

PR #1865, opened 2018-09-14
Only include complete objects with BibIDs in the identifiers-to-reconcile report

### LINKED ISSUE #850: Only include complete items in Identifiers to Reconcile report
Report should only include items that are complete, have ARKs and have BibIDs in the Source Metadata Identifier field

escowles:
Fixes #850


spec/queries/find_identifiers_to_reconcile_spec.rb
```diff
@@ -19,10 +20,15 @@
       change_set.validate(source_metadata_identifier: "123456")
       saved_resource = change_set_persister.save(change_set: change_set)
 
+      change_set = ScannedResourceChangeSet.new(resource3)
```
hackmastera: Can you update the text portion (`it (...)`) to encompass the new requirement?

hackmastera: it's not super clear from the test setup why resource3 should not be retrieved because the state is taken as the default value.

PR #1880, opened 2018-09-17
Find folders with genres pre-loaded

### LINKED ISSUE #1658: EphemeraProjects#folders should get all member genres at once
https://app.datadoghq.com/apm/resource/figgy/rack.request/1f30a0afeec86c4?start=1533590414489&end=1534195214489&env=none&paused=false&traceID=379933605328130459&spanID=3467477280324743646&graphType=span_list

Right now it's sending a query for every folder, which is unnecessarily slow.

Discovered via #1657.

tpendragon:
First attempt at solving N+1 queries in Figgy. Closes #1658.

Edit: It's still spending a bunch of time in Ruby code (maybe just the cost of converting to Valkyrie Resources?), not sure if it's worth merging.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,6 @@ end
 group :development, :test do
   gem "awesome_print"
   gem "bixby"
-  gem "bullet"
   gem "bundler-audit", ">= 0.5.0", require: false
   gem "coveralls"
   gem "dotenv-rails"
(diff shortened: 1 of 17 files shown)


tpendragon:
I think this is as fast as I can get it right now. This brings it from 10-14s to 5-7s, which is better I guess? It looks like it just takes this long to spin up AR/Valkyrie objects for 3k things.

PR #1896, opened 2018-09-20
Don't re-query for filesets in ManifestBuilder

tpendragon:
Should improve efficiency of manifest generation significantly.

Discovered via datadog: https://app.datadoghq.com/apm/resource/figgy/rack.request/68b3ff7466d9457a?start=1537461404405&end=1537465004405&paused=false&env=none&traceID=694789138216265633&spanID=6994033335916105774&graphType=span_list

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -352,7 +352,7 @@ class ManifestBuilder
     end
 
     def display_image_url
-      helper.manifest_image_medium_path(id)
+      helper.manifest_image_medium_path(resource)
     rescue
       ""
     end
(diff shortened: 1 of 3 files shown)


app/services/manifest_builder.rb
```diff
@@ -453,13 +453,21 @@ def manifest_image_path(resource)
     # Retrieve the URL path for an image served over the Riiif
     # @param [String] id identifier for the image resource
```
hackmastera: update yardocs

app/services/manifest_builder.rb
```diff
@@ -453,13 +453,21 @@ def manifest_image_path(resource)
...
     # @return [String]
-    def manifest_image_thumbnail_path(id)
-      file_set = query_service.find_by(id: Valkyrie::ID.new(id))
+    def manifest_image_thumbnail_path(id_or_resource)
+      file_set = if id_or_resource.is_a?(Valkyrie::Resource)
```
escowles: I'm not a big fan of multi-line assignments — would you mind pulling that out to a method since it's used twice?

PR #1933, opened 2018-09-28
Reduce queries in GraphQL queries.

tpendragon:
By pre-loading parents for FileSets in GraphQL queries it avoids running
an inverse query for every page to determine if something is a Geo
FileSet.

### DIFF
--- a/app/graphql/types/resource.rb
+++ b/app/graphql/types/resource.rb
@@ -19,6 +19,12 @@ module Types::Resource
   end
 
   def members
+    # This loads members using FindMembersWithInverseRelationship, which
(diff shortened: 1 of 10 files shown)


app/graphql/types/resource.rb
```diff
@@ -19,7 +19,7 @@ def resolve_type(object, _context)
   end
 
   def members
-    @members ||= Wayfinder.for(object).members
+    @members ||= Wayfinder.for(object).members_with_parents
```
escowles: Can you add a comment here explaining that this retrieves the members with the parent pre-cached?

PR #1942, opened 2018-10-01
Render viewer on edit screen for Folders

### LINKED ISSUE #1935: Show viewer in ephemera edit page
Users would like to view the item while editing metadata (either correcting/enriching minimal cataloging, or doing a scan-first/catalog-from-images workflow for items that we can scan, but need to be returned to their owners quickly)

tpendragon:
Closes #1935

I sent the screenshot and a tabbed option to our PO's for Ephemera to see which they prefer. This PR will change if they don't pick this one.

screen shot 2018-10-01 at 1 00 58 pm


tpendragon:
Confirmed by POs this is the option they prefer.

PR #1952, opened 2018-10-02
Avoid querying for every thumbnail in Structure

### LINKED ISSUE #1951: Structure Editor Slow
It's because of thumbnail's N+1 query. Should be able to solve it the same way as GraphQL

tpendragon:
Closes #1951.


eliotjordan:
This query service method is uncovered:
https://coveralls.io/jobs/40857978/source_files/2073032516#L465

Maybe unneeded?

tpendragon:
Probably - removed.

PR #1962, opened 2018-10-03
Add whitespace:pre-wrap CSS property to description list element

### LINKED ISSUE #1940: Line-breaks in note fields
This ephemera folder has a description field that includes a transcribed poem with meaningful line breaks:

https://figgy.princeton.edu/catalog/4710f460-7bf6-4a5d-ba32-774ea92cdcf4

Can we add line breaks to the output to preserve the line-breaks?

eliotjordan:
Closes #1940 

screenshot 2018-10-03 18 20 13

### DIFF
--- a/app/assets/stylesheets/components/show.scss
+++ b/app/assets/stylesheets/components/show.scss
@@ -14,6 +14,10 @@ ul.tabular {
   padding-left: 0;
 }
 
+.attribute.description {
+  white-space: pre-wrap;
+}
+
 table.dataTable thead .sorting:after, table.dataTable thead .sorting_asc:after,
 table.dataTable thead .sorting_desc:after {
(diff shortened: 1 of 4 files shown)


hackmastera:
this is cool i didn't know about it. it seems like it might be good to also change the value so there aren't any spaces between the beginning/end `<li>` tags and the value of the field; I think it's giving extra indentation on the first line here which will probably look weird for other descriptions.

eliotjordan:
@hackmastera Fixed the extra whitespace issue. We were adding it to the attribute values ourselves by the way we structured the template. Also reorganized and added documentation to the application helper, it was getting a little out of control.

hackmastera:
looks good, and thanks for adding more docs!

PR #1985, opened 2018-10-22
Adds a solr.TextField for case insensitive text searches

### LINKED ISSUE #1883: Search for local identifier should not be case sensitive
Reported by Kim Leaman. When searching for "eg1-0001" it correctly returns one element, but it does not return any results when searching for "Eg1-0001".

jrgriffiniii:
Resolves #1883


tpendragon:
This would need a `pul_solr` PR too, yeah?

Issue #1994, opened 2018-10-23
Can't download PDFs for items in the Soviet Era Books for Children and Youth collection


eliotjordan:
For example: https://figgy.princeton.edu/catalog/7271c39a-2686-4b20-9273-d794ffd9848a

Reported by Thomas Keenan.

eliotjordan:
Maybe need to just clean out the old pdf file sets for these records.

jrgriffiniii:
Datadog logs show that this is generating an error logged in Honeybadger: https://app.honeybadger.io/projects/53391/faults/39116880/249b1fd0-d6df-11e8-93d8-ba5c35982ec1

jrgriffiniii:
Blocked by https://github.com/pulibrary/figgy/issues/1687

jrgriffiniii:
The PDF for this particular ScannedResource appears to have been deleted, while the FileMetadata remains persisted within the database:

```
> res = Valkyrie.config.metadata_adapter.query_service.find_by(id: Valkyrie::ID.new('7271c39a-2686-4b20-9273-d794ffd9848a'))
> cs = ScannedResourceChangeSet.new(res).prepopulate!
> cs.resource.pdf_file
=> #<FileMetadata id=#<Valkyrie::ID:0x00000000089a90e0 @id="707a47d9-fa3a-4b66-8be6-497271271e3c"> [...]
> cs.resource.pdf_file.file_identifiers
=> [#<Valkyrie::ID:0x0000000008a46020 @id="disk:///opt/repository/derivatives/70/7a/47/707a47d9fa3a4b668be6497271271e3c/derivative_pdf.pdf">]
```

When checking for this file however, it does not exist:
```
$ ls /opt/repository/derivatives/70/7a/47/707a47d9fa3a4b668be6497271271e3c/derivative_pdf.pdf
ls: cannot access '/opt/repository/derivatives/70/7a/47/707a47d9fa3a4b668be6497271271e3c/derivative_pdf.pdf': No such file or directory
```

jrgriffiniii:
Manually regenerating this required that the FileMetadata be replaced within the production environment:
```
> pdf_file = PDFGenerator.new(resource: res, storage_adapter: Valkyrie::StorageAdapter.find(:derivatives)).render
> change_set_persister.buffer_into_index do |buffered_changeset_persister|
irb(main):011:1*   cs.validate(file_metadata: [pdf_file])
[...]

PR #1996, opened 2018-10-24
Handles errors in retrieving file nodes for the DownloadsController

### LINKED ISSUE #1687: Valkyrie::StorageAdapter::FileNotFound in downloads # show
This needs to be investigated further, as it's still happening. As a first step, try to get some more useful info into the error message.

Initially we thought:
```This came up in Honeybadger - deleting a FileSet failed because of an error in one of the hooks, but the files are gone now so downloads don't work. It's essentially just an orphan metadata record.```

## Backtrace

    line 27 of [PROJECT_ROOT]/app/storage_adapters/instrumented_storage_adapter.rb: block in find_by
    line 48 of [PROJECT_ROOT]/app/storage_adapters/instrumented_storage_adapter.rb: block in trace
    line 43 of [PROJECT_ROOT]/app/storage_adapters/instrumented_storage_adapter.rb: trace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/39116880)

jrgriffiniii:
Ensures that cases where file nodes with metadata cannot be retrieved within the `DownloadsController` return a 404 response for requests and only log an error

Works towards resolving #1994 and resolves #1687

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -160,6 +160,7 @@ RSpec/NestedGroups:
     - 'spec/controllers/scanned_resources_controller_spec.rb'
     - 'spec/controllers/catalog_controller_spec.rb'
     - 'spec/controllers/simple_resources_controller_spec.rb'
+    - 'spec/controllers/downloads_controller_spec.rb'
 RSpec/VerifiedDoubles:
   Exclude:
     - 'spec/models/search_builder_spec.rb'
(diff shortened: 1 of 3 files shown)


tpendragon:
I think this is a good idea, but we'll have to keep up better on things that are returning a 404

jrgriffiniii:
Should this be closed in favor of still returning a 500 response?  @tpendragon raises an excellent point, but these are logged as errors (and I assume that these could be accessed by filtering Rails server logs using Datadog).

escowles:
I think we should try to generate a new PDF instead of returning a 404.  Only if we cannot generate a new PDF should we return an error to the client.

jrgriffiniii:
Closing in favor of the solution introduced in #1998

PR #2035, opened 2018-10-29
Generate Audio Derivatives from WAV Files

tpendragon:


app/derivative_services/audio_derivative_service.rb
```diff
@@ -0,0 +1,92 @@
...
+    @target_file = target_file
+  end
+
+  def valid?
+    ["audio/x-wav"].include?(mime_type.first)
+  end
+
+  def create_derivatives
+    progress = ActiveEncode::Base.create(file_object.disk_path, outputs: ffmpeg_outputs)
+    sleep(1 / 1000 * 50) until progress.reload.completed? || progress.failed?
```
escowles: Is there any reason to poll at a sub-millisecond frequency?

escowles: Oh, I was assuming the `*` had precedence over `/` — could this be `50 / 1000` for readability?

tpendragon:
@hackmastera `original_file` doesn't return `preservation_master.` What do you think we should do? Document that it won't trigger on MediaResources? Add a validation check to the service so it only runs on FileSets which have a SR parent?

hackmastera:
I guess for now we should document? It seemed like the plan was to bring those resources over to scanned resources. I'm not sure what our long-term plan should be. It seems like something that should be more explicit than it is / will be. I'll try to think about it more.

tpendragon:
@hackmastera Alright, added a few comments at the top.

hackmastera:
nice warnings! 😄

spec/derivative_services/audio_derivative_service_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+RSpec.describe AudioDerivativeService do
+  it_behaves_like "a Valkyrie::Derivatives::DerivativeService"
+  let(:derivative_service) do
+    AudioDerivativeService::Factory.new(change_set_persister: change_set_persister)
+  end
+  let(:adapter) { Valkyrie::MetadataAdapter.find(:indexing_persister) }
+  let(:storage_adapter) { Valkyrie.config.storage_adapter }
+  let(:persister) { adapter.persister }
+  let(:query_service) { adapter.query_service }
+  let(:file) { fixture_file_upload("files/audio_file.wav", "audio/x-wav") }
```
hackmastera: @tpendragon how did you get this mime type to work without using `run_real_characterization: true`?

PR #2038, opened 2018-10-29
Adding models/change_sets/decorators for NumismaticIssue and Coin

### LINKED ISSUE #2001: Data model for numismatics
This will be similar to Ephemera — a couple of linked entities (Issue and Coin), plus some controlled vocabularies (names of depicted rulers, donors, etc.).  Field names and types should be drawn from the database (see #2000), and/or public interface (see #2004)

Child of #2025

escowles:
Fixes #2001

### DIFF
--- a/app/change_sets/coin_change_set.rb
+++ b/app/change_sets/coin_change_set.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+class CoinChangeSet < ChangeSet
+  delegate :human_readable_type, to: :model
+  apply_workflow(DraftCompleteWorkflow)
+
+  include VisibilityProperty
+  property :department, multiple: false, required: false
+  property :size, multiple: false, required: false
+  property :die_axis, multiple: false, required: false
+  property :weight, multiple: false, required: false
+  property :donor, multiple: false, required: false
+  property :deposit_of, multiple: false, required: false
+  property :seller, multiple: false, required: false
+  property :references, multiple: false, required: false
+
+  property :read_groups, multiple: true, required: false
+  property :depositor, multiple: false, required: false
+  property :member_ids, multiple: true, required: false, type: Types::Strict::Array.of(Valkyrie::Types::ID)
+
+  # Virtual Attributes
+  property :files, virtual: true, multiple: true, required: false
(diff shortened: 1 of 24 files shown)


app/change_sets/coin_change_set.rb
```diff
@@ -0,0 +1,39 @@
...
+class CoinChangeSet < ChangeSet
+  delegate :human_readable_type, to: :model
+  apply_workflow(DraftCompleteWorkflow)
+
+  include VisibilityProperty
+  property :department, multiple: false, required: false
+  property :size, multiple: false, required: false
+  property :die_axis, multiple: false, required: false
+  property :weight, multiple: false, required: false
+  property :donor, multiple: false, required: false
```
eliotjordan: The coins table in the database doesn't have a donor field. There are a few notes about donor in the PrivateInfo and Info fields in the Accessions table, but they're not used exclusively for that. Should we keep this?

app/change_sets/coin_change_set.rb
```diff
@@ -0,0 +1,39 @@
...
+  property :deposit_of, multiple: false, required: false
```
eliotjordan: What does this field reference? I can't find anything like it in the existing database.

app/change_sets/coin_change_set.rb
```diff
@@ -0,0 +1,39 @@
...
+  property :seller, multiple: false, required: false
```
eliotjordan: Like donor, there doesn't seem to be a single field for information about the seller. The Accessions table `Info` field is as close as I can find.

app/change_sets/coin_change_set.rb
escowles: I saw these fields displayed on Coin show pages, but I agree that I don't see them in the database.  I'll go ahead and remove them for now, and we can figure out how that should be modeled in another ticket.

eliotjordan:
Looks like the 2.3.7 images aren't supported in Circle anymore.

https://hub.docker.com/r/circleci/ruby/tags/

escowles:
I'll rebase this after https://github.com/pulibrary/figgy/pull/2041 is resolved.

PR #2039, opened 2018-10-29
Extends auth. tokens for Playlists

### LINKED ISSUE #2010: Per-Resource Auth Tokens
Every resource should have an access token that provides access to its manifest and the file downloads of its derivative filesets.  The access token should not be in the record's JSON-LD etc.

Child of #1986

jrgriffiniii:
Resolves #2010


db/schema.rb
```diff
@@ -63,6 +64,12 @@
     t.index ["updated_at"], name: "index_orm_resources_on_updated_at"
   end
 
+  create_table "processed_events", force: :cascade do |t|
```
escowles: Where is this coming from?  I don't see anything in this PR — is this a change from another issue, or a side effect I'm not seeing?

app/models/book_workflow.rb
```diff
@@ -9,9 +9,9 @@ class BookWorkflow < BaseWorkflow
     state :pending, initial: true
     state :metadata_review
     state :final_review
-    state :complete
     state :takedown
     state :flagged
+    state :complete
```
escowles: Is this change needed?  IMHO, the previous place is more logical, since takedown and flagged can only follow complete.

app/views/catalog/_universal_viewer_default.html.erb
```diff
@@ -1,6 +1,5 @@
 <% if resource.class.can_have_manifests? && resource.try(:member_ids).present? %>
   <div class="intrinsic-container intrinsic-container-16x9 uv-container">
-    <%# note: UV html file is '/uv/uv' %>
```
escowles: Very minor thing: would you mind copying the comment to the new location of the UV path — I agree with @hackmastera's comment that the duplicated `/uv` might look like a mistake, so a comment clarifying it's correct is a good idea.

escowles:
@jrgriffiniii This looks good — just a couple of little things.

db/schema.rb
jrgriffiniii: This is related to https://github.com/pulibrary/figgy/pull/2039/files#diff-81be10e318670d2eefc0f09ce40e1179.  Normally `ScannedResources` stored the `token` attribute for `AuthToken` objects into the `auth_token` attribute.  However, when `ScannedResources` are taken down, the `auth_token` attribute is removed.  The `AuthToken` object remains persisted.  Once the `ScannedResource` is marked as complete at a later point in time, this is used to retrieve the existing `AuthToken` and re-link it to this `ScannedResource`.

One error I've just found here is that there may be a case where a taken down `ScannedResource` may then be deleted; under those circumstances, I don't believe that the `AuthToken` would be deleted.

app/models/book_workflow.rb
jrgriffiniii: I can restore this; it was changed in order to ensure that I could use `#final_state?` in https://github.com/pulibrary/figgy/pull/2039/files#diff-58833a622e0d7ae1e03ff31431d8b8deR29 to determine if a `ScannedResource` is complete

escowles: Would `public_readable_state?` be an acceptable alternative?

app/models/scanned_resource.rb
```diff
@@ -12,11 +12,18 @@ class ScannedResource < Resource
   attribute :workflow_note, Valkyrie::Types::Array.of(WorkflowNote).optional
   attribute :file_metadata, Valkyrie::Types::Set.of(FileMetadata.optional)
   attribute :change_set, Valkyrie::Types::String
+  attribute :auth_token, Valkyrie::Types::String
 
   def self.can_have_manifests?
     true
   end
 
+  # Ensure that resources of this class cannot be accessed with an access token
```
hackmastera: Can you double check this comment? To me tokenized_access? being true would seem to imply that it *can* be accessed with a token.

jrgriffiniii: I'm sorry for that oversight, this has been corrected.

db/schema.rb
tpendragon: Rather than de-link, would it be easier to just make the auth token not work on incomplete resources?

escowles: So the auth token would override resource visibility, but not state-based limits to visibility?  That seems reasonable to me — and would allow us to simply leave the auth token in place when a resource was taken down, and not need to delete/re-add later.

PR #2040, opened 2018-10-29
Update UV & Play MP3s

tpendragon:
Note: Haven't had a chance to test books with this new UV.


hackmastera:
these commits don't seem split out right. the "update UV" commit just deletes one file. The other commit adds / changes a ton of stuff in `public/uv` that seems like it should be part of the "update UV" commit.

app/views/catalog/_universal_viewer_default.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if resource.class.can_have_manifests? && resource.try(:member_ids).present? %>
   <div class="intrinsic-container intrinsic-container-16x9 uv-container">
-    <iframe src="/uv#?manifest=<%= request.base_url + main_app.polymorphic_path([main_app,:manifest,resource]) %>" allowfullscreen="true"></iframe>
+    <iframe src="/uv/uv#?manifest=<%= request.base_url + main_app.polymorphic_path([main_app,:manifest,resource]) %>" allowfullscreen="true"></iframe>
```
hackmastera: Do we know why we need this additional `/uv` in the path? if so it would be nice to have a comment since it looks like a typo or something.

tpendragon: Yeah, they just changed the path for the `uv.html` to be within the dist directory.

tpendragon: I tried to change this back and it wouldn't work. :(

tpendragon:
We'll need a PR in Orangelight to be deployed at the same time to allow for the new viewer location.

tpendragon:
PR for Orangelight at https://github.com/pulibrary/orangelight/pull/1450

PR #2041, opened 2018-10-30
Updating CircleCI base image

escowles:


escowles:
This was created to address an issue that turned out to be a DockerHub outtage.  Since builds are working again now that DockerHub is back online, I'm closing this.

PR #2063, opened 2018-11-01
Allow coins to be attached and removed from numismatic issues

### LINKED ISSUE #2027: Moving coins between issues
It should be possible to move a coin from one issue to another.

Child of #2025

eliotjordan:
Closes #2027

### DIFF
--- a/app/views/catalog/_members_numismatic_issue.html.erb
+++ b/app/views/catalog/_members_numismatic_issue.html.erb
@@ -1,30 +1,88 @@
-  <h2>Members</h2>
-  <table id="members" class="table table-striped">
-    <tr>
-     <th>Thumbnail</th>
-     <th>Title</th>
-     <th>Date Uploaded</th>
-     <th>Visibility</th>
-     <th>Actions</th>
-    </tr>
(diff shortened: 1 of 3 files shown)


eliotjordan:
@escowles I was thinking that this provides a way to move coins between issues without developing new UI elements. You add the coin (using it's Figgy ID) to a new Issue and then detach it from the old Issue. Did you have something else in mind?

escowles:
@eliotjordan I saw the add/detach functionality, but they didn't work for me.  When I clicked the buttons, nothing happened.  I also didn't see any controller implementation in the PR (not sure if that functionality is already implemented or not).

eliotjordan:
@escowles I tried it locally and on staging, and it worked for me. You can test this out by creating a stand-alone Coin and Issue, then copy the coin's Figgy ID, paste it into the input box, and click "Attach". 

The logic all lives in existing Javascript, so there was nothing to add to the controllers.

escowles:
Turns out it was a browser compatibility issue — I've created https://github.com/pulibrary/figgy/issues/2073 for that.

PR #2075, opened 2018-11-01
Resolves firefox referencerror

### LINKED ISSUE #2073: Add/Detach functionality does not work in Firefox

christinach:
Resolves reference error for Firefox. 
closes #2073


christinach:
I need to test it in IE

christinach:
@tpendragon can I have a 'browserstack' license or if you could test this PR in IE? Or if sb else can test it in IE?

christinach:
@escowles I tested it with firefox, chrome, opera but not IE.

christinach:
Ok, I used an 11min trial in browserstack and tested in windows 10. It's ok.

PR #2086, opened 2018-11-05
Add Playlist model, change set, decorator

### LINKED ISSUE #2079: Playlist model
Metadata:
* Label
* Visibility (default private)
* Access Token

Child of #2066

hackmastera:
resolves #2079


jrgriffiniii:
Removing https://github.com/pulibrary/figgy/pull/2086/files#diff-dd93d8c91e07bd5785c725a1eff2fc19R6 does not ensure that the visibility is set to public by default (as would be expected given https://github.com/pulibrary/figgy/blob/master/app/change_sets/concerns/visibility_property.rb#L8)

PR #2087, opened 2018-11-05
Fix ranges for Audio

tpendragon:
Audio ranges should play and show labels for each track in the "Index"
portion.


tpendragon:
screen shot 2018-11-05 at 3 51 49 pm

PR #2094, opened 2018-11-06
2045 button to create new media reserve (administrators)

### LINKED ISSUE #2045: Button to create a MediaReserve
Child of #1986
Requires #2044

Should open up a ScannedResource form with the MediaReserve ChangeSet

christinach:
closes #2045

### DIFF
--- a/app/views/shared/_add_works.html.erb
+++ b/app/views/shared/_add_works.html.erb
@@ -1,7 +1,6 @@
 <strong class="menu-heading item-with-options">Add a Work</strong>
 <ul class="item-options quick-classify">
   <% creatable_works.each do |concern| %>
-  <% byebug %>
     <li><%= link_to(
           "New #{concern.human_readable_type}",
           new_polymorphic_path([main_app, concern]),
@@ -10,3 +9,6 @@
         ) %>
     </li>
   <% end %>
+  <li>
+  	<%= link_to("New Media Reserve", new_media_reserve_scanned_resources_path) %>
+  </li>
--- a/app/views/shared/_add_works.html.erb
+++ b/app/views/shared/_add_works.html.erb
@@ -1,11 +1,12 @@
-          <strong class="menu-heading item-with-options">Add a Work</strong>
-          <ul class="item-options quick-classify">
-            <% creatable_works.each do |concern| %>
-                <li><%= link_to(
-                      "New #{concern.human_readable_type}",
-                      new_polymorphic_path([main_app, concern]),
-                      class: "item-option contextual-quick-classify #{dom_class(concern, 'new').gsub('_', '-')}",
-                      role: 'menuitem'
-                    ) %>
-                </li>
-            <% end %>
+<strong class="menu-heading item-with-options">Add a Work</strong>
+<ul class="item-options quick-classify">
+  <% creatable_works.each do |concern| %>
+  <% byebug %>
+    <li><%= link_to(
+          "New #{concern.human_readable_type}",
+          new_polymorphic_path([main_app, concern]),
+          class: "item-option contextual-quick-classify #{dom_class(concern, 'new').gsub('_', '-')}",
+          role: 'menuitem'
+        ) %>
+    </li>
+  <% end %>
--- a/spec/features/root_spec.rb
+++ b/spec/features/root_spec.rb
@@ -11,6 +11,7 @@ RSpec.feature "Home Page" do
   scenario "displays creation links for administrators" do
     click_link "Add"
     expect(page).to have_link "New Scanned Resource"
+    expect(page).to have_link "New Media Reserve", href: new_media_reserve_scanned_resources_path
     expect(page).to have_link "New Media Resource"
     expect(page).to have_link "Add a Collection", href: "/collections/new"
     expect(page).to have_link "Add an Archival Media Collection"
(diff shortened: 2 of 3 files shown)


config/routes.rb
```diff
@@ -75,6 +75,7 @@
         get :text
       end
     end
+    resources :media_reserves, controller: "scanned_resources"
```
escowles: I don't think this is needed — I think that the create route in https://github.com/pulibrary/figgy/blob/c511b407781fdb1b6969ff7c8ea20fb4a776f8bb/config/routes.rb#L90 is the only route that needs to be customized for media reserves, otherwise, the change_set can be looked up from the resource.

app/views/catalog/_home_text.html.erb
```diff
@@ -23,6 +23,10 @@
                                                     types: [NumismaticIssue],
                                                     bulk: [] } %>
 
+<%= render partial: 'catalog/work_types', locals: { label: 'Media Reserves',
+                                                    types: [MediaReserve],
+                                                    bulk: [] } %>
+
```
escowles: I was expecting MediaReserve to be in the same group as ScannedResource, not in its own section.

app/controllers/media_reserves_controller.old.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
escowles: I don't think we need a controller or model for MediaReserve — I thought we were only going to have a ChangeSet

app/views/catalog/_home_text.html.erb
christinach: I added it like this because otherwise I couldn't find a way to add a separate label for the Media Reserves button. When I kept it as [ScannedResource] vs [MediaReserve] then the panel in the home page had a title Media Reserves and a type Scanned Resource.

app/controllers/media_reserves_controller.old.rb
christinach: I meant to remove this.

escowles:
@christinach This implementation is different from what I was expecting.  I thought we would use the MediaReserveChangeSet with the ScannedResource model/controller.  So the links to create new resources would use the [`new_media_reserve` route](https://github.com/pulibrary/figgy/blob/c511b407781fdb1b6969ff7c8ea20fb4a776f8bb/config/routes.rb#L90).

christinach:
I will update. I got confused while I was trying to add the button in the 'plus' symbol based on the logic that is followed to add a button.

app/controllers/media_reserves_controller.old.rb
christinach: The reason I added the model and mapped the media_reserves resource to look the scanned_resources controller in the routes was because I couldn't add an 'New Media Reserve' button in the admin controls (+symbol). But maybe I can map the name only (the second'New Scanned Resource' ) to a new alias 'New Media Reserve'. looking into that. thanks

app/views/catalog/_home_text.html.erb
tpendragon: If this proves bothersome, in the meantime I think we can delete the button and figure that out in another issue.

app/views/shared/_add_works.html.erb
```diff
@@ -1,11 +1,14 @@
...
+    <li><%= link_to(
+          "New #{concern.human_readable_type}",
+          new_polymorphic_path([main_app, concern]),
+          class: "item-option contextual-quick-classify #{dom_class(concern, 'new').gsub('_', '-')}",
+          role: 'menuitem'
+        ) %>
+    </li>
+  <% end %>
+  <li>
+  	<%= link_to("New Media Reserve", new_scanned_resource_url) %>
```
tpendragon: This should be `new_media_reserve_scanned_resource_path`

spec/views/catalog/_home_text.html.erb_spec.rb
```diff
@@ -17,6 +17,10 @@
     it "has a links to bulk ingest resources" do
       expect(rendered).to have_link "Bulk Ingest"
     end
+
+    it "has a Media Reserves panel" do
```
tpendragon: If you remove the box above, you should remove this test too.

spec/features/root_spec.rb
```diff
@@ -11,6 +11,7 @@
   scenario "displays creation links for administrators" do
     click_link "Add"
     expect(page).to have_link "New Scanned Resource"
+    expect(page).to have_link "New Media Reserve"
```
tpendragon: Can you add `href: '/concern/scanned_resources/new/media_reserve` to this?

spec/views/catalog/_home_text.html.erb_spec.rb
christinach: thanks @tpendragon I removed it.

spec/features/root_spec.rb
christinach: yes I will add it

tpendragon:
This is fine with me. @escowles ?

app/views/shared/_add_works.html.erb
```diff
@@ -1,11 +1,14 @@
...
+    <li><%= link_to(
+          "New #{concern.human_readable_type}",
+          new_polymorphic_path([main_app, concern]),
+          class: "item-option contextual-quick-classify #{dom_class(concern, 'new').gsub('_', '-')}",
+          role: 'menuitem'
+        ) %>
+    </li>
+  <% end %>
+  <li>
+  	<%= link_to("New Media Reserve", '/concern/scanned_resources/new/media_reserve') %>
```
tpendragon: Can you make this `new_media_reserve_scanned_resource_path`?

Issue #2080, opened 2018-11-05
Playlists have tracks


escowles:
* Tracks can be added or removed
* Tracks are ordered
* Removing a track from a playlist does not delete the FileSet
* Track titles should default to the FileSet label, but should be able to be customized to the playlist context without updating the FileSet
* Tracks should not show up in search results

Child of #2066

hackmastera:
We've agreed to implement tracks as members on Playlists, where tracks use a new resource class called ProxyResource. ProxyResource have their own id, a label, and the id of the resource they proxy.

jrgriffiniii:
The current progress of this issue is as follows:
- [ ] Tracks can be added or removed
- [x] Tracks are ordered (https://github.com/pulibrary/figgy/pull/2120)
- [x] Removing a track from a playlist does not delete the FileSet (https://github.com/pulibrary/figgy/pull/2121)
- [x] Track titles should default to the FileSet label, but should be able to be customized to the playlist context without updating the FileSet (https://github.com/pulibrary/figgy/pull/2121)
- [x] Tracks should not show up in search results (https://github.com/pulibrary/figgy/pull/2120)

PR #2120, opened 2018-11-07
Extends the PlaylistsController and ChangeSetPersister support for Playlists

jrgriffiniii:
Progresses towards resolving #2080 by addressing the following:
- Ensuring that `Playlist` `ProxyFiles` have their labels overridden by their proxied resources
- Ensuring that `Playlists` use the `DraftCompleteWorkflow`
- Ensuring that `Playlists` can be added by users
- Ensuring that clients can add and remove `ProxyFiles` as members of `Playlists` in the `PlaylistsController`
- Ensuring that `Playlists` and `ProxyFiles` are suppressed from search results


app/helpers/application_helper.rb
```diff
@@ -6,7 +6,7 @@ module ApplicationHelper
   # Figgy resource types that can be created in the UI
   # @return [Array<Resource>]
   def all_works
-    [ScannedResource, MediaResource, ScannedMap, RasterResource, VectorResource, NumismaticIssue]
+    [ScannedResource, MediaResource, ScannedMap, RasterResource, VectorResource, NumismaticIssue, Playlist]
```
escowles: Do we want to suppress discovery of playlists? I thought we wanted to make them private (so end-users couldn't find them), but allow staff to find them.

app/change_set_persisters/change_set_persister/update_proxied_resource.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class ChangeSetPersister
+  class UpdateProxiedResource
```
escowles: I initially expected this class to update the resource that `proxied_file_id` points to.  Maybe `UpdateProxyResource` or `CopyAttributesFromProxiedResource` would be clearer?

jrgriffiniii:
This will need to be reconciled with #2121

app/models/search_builder.rb
```diff
@@ -35,7 +35,7 @@ def readable_states
 
   # This is a blacklist of models that should not be presented in search results
   def models_to_solr_clause
-    [FileSet, EphemeraProject, EphemeraVocabulary, EphemeraTerm, FileMetadata, EphemeraField, Template].join(",")
+    [FileSet, EphemeraProject, EphemeraVocabulary, EphemeraTerm, FileMetadata, EphemeraField, Template, Playlist, ProxyFile].join(",")
```
escowles: Do we want to suppress discovery of Playlists? I thought we wanted to make them private by default, to prevent end users from discovering them, but still allow staff to discover them.

escowles:
👍 This looks good to me — just a couple of little things.

jrgriffiniii:
There is redundancy between https://github.com/pulibrary/figgy/blob/master/app/change_set_persisters/change_set_persister/create_proxy_files.rb#L25 and https://github.com/pulibrary/figgy/pull/2120/files#diff-d1ed91bd9083255e50f7c020760cc9e3R31 that needs to be resolved.

tpendragon:
Looks fine to me. @escowles ?

Issue #2096, opened 2018-11-06
Media Reserve Object Show Page should have "Create Playlist" button


tpendragon:
The create playlist button will make a new playlist titled after the MediaReserve (eg "Playlist: Elvis' Greatest Hits") and will add all of the tracks to the playlist and redirect you there.

hackmastera:
resolved by #2121

PR #2121, opened 2018-11-07
2096 create playlist button

christinach:


hackmastera:
@christinach in future PRs can you put something like "resolves #2096" in the description? Then it will close the ticket automatically when the PR is merged. here is the github documentation https://help.github.com/articles/closing-issues-using-keywords/

christinach:
I will do, thanks. Usually I put `closes and the number'. I missed this one.

hackmastera:
👍 sorry to overexplain.

the only reason i could think that it might not work is that you have to make sure to include the `#` before the number. Oh, also it doesn't work if you put it in the title -- for a long time i had to keep moving it from title to description because i kept doing it wrong 😁

christinach:
I don't know whats the case but I do include the `#`.

PR #2131, opened 2018-11-08
Numismatic References and Citations

### LINKED ISSUE #2115: Nested numismatic references
Coin and NumismaticIssue references are difficult to collapse into a single string. We will probably need to create a References model and nest that in the coin and issue forms. Perhaps with data tables?

Child of #2025

eliotjordan:
Closes #2115


app/controllers/numismatic_citations_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  self.change_set_class = DynamicChangeSet
+  self.resource_class = NumismaticCitation
+  self.change_set_persister = ::ChangeSetPersister.new(
+    metadata_adapter: Valkyrie::MetadataAdapter.find(:indexing_persister),
+    storage_adapter: Valkyrie.config.storage_adapter
+  )
+  before_action :load_numismatic_references, only: [:new, :edit]
+
+  def create
+    # Save append_id and then remove from resource params
```
hackmastera: can you add why to this comment?

hackmastera: you're using the append_id list to get the parents to attach to?

eliotjordan: Yes. The append_id param is used to retrieve the parent numismatic issue or coin. The citation id is then inserted into the parent `numismatic_citation_ids` array.

eliotjordan:
@escowles I fixed an issue with one of the new model comments and expanded on the controller comment a little.

eliotjordan:
@hackmastera This is ready for another review.

- Updated to not overload the `append_id` parameter and moved logic to the change set persister.
- Rebased against recent Figgy file structure re-org.

PR #2132, opened 2018-11-09
Add widget for adding Recordings to Playlists

### LINKED ISSUE #2098: Playlist Show Page should have a search which displays MediaReserves to add
The search should be javascript-y, return all MediaReserves that match the query, and next to that will be the same button that a MediaReserves show page has in #2096

Requires #2096

Implementation ideas: You could query solr, you could use the blacklight API, or you could write a custom query which uses a fulltext index in postgres.

Important Note: Search should match MediaReserve titles, source metadata identifiers, composers, etc. Just by title isn't sufficient.

Child of #2066

Wireframe in #2099

tpendragon:
Closes #2098 
addtracks

### DIFF
--- a/app/change_sets/media_reserve_change_set.rb
+++ b/app/change_sets/media_reserve_change_set.rb
@@ -17,6 +17,7 @@ class MediaReserveChangeSet < ChangeSet
 
   # Virtual Attributes
   property :files, virtual: true, multiple: true, required: false
+  property :pending_uploads, multiple: true, required: false
 
   validates_with StateValidator
   validates_with MemberValidator
--- a/app/javascript/components/document_adder.vue
+++ b/app/javascript/components/document_adder.vue
@@ -1,0 +1,77 @@
+<template>
+  <div>
+    <h2>Search for Recordings to Add Tracks</h2>
+    <form v-on:submit.prevent="search">
+      <div class="input-group">
+        <input v-model="recording_query" placeholder="Search" class="form-control">
+          <span class="input-group-btn">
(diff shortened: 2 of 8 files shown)


tpendragon:
The add tracks button is definitely not how Vue wants you to do this (it's not self contained, relies on some Rails javascript stuff existing, and it won't disable the button), but having trouble replicating the behavior atm

tpendragon:
Okay, the last commit takes care of that.

escowles:
I deployed to staging to check it out, but it doesn't work for me in Firefox or Chrome.  In Chrome, I get an error in the console:

```
application-b84b79c6ee24c7a79f0f.js:1 Uncaught (in promise) TypeError: Cannot read property 'map' of undefined
    at application-b84b79c6ee24c7a79f0f.js:1
    at Array.map (<anonymous>)
    at application-b84b79c6ee24c7a79f0f.js:1
```

tpendragon:
Oh, that might be what happens when there's no media reserves

escowles:
Hmm... I tried to create a MediaReserve, but it wound up being a ScannedResource: https://figgy-staging.princeton.edu/catalog/35268b68-acde-43aa-b3c4-781f1fa9c846

tpendragon:
Old branch might be deployed

hackmastera:
is there a way to tell whether it's a media reserve or a scanned resource? I didn't think there was.

tpendragon:
> is there a way to tell whether it's a media reserve or a scanned resource? I didn't think there was.

The search widget does it by searching for ScannedResources whose change_set value is `media_reserve`

PR #2133, opened 2018-11-09
Indexing Coins' parent issue metadata, but not membership to make them discoverable

### LINKED ISSUE #2100: NumismaticIssue and Coin search visibility
Alan mentioned wanting to have the Coin resources visible to the end user while hiding the Numismatic issue.

Child of #2025

### LINKED ISSUE #2104: Numismatic Issue and Coin metadata are not searchable
Despite testing for this in the catalog controller spec, coin and issue records in figgy-staging don't appear to be searchable.

Child of #2025

escowles:
Fixes #2100; Fixes #2104

### DIFF
--- a/app/indexers/parent_issue_indexer.rb
+++ b/app/indexers/parent_issue_indexer.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+class ParentIssueIndexer
+  delegate :query_service, to: :metadata_adapter
+  attr_reader :resource
+  def initialize(resource:)
+    @resource = resource
+  end
+
+  def to_solr
+    return {} unless resource.is_a?(Coin)
+    return {} unless parents.first
+
+    parent_attributes.map { |k, v| ["#{k}_tesim", v] }.to_h
+  end
+
+  def parent_attributes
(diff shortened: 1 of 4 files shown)


app/indexers/parent_issue_indexer.rb
```diff
@@ -0,0 +1,33 @@
...
+  attr_reader :resource
+  def initialize(resource:)
+    @resource = resource
+  end
+
+  def to_solr
+    return {} unless resource.is_a?(Coin)
+    return {} unless parents.first
+
+    parent_attributes.map { |k, v| ["#{k}_tesim", v] }.to_h
```
tpendragon: We should probably prefix these, otherwise when you load the resource from solr (which admittedly we don't do), it'll have a bunch of the metadata from a coin's issue.

eliotjordan: Sorry. I merged this without reloading the page. Your comment just came up!

eliotjordan: That's an important comment. Created an issue for it: https://github.com/pulibrary/figgy/issues/2165

PR #2148, opened 2018-11-12
Add GraphQL support for Playlists.

### LINKED ISSUE #2141: Add GraphQL for Playlists

tpendragon:
Closes #2141

### DIFF
--- a/app/graphql/types/playlist_type.rb
+++ b/app/graphql/types/playlist_type.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+class Types::PlaylistType < Types::BaseObject
+  implements Types::Resource
+
+  field :manifest_url, String, null: true
+
+  def label
+    Array.wrap(object.decorate.title).first
+  end
+
+  def source_metadata_identifier; end
+
+  def viewing_hint; end
+end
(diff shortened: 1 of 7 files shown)


app/graphql/types/playlist_type.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+class Types::PlaylistType < Types::BaseObject
+  implements Types::Resource
+
+  field :manifest_url, String, null: true
+
+  # Use decorated resource to include portion note in title
```
hackmastera: I don't think this comment is relevant for playlists

app/graphql/types/proxy_file_set_type.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+class Types::ProxyFileSetType < Types::BaseObject
+  implements Types::Resource
+  def viewing_hint; end
+
+  def label
+    object.label.try(:first)
+  end
+
+  # There is no thumbnail for an audio file, so this has to return blank. The
```
hackmastera: some audio files have thumbnails. but not the kind that will be proxied I guess.

tpendragon: Do you want me to change this somehow?

hackmastera: Yeah, I guess it would be misleading if someone unfamiliar with the different collections in figgy read it. Maybe just more specific like "the audio files we plan to proxy for playlists won't have thumbnails, so [etc]"

tpendragon: That look better?

hackmastera: yes thank you!

Issue #2142, opened 2018-11-09
Add members list to Playlist for deleting ProxyFiles


tpendragon:
Child of #2066

jrgriffiniii:
Child of #2080

PR #2154, opened 2018-11-13
Implements the PlaylistMembers and PlaylistMember Vue components for removing Recordings from Playlists

### LINKED ISSUE #2080: Playlists have tracks

jrgriffiniii:
Resolves #2080 and #2142 by addressing the following:

- Implements the `PlaylistMembers` and `PlaylistMember` Vue components for rendering Recordings in Playlists and permitting users to remove them
- Ensuring that users cannot add the same Recording more than once using the `DocumentAdder` component
- Extending `CreateProxyFileSets` in order to handle cases where ProxyFileSets are removed as members

### DIFF
--- a/app/change_set_persisters/change_set_persister/create_proxy_file_sets.rb
+++ b/app/change_set_persisters/change_set_persister/create_proxy_file_sets.rb
@@ -11,35 +11,9 @@ class ChangeSetPersister
     end
 
     def run
-      if detached_proxy_file_set_ids.present?
-        delete_detached_proxy_file_sets
-        change_set.member_ids = attached_proxy_file_set_ids
-      else
-        return if file_set_ids.nil?
-        change_set.member_ids += proxy_file_set_ids
-      end
-
+      return unless file_set_ids.present?
+      change_set.member_ids += proxy_file_set_ids
       change_set.sync
-    end
-
-    def detached_proxy_file_set_ids
-      change_set.try(:detached_member_ids)
-    end
-
-    def detached_proxy_file_sets
-      @detached_proxy_file_sets ||= query_service.find_many_by_ids(ids: detached_proxy_file_set_ids)
-    end
-
-    def attached_proxy_file_set_ids
-      resource.member_ids - detached_proxy_file_set_ids
-    end
-
-    def delete_detached_proxy_file_sets
-      detached_proxy_file_sets.each do |proxy_file_set|
-        cs = ProxyFileSetChangeSet.new(proxy_file_set)
-        cs.prepopulate!
-        change_set_persister.delete(change_set: cs)
-      end
     end
 
     def proxy_file_set_ids
--- a/app/change_set_persisters/change_set_persister/create_proxy_file_sets.rb
+++ b/app/change_set_persisters/change_set_persister/create_proxy_file_sets.rb
@@ -2,8 +2,8 @@
 class ChangeSetPersister
   class CreateProxyFileSets
     attr_reader :change_set_persister, :change_set
-    delegate :query_service, to: :change_set_persister
-    delegate :resource, to: :change_set
+    delegate :query_service, :metadata_adapter, to: :change_set_persister
+    delegate :persister, to: :metadata_adapter
 
     def initialize(change_set_persister:, change_set:)
       @change_set = change_set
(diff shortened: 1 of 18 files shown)


app/change_set_persisters/change_set_persister/create_proxy_file_sets.rb
```diff
@@ -2,18 +2,46 @@
...
+      else
+        return if file_set_ids.nil?
+        change_set.member_ids += proxy_file_set_ids
+      end
+
       change_set.sync
     end
 
+    def detached_proxy_file_set_ids
+      change_set.try(:detached_member_ids)
```
tpendragon: Why is this all wrapped up in the same hook? Also, why does it need to be built into this at all - can't we just use the same delete logic we've always used for members?

app/change_sets/playlist_change_set.rb
```diff
@@ -12,6 +12,7 @@ class PlaylistChangeSet < ChangeSet
   property :read_groups, multiple: true, required: false
 
   property :file_set_ids, virtual: true, type: Valkyrie::Types::Array.of(Valkyrie::Types::ID)
+  property :detached_member_ids, virtual: true, type: Valkyrie::Types::Array.of(Valkyrie::Types::ID)
```
tpendragon: What's the purpose of this?

app/decorators/playlist_decorator.rb
```diff
@@ -23,4 +24,12 @@ def titles
     Array.wrap(label)
   end
   alias title titles
+
+  def proxies
```
tpendragon: If we need this, should be called `decorated_proxies`

app/javascript/components/document_adder.vue
```diff
@@ -53,14 +56,22 @@ export default {
       if (recording.file_set_ids.length < 1) {
         return
       }
+	// Ensure that the new FileSet IDs are unique
+	const new_file_set_ids = recording.file_set_ids.filter( file_set_id => !this.member_file_set_ids.includes(file_set_id) )
```
tpendragon: Rather than handling this here, can we just make the controller action not add them twice? Could even be in another PR.

app/javascript/components/playlist_members.vue
```diff
@@ -0,0 +1,77 @@
...
+		return file_set_id.id
+	    })
+	},
+	submit() {},
+	buildFormData() {
+	    let form = new FormData
+	    form.append('_method', 'patch')
+
+	    for(var id of this.detached_member_ids) {
+		form.append('playlist[detached_member_ids][]', id)
```
tpendragon: Seems like we could just send a DELETE for that ProxyFileSet and everything would clean up.

app/change_set_persisters/change_set_persister/create_proxy_file_sets.rb
```diff
@@ -2,6 +2,8 @@
 class ChangeSetPersister
   class CreateProxyFileSets
     attr_reader :change_set_persister, :change_set
+    delegate :query_service, to: :change_set_persister
```
hackmastera: there is also a `def query_service` so one of the two is redundant.

app/change_set_persisters/change_set_persister/create_proxy_file_sets.rb
```diff
@@ -2,6 +2,8 @@
...
+    delegate :resource, to: :change_set
```
hackmastera: this doesn't look like it's used any longer.

app/controllers/playlists_controller.rb
```diff
@@ -8,7 +8,13 @@ class PlaylistsController < BaseResourceController
   )
 
   def resource_params
-    return super unless params[:recording_id]
+    values = super
+    if !values.nil? && values[:file_set_ids]
+      file_set_ids = values[:file_set_ids]
+      values[:file_set_ids] = file_set_ids.uniq
```
hackmastera: can you just add a comment saying why we get duplicates?

app/controllers/proxy_file_sets_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+class ProxyFileSetsController < BaseResourceController
```
hackmastera: why did we need a controller? oh to power the vue stuff?

jrgriffiniii:
figgy_issues_2142_screenshot_0

PR #2164, opened 2018-11-14
Adding Numismatic Accessions

escowles:
Fixes #2126 
Depends on #2131

PR #2176, opened 2018-11-15
Adding Numismatic Accessions

### LINKED ISSUE #2126: Numismatic Accessions
Models/etc. for numismatic accession records, linked to from Coins.

Attributes from numismatics db:
* date
* person
* firm
* number (auto-incrementing sequence)
* type (one of deposit, gift, purchase, trade, or nil — controlled vocab?)
* references (like coin references)
* cost
* account
* info (description)
* private info (private notes)

Child of #2025

escowles:
Fixes #2126
* Alternative to #2164 that doesn't depend on #2131
* Reorganizes files in new resources layout

### DIFF
--- a/app/resources/numismatic_accessions/numismatic_accession_change_set.rb
+++ b/app/resources/numismatic_accessions/numismatic_accession_change_set.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+class NumismaticAccessionChangeSet < Valkyrie::ChangeSet
+  delegate :human_readable_type, to: :model
+
+  property :date, multiple: false, required: false
+  property :person, multiple: false, required: false
+  property :firm, multiple: false, required: false
+  property :accession_number, multiple: false, required: false
+  property :type, multiple: false, required: false
+  property :cost, multiple: false, required: false
+  property :account, multiple: false, required: false
+  property :note, multiple: false, required: false
+  property :private_note, multiple: false, required: false
+
+  validates_with AutoIncrementValidator, property: :accession_number
+  validates_with MemberValidator
+
+  def primary_terms
+    [
+      :date,
+      :person,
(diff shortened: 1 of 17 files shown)


app/resources/numismatic_accessions/numismatic_accession_change_set.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  validates_with AutoIncrementValidator, property: :accession_number
+  validates_with MemberValidator
+
+  def primary_terms
+    [
+      :date,
+      :person,
+      :firm,
+      :accession_number,
```
eliotjordan: We should probably not let users edit `accession_number` since we're assigning an auto-incremented value.

app/resources/numismatic_accessions/numismatic_accession_change_set.rb
```diff
@@ -0,0 +1,33 @@
...
+  property :person, multiple: false, required: false
+  property :firm, multiple: false, required: false
+  property :accession_number, multiple: false, required: false
+  property :type, multiple: false, required: false
+  property :cost, multiple: false, required: false
+  property :account, multiple: false, required: false
+  property :note, multiple: false, required: false
+  property :private_note, multiple: false, required: false
+
+  property :member_ids, multiple: true, required: false, type: Types::Strict::Array.of(Valkyrie::Types::ID)
```
eliotjordan: It looks as if Accessions are setup here to have Coin members. Is this the relationship we want? We discussed in Slack inserting the `accession_number` into a property on the Coin resource to make the connection. Also, using member_ids this way means we end up with Coins with two parents; an Issue and an Accession.

eliotjordan:
@escowles Looks good. Just need to update those failing `accession` Coin specs.

PR #2182, opened 2018-11-15
Numismatic Issue form reorganization

### LINKED ISSUE #2129: Form field groups?
The existing numismatics UI has a very dense form, with many grouped fields.  Presenting these as a long, flat list of fields will make the form difficult to navigate.  Do we want to add some way of grouping fields to improve navigation?

### LINKED ISSUE #2174: Issue form reorg
Reorganize the Numismatic Issue form/show page to match the order of the existing application:
* department
* object type
* denomination
* metal
* shape
* color
* edge
* first date/last date
* date of object
* non-CD date
* ruler
* region/state/city
* obverse figure, symbol, part, orientation, description, relationship, legend
* reverse figure, symbol, part, orientation, description, relationship, legend
* attributes (textarea)
* notes

Child of #2025

escowles:
* Updates metadata views to support either an Array or a Hash for `primary_terms` — a Hash creates a section with the key as the label, and the values as field names.
* Updates the Numismatic Issue to use a `primary_terms` Hash to group fields in a similar order to the existing application

Fixes #2174 , Fixes #2129

### DIFF
--- a/app/resources/numismatic_issues/numismatic_issue_change_set.rb
+++ b/app/resources/numismatic_issues/numismatic_issue_change_set.rb
@@ -70,47 +70,55 @@ class NumismaticIssueChangeSet < ChangeSet
   validates :visibility, presence: true
 
   def primary_terms
-    [
-      :rights_statement,
-      :rights_note,
-      :member_of_collection_ids,
-      :artist,
-      :color,
-      :date_object,
-      :date_range,
-      :denomination,
-      :department,
-      :description,
-      :edge,
-      :era,
-      :geographic_origin,
-      :issue_number,
-      :master,
-      :metal,
-      :note,
-      :object_type,
-      :obverse_attributes,
-      :obverse_figure,
-      :obverse_figure_description,
-      :obverse_figure_relationship,
-      :obverse_legend,
-      :obverse_orientation,
-      :obverse_part,
-      :obverse_symbol,
-      :place,
-      :reverse_attributes,
-      :reverse_figure,
-      :reverse_figure_description,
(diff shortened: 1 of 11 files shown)


escowles:
Screen shot showing metadata form sections:

screen shot 2018-11-15 at 4 40 14 pm

app/resources/numismatic_issues/numismatic_issue_change_set.rb
```diff
@@ -69,47 +69,55 @@ class NumismaticIssueChangeSet < ChangeSet
...
-      :reverse_symbol,
-      :ruler,
-      :series,
-      :shape,
-      :subject,
-      :workshop
-    ]
+    {
+      "" => [
+        :issue_number,
```
christinach: one question. why is this key empty?

escowles: To let you make a "default" section without a header: https://github.com/pulibrary/figgy/pull/2182/files#diff-f2e3b8d6bd878f3f8c406fc12a48ef01R8

eliotjordan:
This is very nice. I can see this being useful for other resource types.

eliotjordan:
I'm 👍, but this PR changes the edit page styling for all resources by adding a panel border around the metadata elements. I wonder if we should get opinions about this from the rest of the team? Maybe I'm being overly cautious here, but it is a noticeable change.

escowles:
@eliotjordan It's good to make sure everybody is OK with the styling change — here's a screenshot of a flat form to show what they look like:

screen shot 2018-11-15 at 7 31 03 pm

PR #2191, opened 2018-11-16
No instance variables in ChangeSetPersister

### LINKED ISSUE #2190: ChangeSetPersisters are not thread-safe
The problem is this:

For a single ChangeSetPersister in memory, assume I save two resources, each in their own transaction, and they add files. Everything will work fine, except that `buffer_into_index` creates child change set persisters with their own `delayed_queue` and `created_file_sets` which then override the parent's delayed_queue. That happens here: https://github.com/pulibrary/figgy/blob/master/app/change_set_persisters/change_set_persister.rb#L105

We need to make it so that I can save multiple change sets through the same persister, have transactions work, and they won't run into one another.

tpendragon:
Fixes thread-safety when saving multiple resources through the same
ChangeSetPersister. Closes #2190

I can't figure out how to add a test to this.

### DIFF
--- a/app/change_set_persisters/change_set_persister.rb
+++ b/app/change_set_persisters/change_set_persister.rb
@@ -99,11 +99,11 @@ class ChangeSetPersister
     end
 
     def buffer_into_index
+      delayed_queue = nil
       metadata_adapter.persister.buffer_into_index do |buffered_adapter|
         with(metadata_adapter: buffered_adapter) do |buffered_changeset_persister|
           yield(buffered_changeset_persister)
-          @created_file_sets = buffered_changeset_persister.created_file_sets
-          @delayed_queue = buffered_changeset_persister.delayed_queue
+          delayed_queue = buffered_changeset_persister.delayed_queue
         end
       end
       delayed_queue.run


app/change_set_persisters/change_set_persister.rb
```diff
@@ -103,11 +103,12 @@ def buffer_into_index
         yield self
         return
       end
+
+      delayed_queue = nil
       metadata_adapter.persister.buffer_into_index do |buffered_adapter|
         with(metadata_adapter: buffered_adapter) do |buffered_changeset_persister|
           yield(buffered_changeset_persister)
-          @created_file_sets = buffered_changeset_persister.created_file_sets
```
escowles: Are we dependent on the `@created_file_sets` variable being set when the callbacks are run, e.g., `after_commit`: https://github.com/pulibrary/figgy/pull/2191/files#diff-f98978ff0ab2af320d010f45a93af6ceR214

tpendragon: I thought we were, but all the tests pass.

tpendragon: Upload works in development too. I think because the handlers get the created file sets passed to them.

PR #2211, opened 2018-11-19
Ensures that the authorization tokens are passed to the ManifestBuilder and inserted into FileSet URLs

### LINKED ISSUE #2205: Manifest should have download links with auth token when present in URL
Without it, the manifest loads in a viewer but the media can't play (it all redirects to our login page).

jrgriffiniii:
Resolves #2205

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -42,7 +42,9 @@ class Ability
     end
     can :download, FileSet do |resource|
       authorized = geo_file_set?(resource)
-      token_readable?(resource.decorate.parent) || authorized if auth_token
+      authorized_by_token = proxy_parent_readable?(resource) || token_readable?(resource.decorate.parent) if auth_token
+
+      authorized_by_token || authorized
     end
     can :color_pdf, curation_concerns do |resource|
       resource.pdf_type == ["color"]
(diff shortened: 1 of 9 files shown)


app/models/ability.rb
```diff
@@ -221,12 +223,24 @@ def find_by(id:)
...
 
+    def find_proxy_file_sets(resource)
+      query_service.find_inverse_references_by(resource: resource, property: :proxied_file_id)
+    end
+
+    def proxy_parent_readable?(resource)
+      proxies = find_proxy_file_sets(resource)
+      return if proxies.empty?
+
+      proxy = proxies.first
```
tpendragon: I don't think this will work if there are two ProxyFileSets in two playlists pointing to the same FileSet.

app/resources/playlists/playlists_controller.rb
```diff
@@ -25,10 +25,20 @@ def manifest
...
     render json: { message: "No manifest found for #{params[:id]}" }
   end
+
+  private
+
+    def manifest_builder
+      if params[:auth_token]
+        ManifestBuilder.new(@resource, params[:auth_token])
+      else
+        ManifestBuilder.new(@resource)
```
tpendragon: Is there any harm in just passing it in if it's nil?

PR #2214, opened 2018-11-20
Adds numismatic_finds

### LINKED ISSUE #2127: Numismatic Finds
Models/etc. for numismatic finds (locations where coins were found), linked to from Coins.

Attributes from db:
* Place
* date
* find number (auto-incrementing sequence)
* feature
* locus
* description

N.b., there are only 4 rows in the database.

Child of #2025

christinach:
Closes #2127

### DIFF
--- a/app/resources/numismatic_finds/numismatic_find.rb
+++ b/app/resources/numismatic_finds/numismatic_find.rb
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 class NumismaticFind < Resource
   include Valkyrie::Resource::AccessControls
-  attribute :member_ids, Valkyrie::Types::Array
 
   attribute :place
   attribute :date
-  attribute :find_number, Valkyrie::Types::Anything
+  attribute :find_number, Valkyrie::Types::Int
   attribute :feature
   attribute :locus
   attribute :description
--- a/app/resources/numismatic_finds/numismatic_find.rb
+++ b/app/resources/numismatic_finds/numismatic_find.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+class NumismaticFind < Resource
+  include Valkyrie::Resource::AccessControls
+  attribute :member_ids, Valkyrie::Types::Array
+
+  attribute :place
+  attribute :date
+  attribute :find_number, Valkyrie::Types::Anything
+  attribute :feature
+  attribute :locus
+  attribute :description
+
+  def title
+    ["Find: #{find_number}"]
+  end
+end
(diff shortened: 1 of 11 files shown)


app/resources/numismatic_finds/numismatic_find.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+class NumismaticFind < Resource
+  include Valkyrie::Resource::AccessControls
+  attribute :member_ids, Valkyrie::Types::Array
```
escowles: I don't think we need `member_ids`, because the Coins should link to the Finds instead of the other way around.  As with NumismaticAccessions, we could model the Coins of members, but we generally avoid having an object be a member of more than one parent if we can avoid it.

app/resources/numismatic_finds/numismatic_find.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  attribute :place
+  attribute :date
+  attribute :find_number, Valkyrie::Types::Anything
```
escowles: Please make this a `Valkyrie::Types::Int` to make sure it's stored as a number instead of as a string, to make querying for it easier.

app/resources/numismatic_finds/numismatic_find_change_set.rb
```diff
@@ -0,0 +1,24 @@
...
+  property :locus, multiple: false, required: false
+  property :description, multiple: false, required: false
+
+  validates_with AutoIncrementValidator, property: :find_number
+
+  def primary_terms
+    [
+      :place,
+      :date,
+      :find_number,
```
escowles: I think we don't want users to be able to edit the `find_number`, so it shouldn't be in here.

app/resources/numismatic_finds/numismatic_find.rb
christinach: Do you mean NumismaticCitations instead of NumismaticAccessions? The NumismaticAccessions have `member_ids`

christinach:
Thank you for reviewing!

app/resources/numismatic_finds/numismatic_find.rb
escowles: Oh, NumismaticAccessions shouldn't have `member_ids` — I'll make a ticket to remove them.

christinach: oh ok. thanks.

christinach:
@escowles is this ok now?

Issue #2175, opened 2018-11-15
Accession form reorg


escowles:
Reorganize the Numismatic Accession form to match the order of the existing application:
* accession number
* number in accession
* accession date
* number of items
* accession type
* cost
* account
* acquired from person
* acquired from firm
* info
* private info
* references

Child of #2025

christinach:
where is the 'number of items' coming from?

escowles:
For now, let's assume they are going to enter the number of items in the gift/purchase manually.  We may want to calculate this based on the number of coins that are created.

christinach:
ok then I will add it in the PR.

christinach:
Does it make sense the number (in the accession) and the items_number to be Valkyrie::Types::Int ?

escowles:
I think so, but I would look at the data to see if they have any other kinds of data in there.  It sounds like just a simple count, but maybe they have "about 500" or other estimates, too?

christinach:
I thought it was just a number but will check again. thanks

Issue #577, opened 2017-11-21
Add Regenerate Derivative functionality to Vue Filemanager


sdellis:

tpendragon:
@sdellis Necessary? Did this just get moved into the file manager and we can close this>

sdellis:
@tpendragon yeah, that's in file manager now. Closing...

PR #2227, opened 2018-11-20
Reorganize accession form

christinach:


christinach:
Closes #2175

app/resources/numismatic_accessions/numismatic_accession.rb
```diff
@@ -2,13 +2,15 @@
 class NumismaticAccession < Resource
   include Valkyrie::Resource::AccessControls
 
-  attribute :date
-  attribute :person
-  attribute :firm
   attribute :accession_number, Valkyrie::Types::Int
+  attribute :number, Valkyrie::Types::Int
```
escowles: My reading of the screenshot is that the white part is in the coin, and the gray part is the accession record itself.  So this number property probably goes in the coin model (which would have `:accession_number` and `:number_in_accession` properties to record the accession and the coin's number in the accession).

christinach: I don't think the number in accession (in the screenshot) 1317 is the coin's number. I think the 818 is the items_number. (number of coins in this accession)

christinach: I will check the db for the 1317. I don't understand yet what this is.

christinach:
```1> select * from accessions where accessionid = 6
2> go
AccessionID	AccDate	AccPersonID	AccFirmID	AccNumber	AccType	AccRefID	Cost	Account	Info	PrivateInfo
6	Jan  1 1965 12:00:00:000AM	16	NULL	818	gift	0	0.00	NULL	NULL	Accession Book 5, #577-1244, 1258-1405
``` based on this one the 818 is the number_in_accession which I added as items_number but I can update it.

escowles: I'm looking at the screenshot and here's what I think the numbers are:
* Accession: the coin's link to the accession record
* Number in accession: the coin's identifier within the accession's catalog
* No. of items: the number of items in the accession

christinach: yes this is correct

christinach: Why would you assign an accession id in a coin but not a number_in_accession?

escowles: I'm not sure all accessions have their own numbers for their contents.  E.g., if you buy one coin, or a handful of coins, is there even a list?

christinach: For this PR I think I need to keep the `number_in_accession`, in `accessions` and add it also in the `coins` since both of them have it as an identifier. Sometimes it's null which I don't understand yet why. Maybe as you mentioned they add a `number_in_accession` only if they want to group (list) coins in a specific accession. I will also keep the item_number in accession. Does it make sense?

escowles: Yes, I think that makes sense.  `item_number` is the count of items/coins in the accession, and `number_in_accession` is the identifier of the coin in the accession.

christinach:
@escowles if you have a chance can you take a look again?

PR #2253, opened 2018-11-29
Don't check members twice in ManifestBuilder.

tpendragon:
Should halve our manifest build times, bringing them back to where they
were before playlists.


tpendragon:
> Did they need to be pulled out of private to make the test work?

Yes

Issue #2265, opened 2018-12-04
Resolve music reserve items with call numbers beginning "x-"


hackmastera:
We assume each track in such a "recording" exists in another recording with a regular call number. Should have the same filename.

child of #2043

hackmastera:
Also, why did they do this?

PR #2282, opened 2018-12-06
Remove unused `duplicate` field in MRRecording struct

hackmastera:
NOTE: This will invalidate your cache


hackmastera:
@tpendragon do you think we will use this field somehow in the resolution of https://github.com/pulibrary/figgy/issues/2265? if so i'd be fine keeping it in. I mostly thought to remove it so it won't show up and confuse people on the CSV.

tpendragon:
I could see using it. We could also use the recommended bib, though, and just know to make it a Playlist if it starts with x.

hackmastera:
okay let's leave it in; we can always remove it later and if anyone's confused they'll just ask us and we'll tell them.

PR #2283, opened 2018-12-06
Add MusicImportService which produces a bib ids report

### LINKED ISSUE #2139: Try to resolve call numbers to Bib-IDs for MusicReserves
Child of #2043 

See if we can find bib-ids for all call numbers in course numbers that have numbers in them (this means it's a course, and not like the Princeton University Orchestra.)

If you can't resolve the call number, check to see if there are audio files attached to that recording which are attached to another recording and DOES have a valid call number. That means it's a playlist.

You can search for call numbers like this: https://catalog.princeton.edu/browse/call_numbers.json?search_field=browse_cn&q=cd-+2016-03-05. Know that the third one is the match.

An alternative to try might be https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&f[call_number_browse_s][]=CD-%202016-10-23, but it's case sensitive. CD -should- always be capitalized, I'd just try capitalizing everything.

Acceptance Criteria:

1. How many call numbers in all of the Recordings table can be converted to bib-ids
2. How many call numbers recordings which are in numbered courses can be resolved vs un-numbered courses.
3. How many "fake" call numbers are duplicate recordings with the same audio files that we can convert to Playlists.

Include examples of failures for all three above.

hackmastera:
resolves #2139


hackmastera:
whenever this is merged let's do it as a "squash and merge"

app/services/music_import_service/recording_collector.rb
```diff
@@ -0,0 +1,334 @@
...
+    json["response"]["docs"].map { |doc| doc["id"] } # will be [] if no results
+  end
+
+  # apply normalization which replaces a space with a hyphen
+  def space_after_hyphen(call_number)
+    return unless call_number
+    general_normalization(call_number) do |cn|
+      # dashes should be followed by spaces e.g. cd-10994
+      # ls, cass, dat, cd, vcass, dvd
+      if cn.match?(/^((CD)||(DAT)||(CASS)||(LS)||(VCASS)||(DVD))-/)
```
escowles: This is repeated a bunch of times — extract to method?

app/services/music_import_service/recording_collector.rb
```diff
@@ -0,0 +1,334 @@
...
+          results
+        end
+      end
+  end
+
+  def populate_bib_from_title(recording)
+    logger.info "populating bib from title for #{recording.titles.first}"
+    ol_response = JSON.parse(
+      open(
+        "https://catalog-staging.princeton.edu/catalog.json?f[access_facet][]=In+the+Library&f[format][]=Audio&f[location][]=Mendel+Music+Library" \
```
escowles: The other URLs are the production URLs — change this one to match?

bin/ci_free_tds_install.sh
```diff
@@ -0,0 +1,11 @@
+if [ ! -d "tmp/freetds-1.00.108" ]; then
+  mkdir ~/downloads
```
escowles: Can we put the tarball in `tmp` instead of creating a new directory?

hackmastera: I modeled this on https://github.com/pulibrary/figgy/blob/master/bin/ci_simple_tiles_install.sh and the kakadu one also does it this way. do we want a ticket to change all of these?

app/services/music_import_service/recording_collector.rb
hackmastera: or should this be configurable? Or keyed off of the selected db? @tpendragon

app/services/music_import_service/recording_collector.rb
hackmastera: I'll do it this

app/services/music_import_service/recording_collector.rb
tpendragon: Configurable seems easiest

hackmastera: @tpendragon like add a param to the initializer or use an actual config file?

tpendragon: A param in the initializer seems right.

hackmastera: okay great i agree. i'll take care of it

hackmastera: defaulted it to production.

hackmastera:
@escowles requested changes have been made

hackmastera:
ran locally; coverage is now at 100%

PR #2304, opened 2018-12-12
Add ingest_folder to scanned_resources_controller

### LINKED ISSUE #2204: Save and Ingest should look in the music library's ingest directory when ingesting a Recording
Add a way to have configured "Save and Ingest" location change based on change set.

requires #2295

### LINKED ISSUE #2230: Extract title of track from wav file on characterization
We can make the lives of the ingesters easier this way, I think.

hackmastera:
fixes #2204 

    I tried to do this by putting the folder in the change set itself as a
    class property, but was a bit put off by the need to put it in both
    ChangeSet and DynamicChangeSet. So I just went with the example set in
    numismatics.


tpendragon:
@hackmastera I think the title for this PR is wrong, can you fix it?

hackmastera:
that is so weird i don't know how that happened. @tpendragon fixed

PR #2310, opened 2018-12-13
Ensures that the OrderManager only attempts to render resource members with valid thumbnail URLs

### LINKED ISSUE #2305: Order Manager errors for thumbnails need to be isolated
Otherwise invalid thumbnail properties (e. g. missing `iiifServiceUrl` in the response to GraphQL queries for https://figgy.princeton.edu/concern/scanned_resources/2ad03c65-1303-4009-8689-ee474dc41ade/order_manager) will break the Order Manager

jrgriffiniii:
Resolves #2305


tpendragon:
Shouldn't it still display them, but just without a thumbnail?

jrgriffiniii:
I can restructure this, but it assumes that all thumbnails exist for FileSets, and that all thumbnails have valid URLs to IIIF image servers (https://figgy.princeton.edu/concern/scanned_resources/2ad03c65-1303-4009-8689-ee474dc41ade/order_manager was what brought this to my attention).

tpendragon:
I mean, resources that don't have thumbnails work too - like I can load an order manager for audio files.

jrgriffiniii:
I've added a default thumbnail image for cases where the IIIF image server URL cannot be resolved:

figgy_issues_2310_screenshot_0

One can also view this on the branch `issues-2305-jrgriffiniii-order-mgr-thumbnail-errors-test` (where I've ensured that all GraphQL-retrieved resource thumbnails are `null`)

jrgriffiniii:
As brought to my attention by @eliotjordan, these modifications should also be tested with Jest: https://github.com/pulibrary/figgy/tree/master/app/javascript/test

PR #2341, opened 2018-12-20
Add archival_collection_code to scanned resource

### LINKED ISSUE #2339: Add field for pulfa collection id
Archives staff want to search by this value.

Also, we need it in order to match figgy records to updated eads (which are one per collection).

1 idea is to not display it or allow it to be edited but extract it from source_metadata_identifier.

hackmastera:
populate it from source_metadata_identifier
fixes #2339

### DIFF
--- a/app/change_set_persisters/change_set_persister/extract_archival_collection_code.rb
+++ b/app/change_set_persisters/change_set_persister/extract_archival_collection_code.rb
@@ -10,25 +10,25 @@ class ChangeSetPersister
     def run
       return unless change_set.respond_to?(:source_metadata_identifier)
       return unless change_set.model.respond_to?(:archival_collection_code)
-      return unless updated_values
+      return unless updated_values?
       return if PulMetadataServices::Client.bibdata?(change_set.source_metadata_identifier)
       change_set.model.archival_collection_code = extract_collection_code(change_set.source_metadata_identifier)
       change_set
     end
 
+    def extract_collection_code(pulfa_id)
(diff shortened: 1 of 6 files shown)


hackmastera:
don't merge; I forgot I need a data migration task

hackmastera:
added migration task; tested locally and it updated pulfa objects but not voyager objects. Also double-checked locally that the new field is searchable. ready for review if it's green

app/change_set_persisters/change_set_persister/extract_archival_collection_code.rb
```diff
@@ -0,0 +1,34 @@
...
+      # either the source metadata identifier has changed or we're refreshing remote metadata
+      def updated_values
+        change_set.changed?(:source_metadata_identifier) || (
+          change_set.respond_to?(:apply_remote_metadata?) &&
+          change_set.apply_remote_metadata?
+        )
+      end
+
+      def extract_collection_code(pulfa_id)
+        m = pulfa_id.match(/^([a-zA-Z]+[0-9]+)_.*/)
```
escowles: I don't think this regex will match all of the possible EAD variants, which include:
* `RCPXG-5830371.2`
* `C0744.04`
* `C0723.1-47`
* `C0723.306e`
* `MC001.03.03`
* `MC197.r16845`

Maybe something like `/^([A-Za-z0-9.-]+)_c.*/` would work for everything?

hackmastera: okay thanks for catching this. I added a test to show all those cases.

PR #2346, opened 2018-12-21
Link geo resource ARKs to Pulmap

### LINKED ISSUE #1898: ARKs for maps should point to pulmap

eliotjordan:
Closes #1898

### DIFF
--- a/app/services/identifier_service.rb
+++ b/app/services/identifier_service.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 class IdentifierService
   def self.mint_or_update(resource:)
-    if resource.identifier.present?
+    if resource.respond_to?(:geo_resource?) && resource.geo_resource?
+      mint_or_update_geo_resource(resource)
+    elsif resource.identifier.present?
       update_metadata resource
     else
       mint_identifier resource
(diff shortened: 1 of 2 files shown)


app/services/identifier_service.rb
```diff
@@ -47,6 +49,22 @@ class RestrictedArkError < StandardError; end
     "http://findingaids.princeton.edu/collections/#{resource.source_metadata_identifier.first.tr('_', '/')}"
   end
 
+  private_class_method def self.geo_metadata(resource)
+    slug = "princeton-#{resource.identifier.first.gsub(%r(ark:/\d{5}/), '')}"
+    metadata(resource).merge(target: "https://maps.princeton.edu/catalog/#{slug}")
```
escowles: Is this just customizing the URL for the Geo resources?  Couldn't this be done in the `url_for` method? 
Or am I missing something here?

eliotjordan: Generating the slug for the Pulmap URL requires the ARK identifier to be minted first. The code here mints the ARK using the default target URL and then updates the URL to the correct value afterwards.

escowles: 👍 OK, that makes sense.

PR #2363, opened 2019-01-03
Remove "update" form from file set page file listing

### LINKED ISSUE #1091: Remove "update file" from FileSet show page
It doesn't work.

hackmastera:
fixes #1091


hackmastera:
now looks like this
screen shot 2019-01-03 at 3 37 31 pm

PR #2378, opened 2019-01-04
Implementing the FindManyByStringProperty custom query and GraphQL ResourcesByBibIds query

jrgriffiniii:
This provides a GraphQL query for retrieving multiple resources for a set of bib. IDs (this will be necessary for resolving pulibrary/orangelight/issues/1534).  This ensures that only a single query must be made using GraphQL in order to retrieve a set of thumbnail IDs for a set of bib. IDs.

### DIFF
--- a/spec/queries/find_many_by_string_property_spec.rb
+++ b/spec/queries/find_many_by_string_property_spec.rb
@@ -1,0 +1,27 @@
+
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe FindManyByStringProperty do
+  subject(:query) { described_class.new(query_service: query_service) }
+  let(:box) { FactoryBot.create_for_repository(:ephemera_folder, barcode: "1234567") }
+  let(:box2) { FactoryBot.create_for_repository(:ephemera_folder, barcode: "2345678") }
+  let(:query_service) { Valkyrie.config.metadata_adapter.query_service }
+
+  describe "#find_many_by_string_property" do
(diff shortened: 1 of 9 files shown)


spec/queries/find_many_by_string_property_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  let(:box) { FactoryBot.create_for_repository(:ephemera_folder, barcode: "1234567") }
+  let(:box2) { FactoryBot.create_for_repository(:ephemera_folder, barcode: "2345678") }
+  let(:query_service) { Valkyrie.config.metadata_adapter.query_service }
+
+  describe "#find_many_by_string_property" do
+    it "can find objects with strings in it by a property" do
+      output = query.find_many_by_string_property(property: :barcode, values: [box.barcode.first, box2.barcode.first])
+      expect(output.length).to eq 2
+      expect(output.first.id).to eq box.id
+      expect(output.last.id).to eq box2.id
```
escowles: Are these always going to be returned in order?  Or should we make this order-neutral to avoid failures if they come back in the other other?

tpendragon:
Why do we need to find by string property?

escowles:
@tpendragon Is there another query we could use to find Figgy objects related to multiple bib IDs?

tpendragon:
Yeah, you could run a graphql query that runs the query multiple times. This is maybe more efficient though, since it's fewer queries too.

PR #2379, opened 2019-01-04
Fix logical structure for scanned maps

### LINKED ISSUE #2377: Logical structure does not work for MapSets
Because of the way in which ScannedMaps are modeled, the structure manager saves the logical structure for MapSets in way that generates broken manifests. Specifically, the canvas id for a range with a ScannedMap does not match the id of the corresponding canvas.

eliotjordan:
Closes #2377

### DIFF
--- a/app/resources/scanned_maps/scanned_map_wayfinder.rb
+++ b/app/resources/scanned_maps/scanned_map_wayfinder.rb
@@ -12,7 +12,7 @@ class ScannedMapWayfinder < BaseWayfinder
     @geo_members ||=
       begin
         members.select do |member|
-          next unless member.respond_to?(:mime_type)
+          next unless member.respond_to?(:mime_type) && member.mime_type
           ControlledVocabulary.for(:geo_image_format).include?(member.mime_type.first)
         end
       end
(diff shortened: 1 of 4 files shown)


app/resources/scanned_maps/scanned_maps_controller.rb
```diff
@@ -3,4 +3,13 @@ class ScannedMapsController < ScannedResourcesController
...
   include GeoblacklightDocumentController
   self.resource_class = ScannedMap
+
+  # View the structural metadata for a given repository resource
+  def structure
+    @change_set = change_set_class.new(find_resource(params[:id])).prepopulate!
+    @logical_order = (Array(@change_set.logical_structure).first || Structure.new).decorate
+    members = Wayfinder.for(@change_set.resource).logical_structure_members
+    @logical_order = WithProxyForObject.new(@logical_order, members)
+    authorize! :structure, @change_set.resource
```
escowles: Should the `authorize!` call come as early as possible (presumably right after the @change_set is instantiated)?

app/resources/scanned_maps/scanned_map_wayfinder.rb
```diff
@@ -34,4 +34,21 @@ def members_with_parents
...
+
+  # To display and save logical order correctly, this method replaces ScannedMap members that have no ScannedMap children
+  # with their corresponding geo member (TIFF) FileSet. Geo resources are unusual in that each resource has one corresponding
+  # geo member. An atlas with multiple maps, for example, will be a set of multiple ScannedMaps with each resource corresponding
+  # to a single map. ScannedMap members with ScannedMap children are analogous to a volume in a multi-volume work.
+  def logical_structure_members
+    @logical_structure_members ||= begin
+      members_with_parents.map do |member|
+        decorator = member.decorate
+        if decorator.respond_to?(:scanned_map_members) && decorator.scanned_map_members.empty?
```
escowles: Pulling this out to a separate method would probably let you avoid the clunky `||= begin` syntax.

app/resources/scanned_maps/scanned_maps_controller.rb
eliotjordan: Makes sense. This method was copied from the scanned resource controller and modified slightly. I can update that one as well.

app/resources/scanned_maps/scanned_map_wayfinder.rb
eliotjordan: Yeah, it's it a little gnarly. I'll refactor.

eliotjordan: Refactored to a new method, although the comment doesn't show as `outdated` for some reason.

PR #2385, opened 2019-01-07
Change label for template title to "Template Name"

### LINKED ISSUE #1564: Call template title "Template Name" instead of "Title"
When creating a new ephemera folder template, the first field is the template title, with the label "Title".  There is also the main title field, also with the label "Title" further down the page.  The template title should have the label "Template Name"

hackmastera:
fixes #1564


hackmastera:
looks like this 
screen shot 2019-01-07 at 2 38 08 pm

PR #2386, opened 2019-01-07
Ingest Recordings

tpendragon:
This branch was used to migrate MUS-234 into staging. It can handle the best case scenario for migrations, but can't yet handle "recordings used as playlists" or "recordings which don't have bib-ids."


tpendragon:
Yeah, it's not the prettiest, but it was the quickest way I could find to use all the bib-id finding logic for a subset of recordings.

PR #2392, opened 2019-01-08
Redirect the user back to the page they were on after logging in

### LINKED ISSUE #2390: When logging in, I am not sent back to the URL I was on before
When I am on a Figgy page and click the login button, I am taken to CAS, but then I am sent back to the home page instead of the page I was on.  I should be redirected to the page I was on.

escowles:
Fixes #2390

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,10 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
 
+  def after_sign_in_path_for(_resource)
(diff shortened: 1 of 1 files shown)


escowles:
@jrgriffiniii It worked when I was already logged in to CAS, but when I tried in a private window, I'm also being redirected to the home page.  I'll see if I can figure out why.

escowles:
I haven't made any changes, but I did redeploy my branch on staging, and now it's working for me in a private window.  Can you try it again?

PR #2396, opened 2019-01-08
Add link back to recording on playlist members

### LINKED ISSUE #2315: Playlist tracks should have a link to the recording they're from
It's hard to tell where the source of a recording is right now.

hackmastera:
fixes #2315

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -103,6 +103,7 @@ Metrics/LineLength:
   Exclude:
     - 'spec/views/scanned_resources/order_manager.html.erb_spec.rb'
     - 'lib/tasks/music_reserves.rake'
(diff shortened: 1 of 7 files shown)


hackmastera:
i just made the file label into a link to the recording. does this seem sufficient or do we want a separate column?

tpendragon:
Either a separate column or in parenthesis with the recording name would be my preference

hackmastera:
I tried it with an addt'l column but there's an expectation in styles that there only be 2 columns, so the new column is smushed up against the "detach" button. putting it in parens instead, so as not to disrupt other member tables / get lost in special-case css land. looks like 
screen shot 2019-01-09 at 10 18 45 am

PR #2397, opened 2019-01-08
Extract course numbers for selections on ingest.

tpendragon:


spec/change_set_persisters/change_set_persister_spec.rb
```diff
@@ -1439,18 +1439,19 @@
     it "creates ProxyFileSets" do
```
hackmastera: add "preserving the order of files on the playlist"

tpendragon:
> This seems good if you can adjust the spec to mention the order thing you appear to be testing for.
> 
> I'm confused about the order, though; did we not have to use the sort_order data from the reserves db?

Added the comment.

As far as I can tell, the audio files are in order by virtue of their ID.

PR #2405, opened 2019-01-09
Revert "Use embedded title for file label and FileSet title on media files"

### LINKED ISSUE #2307: How should we manage system-generated vs. user-provided titles for Recordings?
Examples:
* User imports Recordings via "save and ingest". System titles them `1`, `2`, `3`, ... then overrides that title with the one extracted from embedded metadata. (probably good)
* We bulk import data from the music reserves db. We migrate the title of each fileset. Then those are overridden by the one extracted from embedded metadata. (probably bad)
* We ingest an archival a/v bag. We name the fileset based on its actual on-disk filename, which is a barcode with side / part info. Then those are overridden by the one extracted from embedded metadata, which was put in by vendors based on the shot list they were provided. (good? bad?)

Options for what to do
* Allow a flag for "override title with the one from embedded metadata" this will be annoying to pass around.
* Have a switch on ingest form/s which allows users to say whether to use the embedded title
* Store both titles. Only display the title retrieved from embedded metadata ("system-generated-title") if there is no user-provided title. Will require that we allow title to be blank. May have other implications we haven't thought of. Similar to the embedded metadata business.

hackmastera:
This reverts commit 446d585961960f9909cd76a88f010ec4ddb0f470.
fixes #2307


hackmastera:
oh there was a follow-up commit i have to revert, too.

PR #2414, opened 2019-01-11
Import Music Reserve Courses

### LINKED ISSUE #2265: Resolve music reserve items with call numbers beginning "x-"

tpendragon:
Adds following functionality:

* Don't reimport an existing recording
* Handle recordings which are really playlists.
* Handle "nested" fake playlists.
* Fix order of audio file import
* Rename playlist audio files appropriately.

Closes #2265


spec/services/music_import_service_spec.rb
```diff
@@ -138,6 +141,163 @@
...
+        playlists = Valkyrie::MetadataAdapter.find(:indexing_persister).query_service.find_all_of_model(model: Playlist)
+        expect(playlists.length).to eq 2
+        expect(playlists.flat_map(&:title)).to contain_exactly "Week 1", "My Selection"
+        playlist = playlists.find { |x| x.title.first == "Week 1" }
+        members = Wayfinder.for(playlist).members
+        expect(members.first.label).to eq ["First File"]
+        expect(members.first.local_identifier).to eq ["54204"]
+      end
+    end
+    context "when a dependent is a fake playlist" do
```
hackmastera: I don't really understand what this means. what is a dependent exactly? and what is the relationship here?

app/services/music_import_service.rb
```diff
@@ -118,23 +195,29 @@ def import!
...
       @change_set ||= RecordingChangeSet.new(resource).prepopulate!
     end
 
     def identifier
-      recording.bibs.first
+      recording.bibs.first || recording.recommended_bib
+    end
+
+    def all_audio_files
+      @all_audio_files = recording_collector.audio_files(recording)
```
hackmastera: did you intend to memoize this?

tpendragon:
> It could use some clarification of nested fake playlists / dependent fake playlists in the spec and/or in the code. Do we know why these exist? Why do we chose not to ingest them? What are the implications of not ingesting them? Does that mean there is a file missing from the final playlist?

I'll add a comment explaining what's going on (or trying to). Basically it's not that they're nested fake playlists, it's that it creates playlists by iterating over audio files, and the same audio file might be in two fake playlists and a real recording. In that case, it doesn't ingest the second fake recording as a dependent of the first fake recording, because it's actually a dependent of the real recording.

app/services/music_import_service.rb
tpendragon: Yes

PR #2478, opened 2019-01-24
Display visibility notice that takes both the visibility property and workflow into account

### LINKED ISSUE #682: Add large note to top about visibility and what the effects are?
Something like:

STATUS: PRIVATE - This item will not sync to discovery sites such as DPUL, Orangelight, PULMap, or LAE.

escowles:
Fixes #682


escowles:
I initially considered adding a banner, but that seemed obnoxious since it's displayed for every item, not just ones that aren't public.  So I opted for expanding the visibility badge with a color-coded notice about the final visibility:

Public:
public

Private because of visibility:
suppressed-visibility

Private because of workflow:
suppressed-workflow

tpendragon:
This seems fine to me, but we should keep track whether or not it helps the "why won't this sync" tickets. The idea behind the banner was that it was annoying and hard to miss.

hackmastera:
I like it. I think it will still stand out pretty significantly.

PR #2479, opened 2019-01-24
Adding Rake task to promote user to admin

escowles:


lib/tasks/dev.rake
```diff
@@ -44,6 +44,15 @@ if Rails.env.development? || Rails.env.test?
         end
       end
     end
+
+    desc "Promote last created user to admin"
+    task set_admin_user: :environment do
+      admin = Role.first || Role.create(name: "admin")
```
tpendragon: Shouldn't this be `Role.where(name: "admin").first`? Or even just `Role.find_or_create_by(name: "admin")`

escowles: 👍

PR #2480, opened 2019-01-24
Generate a default structure for AV materials.

### LINKED ISSUE #2467: Titles are not updated in the player when the track is changed

tpendragon:
Allows for a navigable table of contents, which updates the title of the
viewer.

Closes #2467

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -179,6 +179,6 @@ module ApplicationHelper
   # Generate the path for the Universal Viewer iframe @src attribute
   # @return [String]
   def universal_viewer_path(resource)
-    "/uv/uv#?manifest=#{manifest_url(resource)}"
+    "/uv/uv#?manifest=#{manifest_url(resource)}&config=#{root_url}/uv/uv_config.json"
   end
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -137,6 +137,7 @@ Metrics/MethodLength:
     - 'app/services/music_import_service/recording_collector.rb'
     - 'app/controllers/bulk_ingest_controller.rb'
     - 'config/initializers/sequel_active_support_notification.rb'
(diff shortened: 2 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -179,6 +179,6 @@ def manifest_url(resource)
   # Generate the path for the Universal Viewer iframe @src attribute
   # @return [String]
   def universal_viewer_path(resource)
-    "/uv/uv#?manifest=#{manifest_url(resource)}"
+    "/uv/uv#?manifest=#{manifest_url(resource)}&config=#{root_url}/uv/uv_config.json"
```
escowles: In `/viewer.html` we set the config explicitly:
https://github.com/pulibrary/figgy/blob/master/public/viewer.html#L12

But in `/uv/uv.html` we expect this param:
https://github.com/pulibrary/figgy/blob/master/public/uv/uv.html#L46

Can we just set the config explicitly in both places to make sure they are consistent, and always using our config?  Or is there some reason why we might want to override the config in some cases but not others?

escowles:
I'm 👍 on this pending my comment about the different config approaches.

app/helpers/application_helper.rb
tpendragon: If we want to change it we should switch to viewer.html, the other one is that way because it's how the file that ships with UV works, so it gets overridden when we update.

escowles: I think that would be good — that's what we're using from other apps, right?  Being consistent on which viewer HTML file we're using seems like it'll prevent bugs where the viewer works right some places but not others.

tpendragon: No, we use UV.html in the other apps. Shaun said he had some css problems with viewer.html.

tpendragon:
@escowles Created #2486. Good to go?

Issue #2446, opened 2019-01-18
Order Manager: Default when switching to foliation should be "f." instead of "p." and should pre-populate r and v


tpendragon:
Brought up at all-hands 1/19.

escowles:
This pre-populates r/v now, but does not change "p." to "f."

PR #2481, opened 2019-01-24
Set default foliation options

escowles:
Fixes most of #2446


escowles:
For some reason `$("#unitLabel").val("f. ")` wasn't updating the label box.

PR #2485, opened 2019-01-25
Increase space between save and delete buttons

### LINKED ISSUE #2456: Ephemera: Delete button is too close to other buttons
Separate delete buttons from other buttons to avoid accidentally clicking them

escowles:
Fixes #2456

### DIFF
--- a/app/assets/stylesheets/components/form.scss
+++ b/app/assets/stylesheets/components/form.scss
@@ -11,7 +11,7 @@
 }
 .panel-delete-controls {
   border-top-width: 0px;
-  padding-top: 0px;
+  padding-top: 1em;
 }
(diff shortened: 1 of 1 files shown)


christinach:
@escowles do we have a `confirm message` when the user clicks the delete button?

escowles:
@christinach Yes, but Gabrielle said it was annoying to have it so close, since she accidentally clicked it a few times, and was often worried about clicking it.

tpendragon:
I think we should deploy this to staging and show it to Gabrielle, because to me it doesn't seem like this'd fix the issue

escowles:
I've deployed my branch to staging and emailed Gabrielle to ask her if the update helps.

PR #2509, opened 2019-01-31
Adding visibility badge without verbose note, using in tabular member lists

### LINKED ISSUE #2507: Hide visibility note in related geo resources partial
The new visibility note appears in the geo resource related resources section and it looks terrible. Causes the panels to take up more room than they actually need.

screenshot 2019-01-30 23 40 42

escowles:
Fixes #2507


eliotjordan:
Thanks! 👏

PR #2512, opened 2019-01-31
Change Audio to use HLS

### LINKED ISSUE #2149: Add support for HLS
* What advantages does it have over just handing off to Nginx?
* How hard is it to support?

Child of #1986

tpendragon:
* Generates HLS instead of MP3s
* Speeds up clicking around viewer significantly.
* Regenerating derivatives will delete old MP3 derivatives.
* Passes auth tokens through the HLS manifest.

Closes #2149


app/controllers/downloads_controller.rb
```diff
@@ -13,6 +13,7 @@ def show
   end
 
   def send_content
+    return send_hls if file_desc.mime_type.first.to_s == "application/x-mpegURL" && params[:auth_token].present?
```
escowles: We're only serving HLS if there's an auth token? Are we not using it for the viewer when the user is logged into Figgy?

app/derivative_services/audio_derivative_service.rb
```diff
@@ -38,17 +38,54 @@ def valid?
...
+    change_set = DynamicChangeSet.new(output).prepopulate!
+    change_set.files = [build_file(hls_file, filename: "hls.m3u8")]
     change_set_persister.buffer_into_index do |buffered_persister|
       buffered_persister.save(change_set: change_set)
     end
   end
 
+  def generate_hls_derivatives(dir)
+    _stdout, _stderr, status =
+      Open3.capture3("ffmpeg", "-y", "-i", file_object.disk_path.to_s, "-hls_list_size", "0", "-hls_time", "10", "-f", "hls", "-codec:a", "libmp3lame", dir.join("hls.m3u8").to_s)
```
escowles: Do we want to move any of these settings to the config file like we do for kakadu?

app/resources/nested_resources/file_metadata.rb
```diff
@@ -42,6 +42,10 @@ def derivative?
     use.include?(Valkyrie::Vocab::PCDMUse.ServiceFile)
   end
 
+  def derivative_partial?
+    use.include?(Valkyrie::Vocab::PCDMUse.ServiceFilePartial)
```
escowles: Does this exist in valkyrie?  I don't see it in https://github.com/samvera-labs/valkyrie/blob/master/lib/valkyrie/vocab/pcdm_use.rb

app/controllers/downloads_controller.rb
tpendragon: We're only re-formatting the HLS if there's an auth token. Otherwise it just falls through to nginx and sends the file that's on disk, unmodified.

app/resources/nested_resources/file_metadata.rb
tpendragon: No, it's totally made up, because nothing else fit. Works because that vocab's not a strict one.

app/derivative_services/audio_derivative_service.rb
tpendragon: I don't think that would improve readability, personally. Seems easier to have the command here.

app/controllers/downloads_controller.rb
escowles: 👍 OK, that makes sense — `send_hls` below is rewriting the existing HLS file with auth_token links.

app/derivative_services/audio_derivative_service.rb
escowles: I'm fine with that, and I don't remember the thinking behind putting the jp2 recipes in the config file instead of in code like this.

tpendragon: It was just because that's how Hydra::Derivatives did it.

PR #2513, opened 2019-02-01
Increase HLS timeouts.

tpendragon:
Testing this on staging now.


tpendragon:
Confirmed this loads in staging at least.

PR #2515, opened 2019-02-01
Upgrade Figgy to Valkyrie 1.5.0.RC2

tpendragon:


app/indexers/sorting_indexer.rb
```diff
@@ -7,8 +7,7 @@ def initialize(resource:)
 
   def to_solr
     {
-      title_ssort: sort_title,
-      updated_at_dtsi: resource.updated_at
```
hackmastera: what is this bit about?

tpendragon: @hackmastera Valkyrie does `updated_at_dtsi` now.

PR #2517, opened 2019-02-04
Fix download permissions

### LINKED ISSUE #1568: Allow unauthenticated users to download full-resolution TIFFs
Curators say to provide the best possible scans to users.

### LINKED ISSUE #2458: Permissions not working correctly
All-hands feedback: `staff` users can't download TIFFs

escowles:
* allow anonymous users to download any file attached to a public resource
* allow staff to download all files

Fixes #1568; Fixes #2458


spec/models/ability_spec.rb
```diff
@@ -227,6 +228,7 @@
       is_expected.to be_able_to(:manifest, flagged_scanned_resource)
       is_expected.to be_able_to(:color_pdf, color_enabled_resource)
       is_expected.to be_able_to(:read, :graphql)
+      is_expected.to be_able_to(:download, other_staff_file)
```
hackmastera: can you use a different resource name here? the reference to staff file implies that maybe it's not public.

escowles: This resource is used several places in the tests, and the name is to contrast with `staff_file` where we test staff interacting with their own files vs. with files created by another staff member. 

I'd be happy to use `admin_file` instead, or create a new file called `open_file` if that seems like a clearer illustration.

hackmastera: Yes I think `open_file` would be clearer in this context!

escowles:
We're looking at parents because the requirement changed from "never let users download master files" to "let users download master files of public objects" — so we need to look at the state/visibility of the parent.  I thought it made the most sense to just use `can?(:read, parent)` for that, in case we change our logic around which objects are public.

spec/models/ability_spec.rb
escowles: 👍 i've pushed another commit that switches to `open_file`

PR #2523, opened 2019-02-04
Add clipboard.js and use it for "copy to clipboard" button on auth_link

### LINKED ISSUE #2429: Add a "clipboard" button to access links on playlists
This is a clipboard icon on the token access link for Playlists to make it easy to copy to Aeon.

hackmastera:
fixes #2429

### DIFF
--- a/app/javascript/packs/auth_link_clipboard.js
+++ b/app/javascript/packs/auth_link_clipboard.js
@@ -1,0 +1,12 @@
+import Clipboard from 'clipboard'
+
+document.addEventListener('DOMContentLoaded', () => {
+  let div = document.getElementById('clipboard-trigger-holder')
+  div.innerHTML = '<button class="btn btn-primary" id="clipboard-trigger">Copy link to clipboard</button>'
+  new Clipboard('#clipboard-trigger', {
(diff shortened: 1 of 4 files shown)


hackmastera:
screen shot 2019-02-04 at 5 16 27 pm

app/javascript/packs/auth_link_clipboard.js
```diff
@@ -0,0 +1,12 @@
+import Clipboard from 'clipboard'
+
+document.addEventListener('DOMContentLoaded', () => {
+  let div = document.getElementById('clipboard-trigger-holder')
+  div.innerHTML = '<button class="btn btn-primary" id="clipboard-trigger">Copy link to clipboard</button>'
```
escowles: Are you setting the button text here so it doesn't display if there's a problem loading the javascript? Or is there some other reason not to  include in the view HTML?

hackmastera: yeah, just sort of doing progressive enhancement I guess, since it only works if there's javascript.

escowles: 👍

PR #2530, opened 2019-02-05
Disable all Track detach buttons when one is clicked

### LINKED ISSUE #2528: Disable "detach" buttons after you click them in the Playlist members view.
Stops folks from accidentally removing multiple items with no feedback.

hackmastera:
closes #2528

### DIFF
--- a/app/javascript/components/playlist_member.vue
+++ b/app/javascript/components/playlist_member.vue
@@ -35,7 +35,14 @@ export default {
   },
   methods: {
     detach: function (event) {
+      // add spinner to button that was clicked
       this.detachClicked = true
+      // disable all detach buttons
+      let buttons = document.getElementsByClassName('detach-btn')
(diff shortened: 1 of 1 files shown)


hackmastera:
screen shot 2019-02-05 at 1 10 11 pm

christinach:
why are we disabling all the Detach buttons and not only the one which is detached?

hackmastera:
@christinach the whole page reloads when you click a button. it's not really intended to support multiple detach actions at once.

christinach:
I understand. Thank you @hackmastera . Maybe this is something that we can improve. -> not loading the whole page when detaching.

hackmastera:
@christinach for example, I just tried it with just disabling the one I click. I was able to click 3 buttons before the page reloaded, and see them all spin, but then when the reload came through, 2 of those had been removed and the 3rd was still there.

PR #2546, opened 2019-02-07
Decision: Queue priorities and guidelines

escowles:
Will need to be rebased after #2545


tpendragon:
@escowles Needs rebased.

PR #2550, opened 2019-02-07
Add bulk complete UI

### LINKED ISSUE #697: Bulk edit: workflow status complete, rights statement
e.g., complete all items in current search results, or in a collection
requires #1651 
Wireframes copied over from #1651:

Button above search results:
1651 bulk edit button

Bulk edit form:
1651 bulk edit form 1

### LINKED ISSUE #1860: write bulk_edit_controller#new and #edit
`#new` should expect a querystring like the one passed to a search, then embed that info into hidden form fields so that the edit action can update the correct set of records

child of #697 

As long as we don't add the button onto the search results page just yet, we should be able to develop this feature in small bits to the extent required.

### LINKED ISSUE #2494: Add Bulk Complete UI to search results
Child of #697

christinach:
- Bulk edit button displays on search results page when a collection facet
is selected.
- search params are passed around so we can retrieve the same results
set
- Bulk edit controller queues 1 update job per 50 resources

Closes #1860 
Closes #2494 
Closes #697 
Other functionality moved to #2586

### DIFF
--- a/app/controllers/bulk_edit_controller.rb
+++ b/app/controllers/bulk_edit_controller.rb
@@ -8,34 +8,32 @@ class BulkEditController < ApplicationController
   end
 
   def resources_update
-    batch_size = params["batch_size"] || 50
     args = {}.tap do |hash|
       hash[:mark_complete] = (params["mark_complete"] == "1")
     end
-    builder_params = { q: params["search_params"]["q"], f: params["search_params"]["f"] }
-    builder = search_builder.with(builder_params)
-    builder.rows = batch_size
-    resources_count = spawn_update_jobs(builder, args)
+    batches = prepare_batches
+    batches.each do |ids|
+      BulkUpdateJob.perform_later(ids: ids, args: args)
+    end
+    resources_count = batches.map(&:count).reduce(:+)
     flash[:notice] = "#{resources_count} resources were queued for bulk update."
     redirect_to root_url
   end
 
   private
 
-    # collects batches of ids then enqueues jobs
-    def spawn_update_jobs(builder, args)
-      batches = []
-      num_results = 0
-      loop do
-        response = repository.search(builder)
-        num_results = response["response"]["numFound"]
-        batches << response.documents.map(&:id)
-        break if (builder.page * builder.rows) >= num_results
-        builder.page += 1
+    # Prepare / execute the search and process into id arrays
+    def prepare_batches
+      builder_params = { q: params["search_params"]["q"], f: params["search_params"]["f"] }
+      builder = search_builder.with(builder_params)
+      builder.rows = params["batch_size"] || 50
+      [].tap do |batches|
+        loop do
+          response = repository.search(builder)
+          batches << response.documents.map(&:id)
+          break if (builder.page * builder.rows) >= response["response"]["numFound"]
+          builder.page += 1
+        end
       end
-      batches.each do |ids|
-        BulkUpdateJob.perform_later(ids: ids, args: args)
-      end
-      num_results
     end
 end
--- a/app/controllers/bulk_edit_controller.rb
+++ b/app/controllers/bulk_edit_controller.rb
@@ -32,6 +32,7 @@ class BulkEditController < ApplicationController
           response = repository.search(builder)
           batches << response.documents.map(&:id)
           break if (builder.page * builder.rows) >= response["response"]["numFound"]
+          builder.start = builder.rows * builder.page
           builder.page += 1
         end
       end
--- a/app/controllers/bulk_edit_controller.rb
+++ b/app/controllers/bulk_edit_controller.rb
@@ -3,18 +3,31 @@ class BulkEditController < ApplicationController
   include Blacklight::SearchHelper
 
   def resources_edit
-    (@solr_response, @document_list) = search_results(q: params["q"], f: params["f"])
+    (solr_response, _document_list) = search_results(q: params["q"], f: params["f"])
+    @resources_count = solr_response["response"]["numFound"]
   end
 
   def resources_update
-    (_, document_list) = search_results(q: params["search_params"]["q"], f: params["search_params"]["f"])
-    ids = document_list.map(&:id)
+    batch_size = params["batch_size"] || 50
     args = {}.tap do |hash|
       hash[:mark_complete] = (params["mark_complete"] == "1")
     end
-    ids.each_slice(50) do |batch|
-      BulkUpdateJob.perform_later(batch, args)
-    end
+    builder_params = { q: params["search_params"]["q"], f: params["search_params"]["f"] }
+    builder = search_builder.with(builder_params)
+    builder.rows = batch_size
+    spawn_update_jobs(builder, args)
     redirect_to root_url
   end
+
+  private
+
+    def spawn_update_jobs(builder, args)
+      loop do
+        response = repository.search(builder)
+        ids = response.documents.map(&:id)
+        BulkUpdateJob.perform_later(ids, args)
+        break if (builder.page * builder.rows) >= response["response"]["numFound"]
+        builder.page += 1
+      end
+    end
 end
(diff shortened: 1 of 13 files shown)


christinach:
closes #697

hackmastera:
screen shot 2019-02-12 at 3 55 01 pm

hackmastera:
screen shot 2019-02-12 at 4 00 17 pm

app/controllers/bulk_edit_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+class BulkEditController < ApplicationController
+  include Blacklight::SearchHelper
+
+  def resources_edit
+    (@solr_response, @document_list) = search_results(q: params["q"], f: params["f"])
+  end
+
+  def resources_update
+    (_, document_list) = search_results(q: params["search_params"]["q"], f: params["search_params"]["f"])
```
escowles: Does this retrieve the whole set of results, or just the first page?

hackmastera: good question; we will look into this.

logfile
```diff
@@ -0,0 +1,2 @@
+2019-02-11 10:48:40.115 EST [76779] FATAL:  lock file "postmaster.pid" already exists
+2019-02-11 10:48:40.115 EST [76779] HINT:  Is another postmaster (PID 501) running in data directory "/usr/local/var/postgres"?
```
escowles: Looks like a stray logfile got added by mistake.

app/controllers/bulk_edit_controller.rb
tpendragon: I'd recommend adding a FL key with "id" too if that works through BL.

logfile
christinach: probably that was mine

app/controllers/bulk_edit_controller.rb
hackmastera: can't seem to provide `fl` to the search builder.

christinach:
screen shot 2019-02-13 at 5 29 05 pm

escowles:
In manually testing this, I'm seeing some (but not all) of my objects get completed.  For example, I bulk completed 76 objects in staging, and 22 were left pending.  I created 150 pending objects in my local dev instance, and 65 were completed, and 85 were left pending.

I can try this again, but I think it might make sense to take a look at the search logic to see if objects being completed are changing the search results as we page through them.

hackmastera:
@escowles i would totally believe that is happening. because it queues the batch update job and then queries solr for the next page.

I think we can resolve this by creating all the jobs first, then queuing them. But I'm not sure how to write a regression test for a race condition like this.

escowles:
@hackmastera Yes, I agree it would be very hard to test — I'd be fine with manually testing to make sure any updates behave as expected.

hackmastera:
@escowles I updated the way the jobs are queued. do you want to try it now?

escowles:
@hackmastera Yep, I'll try it out now

escowles:
I'm still seeing only a portion of the records get updated. I've got my local environment setup to test this — do you want me to see if I can get it to complete them all?

hackmastera:
@escowles sure, thanks!

hackmastera:
@escowles are the resources more or less identical? or do they have different metadata, etc?

escowles:
I'm testing with identical resources (other than titles being "Resource #{n}".  I think I've figured out why it's updating only the first page of resources (and have a fix).

hackmastera:
i will update the test / fix rubocop probably in the morning.

app/controllers/bulk_edit_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+class BulkEditController < ApplicationController
+  include Blacklight::SearchHelper
+
+  def resources_edit
+    (solr_response, _document_list) = search_results(q: params["q"], f: params["f"])
+    @resources_count = solr_response["response"]["numFound"]
+  end
+
+  def resources_update
+    args = {}.tap do |hash|
```
tpendragon: Why not just do
```ruby
args = { mark_complete: params["mark_complete"] == "1 }
```
?

app/controllers/bulk_edit_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      hash[:mark_complete] = (params["mark_complete"] == "1")
+    end
+    batches = prepare_batches
```
tpendragon: Can you just rename `prepare_batches` to `batches` and memoize?

app/jobs/bulk_update_job.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+class BulkUpdateJob < ApplicationJob
+  def perform(ids:, args:)
+    return unless args[:mark_complete]
```
tpendragon: Why would this ever get queued without mark_complete? I mean, I guess it would right now, but wouldn't it be more efficient to just not queue a thing if it didn't need updated?

spec/features/bulk_edit_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Bulk edit", js: true do
```
tpendragon: Why do we need JS?

hackmastera: because we test/accept the alert box

app/jobs/bulk_update_job.rb
hackmastera: I think this was a sort of breadcrumb for the fact that we will add new options to the form as we go along. but it doesn't really make sense the way it is; I'll look at moving or removing it.

app/controllers/bulk_edit_controller.rb
hackmastera: To set the pattern for adding more arguments. lmk if you prefer to change it for now.

app/controllers/bulk_edit_controller.rb
hackmastera: 👍

app/controllers/bulk_edit_controller.rb
tpendragon: If you like it the way it is I won't argue.

hackmastera:
@tpendragon I answered some of these questions and made some changes. memoizing batches increased its ABC so I had to extract another method. take another look? let me know what you think.

app/controllers/bulk_edit_controller.rb
hackmastera: Okay because I just added Another One! hahaha *tap happy*

app/jobs/bulk_update_job.rb
```diff
@@ -0,0 +1,37 @@
...
+class BulkUpdateJob < ApplicationJob
+  def perform(ids:, args:)
+    change_set_persister.buffer_into_index do |buffered_change_set_persister|
+      ids.each do |id|
+        resource = query_service.find_by(id: id)
+        change_set = DynamicChangeSet.new(resource).prepopulate!
+        attributes = {}.tap do |attrs|
+          attrs[:state] = "complete" if args[:mark_complete] && !resource.state.include?("complete")
+        end
+        change_set.validate(attributes)
```
tpendragon: Should we check the value of this so it doesn't update it if there's a validation error? Maybe even have it throw an exception so this batch dies?

hackmastera: yes!

hackmastera: @tpendragon I just added this please take a look. Thanks!

app/jobs/bulk_update_job.rb
```diff
@@ -0,0 +1,40 @@
...
+        if change_set.changed?
```
tpendragon: I don't think this is going to do what you want. Should be `if change_set.validate(attributes)` or `if change_set.valid?`.

`change_set.changed?` will just tell you if a property is different, which it really should be if it's being bulk updated.

I'm actually a little concerned the tests below pass in light of this.

tpendragon: OH, I see what's going on. Ignore me.

PR #2552, opened 2019-02-07
Run specs in parallel in CircleCI

tpendragon:
This gets CI time from 10 minutes to 7 minutes.

That being said, I'm unsure it's worth it, since it ties us to Coveralls for sure. What -is- interesting, is using the Solr box in docker caused some indexing errors related to nested documents we had never seen before. The properties were definitely wrong and looked nested, so it's strange that our Solr hasn't been complaining.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,12 @@ jobs:
           POSTGRES_USER: figgy
           POSTGRES_DB: figgy_test
           POSTGRES_PASSWORD: ""
+      - image: solr:7.1-alpine
+        command: bin/solr -cloud -noprompt -f -p 8984
+    parallelism: 4
+    environment:
+      SPEC_OPTS: --format RspecJunitFormatter --out /tmp/test-results/rspec.xml --format progress
+      COVERALLS_PARALLEL: true
     steps:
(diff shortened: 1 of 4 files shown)


app/indexers/parent_issue_indexer.rb
```diff
@@ -10,7 +10,7 @@ def to_solr
     return {} unless resource.is_a?(Coin)
     return {} unless parents.first
 
-    parent_attributes.map { |k, v| ["issue_#{k}_tesim", v] }.to_h
+    parent_attributes.map { |k, v| ["issue_#{k}_tesim", Array.wrap(v.to_s)] }.to_h
```
jrgriffiniii: These changes concern me given what I'm seeing in https://github.com/pulibrary/figgy/pull/2552/files#diff-1d37e48f9ceff6d8030570cd36286a61R100.  I'll merge, but I'm not certain how the affected tests were passing in the past given that the configurations are identical.

tpendragon: Yeah, I commented on this some above. It's weird that this hasn't caused an issue for us before. Solr version diference?

PR #2557, opened 2019-02-08
Add performance recordings to collections

### LINKED ISSUE #2284: Add collection name lookup during music reserves import
to use data generated from #2266

child of #2043
depends on #2265

hackmastera:
and don't make playlists for them unless they're also in another type of
course

closes #2284


app/services/music_import_service.rb
```diff
@@ -162,7 +162,7 @@ def ingest_fake_playlist
           # Find previously imported dependent file sets
           file_set_ids = selection_files.map do |file|
             {
-              buffered_change_set_persister.query_service.custom_queries.find_by_property(property: :local_identifier, value: file.entry_id.to_s).first&.id => file.id
+              query_service.custom_queries.find_by_property(property: :local_identifier, value: file.entry_id.to_s).first&.id => file.id
```
tpendragon: I'm a little surprised this works. I guess Sequel holds the transaction open for a given thread?

hackmastera: How would you expect it to break?

tpendragon: It's the query service for the non-buffered adapter, which I would have thought would be outside the transaction, but I guess that's not how it works.

hackmastera: Okay! Yeah I guess it didn't occur to me to think that it might somehow happen outside the transaction.

PR #2560, opened 2019-02-08
Don't index nested objects

tpendragon:
This will be a problem in future Solr versions, and wasn't intended in
the first place.


tpendragon:
> now that this is by itself I wonder if there's a test we should update or add to?

Yup, could you add a blocking review?

tpendragon:
@hackmastera Tested!

PR #2562, opened 2019-02-08
Remove Coveralls

tpendragon:
CircleCI has access to the simplecov HTML report, and will now error if
it drops below 100%.


hackmastera:
hilariously, 3 days later, this PR is waiting on status from coveralls

PR #2571, opened 2019-02-13
WIP: Convert finding aids EAD XML to MARC records

### LINKED ISSUE #2019: EAD to MARC
Develop a repeatable process for creating MARC records from PULFA EADs (note: some collections already have MARC records, e.g., https://bibdata.princeton.edu/bibliographic/3877093).  The primary objective is to make finding aids more easily discoverable in the catalog, without having to maintain ~2400 MARC records that duplicate what's in the (actively-maintained) EADs.

There is a partial implementation of extracting relevant metadata from PULFA in `PulMetadataServices::PulfaRecord` which might be a good starting place.  But it's just as likely this should be implemented in the PULFA indexer, or in `marc_liberation`, or somewhere else.

Some notes on how the existing MARC record referenced above lines up with the EAD:
https://docs.google.com/spreadsheets/d/1UMyylbaDC-hECIxh5nCNbx-B3cdT5_hR7DecIkHM0aU/edit

escowles:
* Allows discovery of finding aids in the catalog, by syncing changes from finding aids to MARC records
* Based on `ead-to-marc.xsl` from Don Thornbury
* refactoring SvnParser for access

Fixes #2019

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -115,6 +115,7 @@ gem "string_rtl"
 gem "tiny_tds"
 gem "valkyrie-sequel", github: "samvera-labs/valkyrie-sequel", branch: "valkyrie_1.5"
 gem "whenever", "~> 0.10"
+gem "zipline"
 
 gem "blacklight_iiif_search", github: "boston-library/blacklight_iiif_search"
 gem "graphiql-rails", group: :development
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -259,6 +259,7 @@ GEM
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     crass (1.0.4)
+    curb (0.9.8)
     damerau-levenshtein (1.3.1)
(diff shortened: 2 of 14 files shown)


tpendragon:
@escowles Can you give a brief overview on what this does? I really don't understand how this closes #2019. It looks like it just adds an XSLT stylesheet that nothing uses and then refactors the SVN service (which was for auto-updating Figgy metadata when EAD was updated right?), which is something we turned off.

escowles:
@tpendragon You are absolutely right — this doesn't hook those two pieces together yet, so I've marked the PR WIP.

The eventual goal is to use the SVN service to identify EADs that have changed recently, and then use the XSL stylesheet to convert them to MARC XML, and let the user download them.  But the part where we actually do the XSLT and serve the results hasn't been done yet.

escowles:
Closing this PR as the work was completed another way.

PR #2578, opened 2019-02-14
Reorganize numismatic issue and coin forms

### LINKED ISSUE #2455: Coin Form Reorganization
Feedback during all-hands:
> Die axis, size, weight should be together
> In general we can have some groupings: physical things, location things.

Child of #2025

eliotjordan:
Reorganize numismatic issue and coin forms according to feedback in this google doc:

https://docs.google.com/document/d/1OfD_kEpF980U_Ip5SRYZhRkuJZROHy00cXmRUnK61Mc

- Reorders edit and show page attributes.
- Removes `place`, `find`, and `object_type` from Coin resources.
- Removes `description` and `geographic_origin` from Numismatic Issue resources.
- Adds `numismatic_collection` to Coins.

Closes #2455


eliotjordan:
Should we remove the `find` model from Figgy? Looking at the existing database, it appears that each coin has a unique set of find-related fields that aren't shared between coins. The find table itself only has four entries.

escowles:
I would be fine with removing the Find model, or making it an embedded resource for grouping the find-related attributes if we want to do that.

PR #2591, opened 2019-02-15
Upgrade to Valkyrie 2.0.0

tpendragon:
Confirmed this is working on Staging.


scripts/benchmark_conversion.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
```
escowles: Is this included by accident? It seems like a nice thing to have checked-in, but only tangentially related to the rest of the PR..

tpendragon: It was an accident, but then purposefully left in. Seemed useful to have around.

escowles: 👍 seems reasonable to me — just wanted to make sure you wanted it in.

PR #2599, opened 2019-02-18
Ensures that when PULFA ARKs are encountered they are not updated

### LINKED ISSUE #2595: Refuse to update PULFA ARKs, but don't error
We currently raise an exception when refusing to update PULFA ARKs:

https://github.com/pulibrary/figgy/blob/master/app/services/identifier_service.rb#L28

But we should notify the user that the ARK won't be updated, but still allow saving and other actions to work.

jrgriffiniii:
Resolves #2595

### DIFF
--- a/app/controllers/concerns/resource_controller.rb
+++ b/app/controllers/concerns/resource_controller.rb
@@ -59,9 +59,18 @@ module ResourceController
   end
 
   def update
-    @change_set = new_change_set.prepopulate!
+    @change_set = change_set_class.new(find_resource(params[:id])).prepopulate!
     authorize! :update, @change_set.resource
-    persist_updates(@change_set)
+    if @change_set.validate(resource_params)
+      @change_set.sync
+      obj = nil
+      change_set_persister.buffer_into_index do |persist|
+        obj = persist.save(change_set: @change_set)
+      end
+      after_update_success(obj, @change_set)
+    else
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/resource_controller.rb
```diff
@@ -59,21 +59,9 @@ def edit
   end
 
   def update
-    @change_set = change_set_class.new(find_resource(params[:id])).prepopulate!
+    @change_set = new_change_set.prepopulate!
     authorize! :update, @change_set.resource
-    if @change_set.validate(resource_params)
```
tpendragon: I think these extractions make this method harder to parse. What are other's thoughts?

escowles: I find this diff very hard to read, but looking at the end state, the `update` method looks very clean and simple to me.

I'm not sure extracting the `change_set` and `new_change_set` methods is worth it (in fact, I couldn't find any place that's using the new `change_set` method which line 62 duplicates).

hackmastera: Yeah I guess it feels more straightforward to have the logic in the update action if it fits. Did rubocop complain?

jrgriffiniii: Not about this restructuring, but I can return to the original form of the `#update` action.

PR #2606, opened 2019-02-19
Adding time-based ARK reports for PULFA resources

### LINKED ISSUE #2016: New Report: Figgy manifests and Component IDs for recently-ingested items
Allow users to select a date, and provide a CSV report mapping PULFA component IDs to Figgy manifest URLs for items ingested since that date.

escowles:
Fixes #2016

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -34,7 +34,7 @@ class ReportsController < ApplicationController
   def pulfa_ark_report
     authorize! :show, Report
     if params[:since_date]
-      @resources = ark_report_resources(params[:since_date]) || []
+      @resources = query_service.custom_queries.updated_archival_resources(since_date: params[:since_date])
     end
 
     respond_to do |format|
@@ -47,26 +47,6 @@ class ReportsController < ApplicationController
   end
 
   private
-
-    def ark_report_resources(since_date)
-      updated = []
-      query_service.custom_queries.memory_efficient_all(except_models: excluded).each_slice(100) do |records|
-        updated << updated_archival_resources(records, since_date)
-      end
-      updated.flatten!
-    end
-
-    def excluded
-      [Collection, EphemeraBox, EphemeraFolder, EphemeraProject, EphemeraTerm, EphemeraVocabulary, FileSet]
-    end
-
-    def updated_archival_resources(records, since_date)
-      records.select do |r|
-        metadata_id = Array.wrap(r.source_metadata_identifier).first
-        updated = r.updated_at
-        metadata_id.present? && !PulMetadataServices::Client.bibdata?(metadata_id) && updated > since_date
-      end
-    end
 
     def resource_hash(resource)
       {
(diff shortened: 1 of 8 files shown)


app/controllers/reports_controller.rb
```diff
@@ -31,8 +31,61 @@ def identifiers_to_reconcile
...
+    end
+
+    def excluded
+      [Collection, EphemeraBox, EphemeraFolder, EphemeraProject, EphemeraTerm, EphemeraVocabulary, FileSet]
+    end
+
+    def updated_archival_resources(records, since_date)
+      records.select do |r|
+        metadata_id = Array.wrap(r.source_metadata_identifier).first
+        updated = Array.wrap(r.source_metadata_identifier).first
```
tpendragon: This isn't right, right? Maybe we need some more tests?

app/controllers/reports_controller.rb
```diff
@@ -31,8 +31,61 @@ def identifiers_to_reconcile
...
+      format.csv do
+        send_data hashes_to_csv(["id", "component_id", "ark", "url"], @resources.map { |r| resource_hash(r) }),
+                  filename: "pulfa-ark-report-#{params[:since_date]}-to-#{Time.zone.today}.csv"
+      end
+    end
+  end
+
   private
 
+    def ark_report_resources(since_date)
```
tpendragon: Maybe we should extract this to a query, so we have somewhere to make this more efficient eventually?

app/controllers/reports_controller.rb
escowles: Oops — I'll add a test that checks the updated date because it's obviously not being tested now.

app/controllers/reports_controller.rb
escowles: 👍 Yes, I've pulled that out to a custom query and switched to doing it with SQL, which will surely perform much, much better: https://github.com/pulibrary/figgy/pull/2606/files#diff-4a0243fdb3248f71bcb779a46ce59511

app/queries/updated_archival_resources.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def updated_archival_resources(since_date:)
+    run_query(query(since_date))
+  end
+
+  def query(since_date)
+    <<-SQL
+      select * FROM orm_resources
+      WHERE metadata->>'archival_collection_code' is not null AND updated_at > '#{since_date}'
```
tpendragon: Can you change this to be `WHERE metadata @> '{"archival_collection_code": []}' AND updated_at > ?`, then change the above to be `run_query(query, since_date.to_s)`?

`metadata->>'archival_collection_code'` doesn't use an index (unless you explicitly create one for that field), and we should use parameterized queries when we can just to be safe.

tpendragon: Forgot to request changes.

tpendragon: For reference, just ran these two queries on prod and it's a difference of 35s vs 7ms.

escowles: 👍 Just updated the query

Issue #2604, opened 2019-02-19
delete any subject field entry with value ab8236da-b222-4d4f-aedb-555456d10401


hackmastera:
this term has been deleted and reindexing errors if it comes up.

escowles:
I think this is a special case of one of the oldest Figgy tickets: https://github.com/pulibrary/figgy/issues/241

hackmastera:
@escowles this one is to clean up after the one specific deletion. I think https://github.com/pulibrary/figgy/issues/2603 is the special case.

hackmastera:
oh i see what you mean. yeah this is a manifestation of that ticket.

hackmastera:
3 ephemera folders will need to be updated manually by a metadata expert. 2 of them had only this one subject, and a subject is required by the form. the other lacks a different required field, the folder number:

- [ ] https://figgy.princeton.edu/catalog/7fa1ce90-032b-4209-a3c7-4d50330f685e
- [ ] https://figgy.princeton.edu/catalog/3eecb0d5-6681-44a5-bf5c-12d2da1826e5
- [ ] https://figgy.princeton.edu/catalog/5e988ed6-7108-4870-a77d-a83f86b43427

PR #2616, opened 2019-02-21
Add a job to delete a subject from all ephemera folders

hackmastera:
I will probably just run this job from the console for now, but I wanted some tested code
refs #2604

### DIFF
--- a/app/jobs/clean_deleted_subject_job.rb
+++ b/app/jobs/clean_deleted_subject_job.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+class CleanDeletedSubjectJob < ApplicationJob
+  def perform(subject_id, logger: Logger.new(STDOUT))
+    adapter = Valkyrie::MetadataAdapter.find(:indexing_persister)
+    qs = adapter.query_service
+    folders = qs.find_inverse_references_by(property: :subject, id: subject_id)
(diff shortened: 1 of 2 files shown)


escowles:
This PR handles the cases where there is a leftover ID, so it's complementary to https://github.com/pulibrary/figgy/pull/2609, right?

hackmastera:
@escowles yes. there are 40-50 instances for the id we ran into this week. If there are others they would probably prohibit a full reindex.

app/jobs/clean_deleted_subject_job.rb
```diff
@@ -0,0 +1,15 @@
...
+  def perform(subject_id)
+    adapter = Valkyrie::MetadataAdapter.find(:indexing_persister)
+    qs = adapter.query_service
+    folders = qs.find_inverse_references_by(property: :subject, id: subject_id)
+    csp = ChangeSetPersister.new(metadata_adapter: adapter, storage_adapter: Valkyrie.config.storage_adapter)
+    folders.each do |folder|
+      cs = DynamicChangeSet.new(folder).prepopulate!
+      subjects = cs.subject.reject { |id| id.to_s == subject_id }
+      cs.validate(subject: subjects)
+      csp.save(change_set: cs) if cs.valid?
```
escowles: Do you want to log an error if it's not valid?

hackmastera: sure good idea

PR #2621, opened 2019-02-21
Add orangelight document builder for coins

eliotjordan:
- Adds generic Orangelight document service and a builder for Coin resources.
- Fixes issue with nil values in numismatic monogram input.
- Increases flexibility of the year validator to accommodate BCE and years with less than four digits.

### DIFF
--- a/spec/resources/coins/coins_controller_spec.rb
+++ b/spec/resources/coins/coins_controller_spec.rb
@@ -123,4 +123,15 @@ RSpec.describe CoinsController, type: :controller do
       end
     end
   end
+  describe "GET /coins/:id/orangelight" do
+    let(:user) { FactoryBot.create(:admin) }
+
+    it "renders an orangelight document" do
+      coin = FactoryBot.create_for_repository(:coin)
+      FactoryBot.create_for_repository(:numismatic_issue, member_ids: [coin.id])
+      get :orangelight, params: { id: coin.id, format: :json }
+      doc = JSON.parse(response.body)
+      expect(doc["id"]).to eq "coin-#{coin.coin_number}"
+    end
+  end
 end
--- a/spec/services/orangelight_document_spec.rb
+++ b/spec/services/orangelight_document_spec.rb
@@ -1,0 +1,140 @@
+# frozen_string_literal: true
+require "rails_helper"
+include ActionDispatch::TestProcess
+
+describe OrangelightDocument do
+  describe "#to_json" do
+    context "when a resource does not have an associated builder class" do
+      subject(:builder) { described_class.new(scanned_resource) }
+      let(:scanned_resource) { ScannedResource.new }
+
+      it "raises a NotImplementedError" do
+        expect { builder.to_json }.to raise_error(NotImplementedError)
+      end
+    end
+
+    context "with an issue and a child coin" do
+      subject(:builder) { described_class.new(coin) }
+      let(:file) { fixture_file_upload("files/example.tif", "image/tiff") }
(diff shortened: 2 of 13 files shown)


spec/resources/coins/coins_controller_spec.rb
```diff
@@ -123,4 +123,15 @@
...
   end
+  describe "GET /coins/:id/orangelight" do
+    let(:user) { FactoryBot.create(:admin) }
+
+    it "renders an orangelight document" do
+      coin = FactoryBot.create_for_repository(:coin)
+      FactoryBot.create_for_repository(:numismatic_issue, member_ids: [coin.id])
+      get :orangelight, params: { id: coin.id, format: :json }
+      doc = JSON.parse(response.body)
+      expect(doc["id"]).to eq "coin-1"
```
escowles: Any reason not to test for `"coin-#{coin.coin_number}"` ?

spec/services/orangelight_document_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+
+      before do
+        stub_ezid(shoulder: "99999/fk4", blade: "123456")
+        coin
+        issue
+      end
+
+      it "returns a IIIF manifest for a resource with a file" do
+        output = MultiJson.load(builder.to_json, symbolize_keys: true)
+        expect(output[:id]).to eq "coin-1"
```
escowles: Same question as above: any reason not to test for `"coin-#{coin.coin_number}"` ?

app/services/orangelight_document/coin_builder.rb
```diff
@@ -0,0 +1,83 @@
...
+      @parent = decorator.decorated_parent
+    end
+
+    def build
+      clean_document(document_hash)
+    end
+
+    private
+
+      def clean_document(hash)
```
escowles: I was thinking about whether this class should go in `app/resources/coins`, and one of the benefits of having them all in `app/services/orangelight_document` is that it would make it easier to build common functionality in one place.  Do you want to start doing that, e.g., by putting `build` and `clean_document` in `OrangelightDocument` and making this a subclass?

spec/resources/coins/coins_controller_spec.rb
eliotjordan: Nope. That's a better test. Updated.

eliotjordan: Updated.

app/services/orangelight_document/coin_builder.rb
eliotjordan: I don't think CoinBuilder is a subclass of OrangelightDocument. However, I could see adding a base builder class that has utility methods like `clean_document`. I'm on the fence about moving it into `app/resources/coins/`. Since we would be removing it from it's namespace, we could rename it to `CoinOrangelightBuilder` or `OrangelightCoinBuilder` to better indicate its function.

eliotjordan: @escowles On further reflection, putting builders in the `app/resources/*` would be consistent with our current pattern. What do you think?

escowles: I do like the idea of moving them into `app/resources/*` — and we can probably wait to create the base class until we have another type with an Orangelight builder.

eliotjordan: 👍 I'll re-arrange the classes and push up an update.

eliotjordan: Updated.

eliotjordan:
@escowles Moved the coin builder class to `app/resources/`. Ready for another review.

PR #2658, opened 2019-02-28
 Add graphql queries for coins and orangelight ids

### LINKED ISSUE #2655: Add GraphQL query to lookup by Orangelight slug
This is to support numismatics in Orangelight. A more general query that runs a regex on the slug pattern to find the correct property to query on.

eliotjordan:
Adds new queries for orangelight ids rather than just bib_ids.

Closes #2655

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -59,7 +59,7 @@ class Types::QueryType < Types::BaseObject
   end
 
   def resources_by_orangelight_id(id:)
-    if id.start_with?("coin-")
+    if coin_id?(id)
       resources_by_coin_number(coin_number: id.gsub("coin-", ""))
     else
       resources_by_bibid(bib_id: id)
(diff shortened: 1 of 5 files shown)


app/queries/find_many_by_non_string_property.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class FindManyByNonStringProperty
```
escowles: Do we need both this class and `FindManyByStringProperty`? We worked around a similar difference to create a single `FindByProperty` query: https://github.com/pulibrary/figgy/pull/2472/files#diff-795a87e9dbaac53a47575a81d2aa06f2

eliotjordan: I hadn't noticed that. I'll see if I can reconfigure the query classes.

eliotjordan: @escowles Created a new FindManyByProperty query class to replace those two.

PR #2659, opened 2019-02-28
Track file sets being created on ChangeSet

### LINKED ISSUE #2654: CharacterizationJob is sometimes just not getting queued?
The result is a file which has been ingested but has mime_type "text/plain" and no derivatives.

tpendragon:
The change_set_persister is re-used for multiple interactions with a
variety of ChangeSets. As such, tracking it there was leading to the
created_file_sets getting wiped, resulting in no characterization being
run.

Closes #2654

### DIFF
--- a/app/change_set_persisters/change_set_persister/characterize.rb
+++ b/app/change_set_persisters/change_set_persister/characterize.rb
@@ -1,12 +1,11 @@
 # frozen_string_literal: true
 class ChangeSetPersister
   class Characterize
-    attr_reader :change_set_persister, :change_set, :post_save_resource, :created_file_sets
+    attr_reader :change_set_persister, :change_set, :post_save_resource
     delegate :characterize?, to: :change_set_persister
-    def initialize(change_set_persister:, change_set:, created_file_sets:, post_save_resource: nil)
+    def initialize(change_set_persister:, change_set:, post_save_resource: nil)
       @change_set = change_set
       @change_set_persister = change_set_persister
-      @created_file_sets = created_file_sets
       @post_save_resource = post_save_resource
     end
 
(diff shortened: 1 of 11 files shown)


app/change_set_persisters/change_set_persister/characterize.rb
```diff
@@ -17,5 +16,9 @@ def run
         ::CharacterizationJob.set(queue: change_set_persister.queue).perform_later(file_set.id.to_s)
       end
     end
+
+    def created_file_sets
+      change_set.respond_to?(:created_file_sets) ? change_set.created_file_sets : []
```
hackmastera: would `change_set.try(:created_file_sets) || []` also work? I know you hate ternaries...

app/models/controlled_vocabulary.rb
```diff
@@ -30,7 +30,7 @@ def initialize(key: nil)
 
   # Method returning all terms within the vocabulary
   # This should be overridden
-  def all
+  def all(_scope = nil)
```
hackmastera: what is this for?

spec/change_set_persisters/change_set_persister_spec.rb
```diff
@@ -431,6 +431,30 @@
       end
     end
 
+    it "runs characterization for subsequent files", run_real_derivatives: true do
```
hackmastera: Can you expand on "subsequent files" here? Maybe "runs characterization for all files when added in sequence with the same persister"? Or something that points just a little bit more to the source of the scenario?

app/change_set_persisters/change_set_persister/characterize.rb
tpendragon: That's a hidden ternary! It's ternaries all the way down!

But yeah, sure.

app/models/controlled_vocabulary.rb
tpendragon: It's...unrelated. It popped up because of a bug in a previous implementation. It's definitely broken and doesn't fulfill the expected interface, but we have nothing testing it. I'll change it back and make an issue.

PR #2661, opened 2019-02-28
Show issue in breadcrumb when parent_id is not in the path

### LINKED ISSUE #2342: Coin: show Issue breadcrumbs even if not using parent path
When I go to a coin at the `/catalog/[id]`, it should show the Numismatic Issue parent in the breadcrumbs.

Child of #2025

christinach:
closes #2342

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -132,7 +132,15 @@ class CatalogController < ApplicationController
       query = "member_ids_ssim:id-#{params['id']}"
       _, result = search_results(q: query, rows: 1)
       @parent_document = result.first if result.first
+    elsif params[:parent_id].nil? && @document.decorated_resource.is_a?(CoinDecorator)
+      set_coin_parent
     end
+  end
+
+  def set_coin_parent
+    return unless CoinDecorator.new(resource).decorated_parent
+    params[:parent_id] = @document.decorated_resource.decorated_parent["id"].id
+    _, @parent_document = fetch(params[:parent_id])
   end
 
   def lookup_manifest
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -132,14 +132,13 @@ class CatalogController < ApplicationController
       query = "member_ids_ssim:id-#{params['id']}"
       _, result = search_results(q: query, rows: 1)
       @parent_document = result.first if result.first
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -132,9 +132,17 @@ def set_parent_document
       query = "member_ids_ssim:id-#{params['id']}"
       _, result = search_results(q: query, rows: 1)
       @parent_document = result.first if result.first
+    elsif params[:parent_id].nil? && @document.decorated_resource.is_a?(CoinDecorator)
```
tpendragon: This seems like a good feature to do for everything. Maybe change it to `@document.decorated_resource.try(:decorated_parent)`

app/controllers/catalog_controller.rb
```diff
@@ -132,9 +132,17 @@ def set_parent_document
...
+      set_coin_parent
     end
   end
 
+  def set_coin_parent
+    return unless CoinDecorator.new(resource).decorated_parent
```
tpendragon: You can get rid of this if you check for it above.

spec/resources/coins/coin_feature_spec.rb
```diff
@@ -1,16 +1,21 @@
 # frozen_string_literal: true
 require "rails_helper"
 
-RSpec.feature "Coins" do
+RSpec.feature "Coins", js: true do
```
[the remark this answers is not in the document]
christinach: you're right. it doesn't. I think I tested in the beginning a style with selenium and added it. removed

christinach:
@tpendragon thanks

christinach:
I'm checking a local error with a 'boxless ephemera project'.

christinach:
ok, it looks like I had to rebase.

app/controllers/catalog_controller.rb
```diff
@@ -132,9 +132,16 @@ def set_parent_document
       query = "member_ids_ssim:id-#{params['id']}"
       _, result = search_results(q: query, rows: 1)
       @parent_document = result.first if result.first
+    elsif params[:parent_id].nil? && @document.decorated_resource.try(:decorated_parent)
+      set_coin_parent
     end
   end
 
+  def set_coin_parent
+    params[:parent_id] = @document.decorated_resource.decorated_parent["id"].id
```
tpendragon: Oh, instead of `.id` I'd do `.to_s`.

christinach: thanks @tpendragon

PR #2665, opened 2019-03-01
Include artist table in coin

### LINKED ISSUE #2582: Numismatics issue artist table needs to be included in coin models
The table includes the fields person, signature, role, side and describes info about the issue artist

christinach:
closes #2582

### DIFF
--- a/spec/resources/coins/coin_decorator_spec.rb
+++ b/spec/resources/coins/coin_decorator_spec.rb
@@ -3,10 +3,11 @@ require "rails_helper"
 
 RSpec.describe CoinDecorator do
   subject(:decorator) { described_class.new(coin) }
-  let(:coin) { FactoryBot.create_for_repository(:coin, numismatic_citation_ids: [citation.id], accession_number: accession.accession_number) }
+  let(:coin) { FactoryBot.create_for_repository(:coin, numismatic_citation_ids: [citation.id], accession_number: accession.accession_number, numismatic_artist_ids: [artist.id]) }
   let(:citation) { FactoryBot.create_for_repository(:numismatic_citation, numismatic_reference_id: [reference.id]) }
   let(:accession) { FactoryBot.create_for_repository(:numismatic_accession, accession_number: 234) }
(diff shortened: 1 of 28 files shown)


escowles:
I think the artist property should probably be an embedded object like references/citations — because one coin can have multiple artists (e.g., one for the obverse and one for the reverse). So the side, role, and signature properties are all properties of the relationship between an issue/coin and the artist.

christinach:
This is a [WIP] PR. I can't update the title. github bug.

spec/views/catalog/_members_coin.html.erb_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it "shows Citation table" do
+      expect(rendered).to have_selector "h2", text: "Citations"
+      expect(rendered).to have_link "View"
+      expect(rendered).to have_link "Edit"
+    end
+
+    it "shows Artist table" do
+      expect(rendered).to have_selector "h2", text: "Artists"
+      expect(rendered).to have_link "View"
+      expect(rendered).to have_link "Edit"
```
escowles: Can you update these tests to have an href or CSS selector or some other limit so the same View/Edit links wouldn't match for these and the Citation tests?

spec/resources/coins/coin_decorator_spec.rb
```diff
@@ -20,6 +21,12 @@
     end
   end
 
+  describe "#artists" do
+    it "renders the linked artists" do
+      expect(decorator.artists).to eq(["artist person artist role"])
```
escowles: Do you think it would be better to separate the role with a comma, or put it in parentheses?

spec/views/catalog/_members_coin.html.erb_spec.rb
christinach: yes I can, thank you

escowles:
@christinach This looks good to me — just one or two little things.

spec/resources/coins/coin_decorator_spec.rb
christinach: I'll add a comma

Issue #2657, opened 2019-02-28
intermittent order manager save problems


hackmastera:
Report from Tom: The past two days, if I go into Order Manager for an item and order-paginate or otherwise rename images, I find that the changes do not hold for several (seemingly random) images.  It usually takes a few iterations for me going back in and manually renaming the images for the changes to hold.  I am clicking "Apply Changes" after each individual name change and waiting until the change is confirmed before moving on

tpendragon:
We're having trouble replicating this. We need to schedule a call with Tom to see this in action.

tpendragon:
The problem here is background jobs overriding FileSets. Tom is setting an OCR language, which spawns a bunch of RunOCR jobs. When he edits the files in the OrderManager, the OCR jobs are running with out-of-date data, and when they save they override the changes he's made. So it goes RunOCRJob Starts -> Order manager Saves -> Run OCR Job Saves -> Order manage changes are gone.

To stop data corruption (maybe with a not-great user experience, but better than nothing) we'll have to implement optimistic locking (see #1874).

A short term band-aid would be to make RunOCRJob have as little time as possible between pulling the object and saving it, and making it so updates don't spawn CheckFixityJobs (as per #2431).

escowles:
Optimistic locking should work well for this: when the OCR job tries to save, it'll throw an error if the fileset has been updated. Then the OCR job can reload the fileset, add the OCR text, and retry saving.

tpendragon:
The "band-aid" solution is turning out a little more involved than we expected. It's turning out that DerivativeServices should be pulling the resource at the beginning of its processing if it's going to be updating it, so we're changing the method signature to accept an ID instead of a ChangeSet.

PR #2676, opened 2019-03-06
Don't Check Fixity After Saving/Reload Objects in DerivativeService

tpendragon:
These two commits combine to work as a band-aid solution to #2657 until we get optimistic locking enabled.


spec/derivative_services/audio_derivative_service_spec.rb
```diff
@@ -4,7 +4,7 @@
 include ActionDispatch::TestProcess
 
 RSpec.describe AudioDerivativeService do
-  it_behaves_like "a Valkyrie::Derivatives::DerivativeService"
+  # it_behaves_like "a Valkyrie::Derivatives::DerivativeService"
```
escowles: Do we want to change the shared spec? Alternatively, we could keep the signature the same, but use the change_set's ID to reload the object instead of passing the ID...

tpendragon: I think we'll want to change the shared spec at some point.

escowles: 👍 I created a ticket for tracking whether we want to change the API: https://github.com/samvera-labs/valkyrie-derivatives/issues/3

Meanwhile, do you want to mark the shared spec pending?

tpendragon: :+1:

Issue #2680, opened 2019-03-07
Bulk update local identifiers for cicognara items


hackmastera:
It looks like any "local identifier" that starts with `cico:` needs to be modified to `dcl:`.

hackmastera:
related to pulibrary/cicognara-rails#330

escowles:
Completed 03/12/12

PR #2683, opened 2019-03-08
Add rake task to update local ids from `cico:` to `dcl:`

hackmastera:
refs #2680


escowles:
@tpendragon I fixed a rubocop error, so it dismissed your review — can you re-review?

PR #2696, opened 2019-03-11
Remove order manager button from monogram show page

### LINKED ISSUE #2672: Remove Order manager button from monograms

christinach:
closes #2672

### DIFF
--- a/app/decorators/valkyrie/resource_decorator.rb
+++ b/app/decorators/valkyrie/resource_decorator.rb
@@ -79,6 +79,10 @@ class Valkyrie::ResourceDecorator < ApplicationDecorator
     manageable_files?
   end
 
+  def manageable_order?
+    true
+  end
+
   def manageable_structure?
     false
   end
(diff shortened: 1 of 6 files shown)


app/views/catalog/_display_order_manager_button.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if can?(:order_manager, resource) %>
+  <% if decorated_resource.orderable_files? && !decorated_resource.is_a?(NumismaticMonogramDecorator) %>
```
escowles: I think it would be nice to create a new method on ChangeSet called something like `manageable_order?` to control whether this button is displayed or not.

christinach: ok will do, thanks

christinach:
@escowles I added a method in the `class Valkyrie::ResourceDecorator < ApplicationDecorator`.

app/decorators/valkyrie/resource_decorator.rb
```diff
@@ -79,6 +79,10 @@ def orderable_files?
     manageable_files?
   end
 
+  def manageable_order?
+    true unless NumismaticMonogramDecorator
```
escowles: Can you override this in `NumismaticMonogramDecorator` instead of having a conditonal here?

christinach: yes I figured out I was wrong

PR #2700, opened 2019-03-11
Foliation not happen when in paginate mode

### LINKED ISSUE #2510: Bulk Labeler shouldn't use both pagination and foliation settings
When using the bulk labeler in pagination mode, the default foliation prefixes (r. and v.) are also being applied.  It should only use the pagination prefix in pagination mode, and foliation prefixes in foliation mode.

tpendragon:
Closes #2510

### DIFF
--- a/app/javascript/components/OrderManagerFilesetsForm.vue
+++ b/app/javascript/components/OrderManagerFilesetsForm.vue
@@ -9,21 +9,21 @@
     >
       <input-text
         id="unitLabel"
-        v-model="labelerOpts.unitLabel"
+        v-model="unitLabel"
         label="Label"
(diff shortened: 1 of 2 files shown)


jrgriffiniii:
I could not access the pagination/foliation menu on this branch in staging:
screenshot2

this was rendered in production:

screenshot1

jrgriffiniii:
The pagination/foliation menu now renders for me, but when bulk labeling by pagination, then by foliation, and attempting to relabel once again by pagination, the results seem to retain the foliation characters:

image

I don't think that this should necessarily block the PR, but it might raise some questions from users who happen upon this by mistake.

PR #2704, opened 2019-03-12
Update form_title based on the change_set value

### LINKED ISSUE #2518: New recording form should have title "New recording"
Currently says "New scanned resource"

christinach:
closes #2518

### DIFF
--- a/app/helpers/metadata_form_helper.rb
+++ b/app/helpers/metadata_form_helper.rb
@@ -2,13 +2,13 @@
 module MetadataFormHelper
   def form_title(params)
     if form_custom_title(params)
-      "#{params['action'].capitalize} #{params['change_set'].humanize.singularize.downcase}"
+      "#{params['action'].capitalize} #{params['change_set'].humanize.singularize.downcase} resource"
     else
       "#{params['action'].capitalize} #{params['controller'].humanize.singularize.downcase}"
     end
   end
 
   def form_custom_title(params)
-    params["change_set"].present? && params["change_set"] == "recording"
+    params["change_set"].present?
   end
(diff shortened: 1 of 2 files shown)


app/helpers/metadata_form_helper.rb
```diff
@@ -1,6 +1,14 @@
...
-    "#{params['action'].capitalize} #{params['controller'].humanize.singularize.downcase}"
+    if form_custom_title(params)
+      "#{params['action'].capitalize} #{params['change_set'].humanize.singularize.downcase}"
+    else
+      "#{params['action'].capitalize} #{params['controller'].humanize.singularize.downcase}"
+    end
+  end
+
+  def form_custom_title(params)
+    params["change_set"].present? && params["change_set"] == "recording"
```
escowles: Why limit this just to recordings, if we expect to expand this pattern to cover other custom change sets in the future? Are there any custom change sets where we wouldn't want to use the change set name for the form labels?

christinach: My first commit which I squashed in the end was a general pattern without limiting to the recording.
I checked the :change_set simple in the scanned_resources; it will not fire up in the title when we call new or edit scanned resource. I think currently there is no limitation but I wanted to make sure with the rest before I introduce the general.

christinach: I will push again the general pattern commit

christinach: I think the new simple was the main reason I limited the commit to the recording maybe until we discuss how to handle the name. I could add the word resource in the general pattern.

PR #2708, opened 2019-03-12
Don't bulk update takedown states.

### LINKED ISSUE #2635: Bulk Edit should not complete items in takedown state
Users can work around this by faceting on state to select only items of one particular state (e.g., pending).  But we wouldn't want a bulk complete action to accidentally restore access to items marked takedown.

tpendragon:
Closes #2635

### DIFF
--- a/app/jobs/bulk_update_job.rb
+++ b/app/jobs/bulk_update_job.rb
@@ -6,7 +6,7 @@ class BulkUpdateJob < ApplicationJob
         resource = query_service.find_by(id: id)
         change_set = DynamicChangeSet.new(resource).prepopulate!
         attributes = {}.tap do |attrs|
-          attrs[:state] = "complete" if args[:mark_complete] && !resource.state.include?("complete")
+          attrs[:state] = "complete" if args[:mark_complete] && !blacklisted_states.include?(resource.state.first)
         end
         change_set.validate(attributes)
         if change_set.changed?
@@ -18,6 +18,13 @@ class BulkUpdateJob < ApplicationJob
   end
 
   private
+
+    def blacklisted_states
+      [
+        "complete",
+        "takedown"
+      ]
+    end
 
     def query_service
       metadata_adapter.query_service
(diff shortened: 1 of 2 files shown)


app/jobs/bulk_update_job.rb
```diff
@@ -19,6 +19,17 @@ def perform(ids:, args:)
 
   private
 
+    def resource_has_updateable_state?(resource)
+      (resource.state & blacklisted_states) == []
```
escowles: Since we know there can only be one value in `resource.state`, why  not do `!blacklisted_states.include?(resource.state.first)`?

spec/jobs/bulk_update_job_spec.rb
```diff
@@ -25,6 +25,19 @@
...
+    context "one of the resources is taken down" do
+      let(:resource2) do
+        Timecop.freeze(Time.now.utc - 1.day) do
+          FactoryBot.create_for_repository(:scanned_resource, state: "takedown")
+        end
+      end
+      it "doesn't persist the one that was marked taken down" do
+        described_class.perform_now(ids: ids, args: args)
+        r2 = query_service.find_by(id: resource2.id)
+        expect(r2.updated_at.to_date).to be < Time.current.to_date
```
escowles: Can you verify that the state is still takedown? When we add support for other fields to the bulk edit, we might want to update the rights statement, or add to a collection, but still leave the state alone.

app/jobs/bulk_update_job.rb
tpendragon: I thought I should deal with the structure I had rather than assume what's in it. I can do that, though - I imagine we should eventually just make state single-valued.

escowles: I appreciate the idea of dealing with the structure as it is — though I prefer to avoid infrequently used operators like this to avoid someone misreading it later.

tpendragon:
@escowles Should be ready for re-review

PR #2726, opened 2019-03-14
Don't pre-populate everywhere

tpendragon:
Reform pre-population is meant as a way to build up filler nested objects for display in an HTML form, not to transform values. This should make it easier to follow Reform documentation.

### DIFF
--- a/spec/derivative_services/external_metadata_derivative_service_spec.rb
+++ b/spec/derivative_services/external_metadata_derivative_service_spec.rb
@@ -57,6 +57,6 @@ RSpec.describe ExternalMetadataDerivativeService do
     parent = query_service.find_by(id: parent_resource.id)
     expect(parent.title).to eq ["China census data by county, 2000-2010"]
     expect(parent.visibility).to eq ["open"]
-    expect(event_generator).to have_received(:record_updated).twice
+    expect(event_generator).to have_received(:record_updated).exactly(3).times
   end
 end
--- a/spec/queries/find_members_with_relationship_spec.rb
+++ b/spec/queries/find_members_with_relationship_spec.rb
@@ -7,16 +7,16 @@ RSpec.describe FindMembersWithRelationship do
 
   describe "#find_members_with_relationship" do
     it "returns members with `loaded_relationship` loaded" do
-      genre = FactoryBot.create_for_repository(:ephemera_term, label: "Genre")
-      genre2 = FactoryBot.create_for_repository(:ephemera_term, label: "Genre2")
-      folder = FactoryBot.create_for_repository(:ephemera_folder, genre: [genre.id, genre2.id])
-      folder2 = FactoryBot.create_for_repository(:ephemera_folder, genre: [genre.id])
+      genre = FactoryBot.create_for_repository(:ephemera_term, label: "Subject1")
+      genre2 = FactoryBot.create_for_repository(:ephemera_term, label: "Subject2")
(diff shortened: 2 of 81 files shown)


spec/queries/find_members_with_relationship_spec.rb
```diff
@@ -7,16 +7,16 @@
 
   describe "#find_members_with_relationship" do
     it "returns members with `loaded_relationship` loaded" do
```
escowles: How are these changes related to the `prepopulate` change?

spec/resources/coins/coin_feature_spec.rb
```diff
@@ -127,7 +127,7 @@
     let(:artist) { FactoryBot.create_for_repository(:numismatic_artist) }
 
     before do
-      visit solr_document_path coin
+      visit solr_document_path id: coin.id.to_s
```
escowles: Why is this change needed?  Wouldn't it route to the same page either way?

spec/services/manifest_builder/license_builder_spec.rb
```diff
@@ -20,21 +20,5 @@
         expect(manifest["license"]).to eq "https://creativecommons.org/licenses/by-nc/4.0/"
       end
     end
```
escowles: How are these changes related to the `prepopulate` change?

spec/derivative_services/external_metadata_derivative_service_spec.rb
```diff
@@ -57,6 +57,6 @@
     parent = query_service.find_by(id: parent_resource.id)
     expect(parent.title).to eq ["China census data by county, 2000-2010"]
     expect(parent.visibility).to eq ["open"]
-    expect(event_generator).to have_received(:record_updated).twice
+    expect(event_generator).to have_received(:record_updated).exactly(3).times
```
escowles: Why did this change?

tpendragon: I actually dug around this, and I have no idea. It seemed like it was firing everything at correct times, I'm not sure how it only fired twice before.

spec/services/manifest_builder/license_builder_spec.rb
tpendragon: I have no idea how this worked before! It seems like it was a bug. If there's a license field, why would it just be empty if you save a value?

spec/resources/coins/coin_feature_spec.rb
tpendragon: Oh, yes, I can get rid of this. The problem was that it was only persisting to postgres before, but the change set was being marked as changed as part of creating the record, so the change_set_persister was being fired which would re-save it and it'd end up in solr.

spec/queries/find_members_with_relationship_spec.rb
tpendragon: Oh! This one was interesting - it wasn't pre-populating before. It's actually impossible to save a change_set with two genres because it's marked as a single field in the ChangeSet. Now that it happens on initialize, I had to use a field which was marked as multiple to get two things to show up.

spec/derivative_services/external_metadata_derivative_service_spec.rb
tpendragon: I just dug through this again. There are 3 events:

1. The change_set is saved for the first time.
2. Saving the change_set for the first time fixes the visibility of the FileSet that's attached.
3. The metadata is extracted, firing the updated for the record again.

PR #2733, opened 2019-03-14
Letter Change Sets

tpendragon:
Depends on #2726 

This enables some new functionality for nested forms, as well as an experimental method of grouping ChangeSet properties by feature. It's pretty similar to multiple inheritance, but all in one place - something to think about?

The form looks like this: 

Screen Shot 2019-03-14 at 4 37 18 PM

### DIFF
--- a/app/resources/scanned_resources/letter_change_set.rb
+++ b/app/resources/scanned_resources/letter_change_set.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+class LetterChangeSet < ChangeSet
+  apply_workflow(DraftCompleteWorkflow)
+  core_resource(change_set: "letter")
+  enable_order_manager
+  enable_pdf_support
+
+  property :sender, multiple: false, required: false, form: NameWithPlaceChangeSet, populator: :populate_name_with_place
+  property :recipient, multiple: false, required: false, form: NameWithPlaceChangeSet, populator: :populate_name_with_place
+
+  def primary_terms
+    feature_terms.dup.insert(
+      2,
+      [
+        :sender,
+        :recipient,
+        :member_of_collection_ids
+      ]
+    ).flatten
+  end
+
+  def prepopulate!(_args = {})
+    self.sender ||= NameWithPlace.new
+    self.recipient ||= NameWithPlace.new
+    super
+  end
+
+  def populate_name_with_place(fragment:, as:, **)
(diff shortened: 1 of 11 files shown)


app/resources/scanned_resources/letter_change_set.rb
```diff
@@ -0,0 +1,35 @@
...
+  apply_workflow(DraftCompleteWorkflow)
+  core_resource(change_set: "letter")
+  enable_order_manager
+  enable_pdf_support
+
+  property :sender, multiple: false, required: false, form: NameWithPlaceChangeSet, populator: :populate_name_with_place
+  property :recipient, multiple: false, required: false, form: NameWithPlaceChangeSet, populator: :populate_name_with_place
+
+  def primary_terms
+    feature_terms.dup.insert(
```
hackmastera: this is a little awkward, but in general i like the property shortcuts idea.  A bit of documentation to go along with it in the figgy wiki? If so, pointing out this pattern is probably a good idea.

app/resources/scanned_resources/letter_change_set.rb
```diff
@@ -0,0 +1,35 @@
...
+    ).flatten
+  end
+
+  def prepopulate!(_args = {})
+    self.sender ||= NameWithPlace.new
+    self.recipient ||= NameWithPlace.new
+    super
+  end
+
+  def populate_name_with_place(fragment:, as:, **)
```
hackmastera: Why does this live here instead of the name with place change set? Is the idea that different change sets may want to prepopulate it differently?

hackmastera: I see fragment in the reform documentation; where is `as` coming from?

tpendragon: This is just how the reform documentation says to populate a nested object - I don't think there's a way to do it on the ChangeSet, especially to ignore population if the params are empty.

`as` is just one of the arguments, I noticed it when I was fiddling. I'm having a hard time finding it documented somewhere, but it's definitely passed in.

app/resources/scanned_resources/letter_change_set.rb
tpendragon: Added to https://github.com/pulibrary/figgy/wiki/ChangeSets-in-Figgy

PR #2734, opened 2019-03-15
Updates the BulkIngestController and BaseResourceController for directories selected with browse-everything

### LINKED ISSUE #1536: Improve Bulk Ingest Efficiency for Large Books
Also want to bulk ingest, e.g. 5 books
Need folder selector; browse_everything only provides file-level selection. Want just the name of the folder.

discussed at digital studio users meeting 7/19/2018

jrgriffiniii:
This required some significant adjustments to the `browse-everything` branch which already integrated support for paginating Google Drive file entries: https://github.com/jrgriffiniii/browse-everything/compare/gdrive-paging-jrgriffiniii-rebased...jrgriffiniii:figgy-issues-1536-jrgriffiniii?expand=1

In order to retain backward compatibility with 1.x releases, this duplicates form fields and parameters passed in the `POST` request which disambiguate selected files from directories.  Additionally, this also provides a `BrowseEverything::Resource` class (which might be better suited for the `browse-everything` branch instead).  Resolves #1536

### DIFF
--- a/app/controllers/bulk_ingest_controller.rb
+++ b/app/controllers/bulk_ingest_controller.rb
@@ -38,29 +38,29 @@ class BulkIngestController < ApplicationController
           persisted_members = []
 
           # Only append one file as a FileSet to one member of one parent until browse-everything can provide links to parent resource IDs
-          selected_files.each do |selected_file|
-            member_change_set = build_change_set(title: selected_file.file_name, pending_uploads: [selected_file])
+          new_pending_uploads.each do |pending_upload|
+            member_change_set = build_change_set(title: pending_upload.file_name, pending_uploads: [pending_upload])
             persisted_members << buffered_changeset_persister.save(change_set: member_change_set)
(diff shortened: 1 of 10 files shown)


app/controllers/bulk_ingest_controller.rb
```diff
@@ -133,22 +135,49 @@ def build_change_set(attrs)
       change_set
     end
 
-    def selected_files_param
-      return {} unless params.key?(:selected_files) && !params[:selected_files].empty?
-      params[:selected_files].to_unsafe_h
+    # Retrieve the selected_files parameter from the request
```
escowles: Do you think we should extract this code to a module that could be shared with BaseResourceController?  This doesn't need to be addressed in this PR, but it would be good to create an issue if the code can be shared.

escowles: I've created https://github.com/pulibrary/figgy/issues/2784 to track that.

PR #2752, opened 2019-03-19
Adding PDF support to ephemera folders

escowles:


app/resources/ephemera_folders/ephemera_folder_decorator.rb
```diff
@@ -71,6 +73,22 @@ def first_range
     @first_range ||= Array.wrap(date_range).map(&:decorate).first
   end
 
+  def pdf_file
+    pdf = file_metadata.find { |x| x.mime_type == ["application/pdf"] }
+    pdf if pdf && Valkyrie::StorageAdapter.find(:derivatives).find_by(id: pdf.file_identifiers.first)
```
jrgriffiniii: Is `if pdf` here necessary?  Should `Valkyrie::StorageAdapter.find(:derivatives).find_by(id: pdf.file_identifiers.first)` raise an error, that would be handled in the `rescue` below, and returning `pdf` regardless of whether or not it is `nil` seems expected.

escowles: I think it's needed, if I remove it, I get the error:

```
NoMethodError: undefined method `file_identifiers' for nil:NilClass
```

PR #2754, opened 2019-03-19
Enables users to disable the downloading of resources within the Universal Viewer

### LINKED ISSUE #2628: Users should be able to mark a resource as non-downloadable
This will let them see the synced resource and go through the viewer, but not download a PDF, images, or anything else. Should disable the download frame in UV.

jrgriffiniii:
Resolves #2628 by addressing the following:
- Providing downloadable behavior for `ResourceDecorators`
- Ensuring that the downloadable attribute is set to a controlled vocabulary
- Adding downloadable as a common attributes, as an attribute for `EphemeraBoxes`,
`EphemeraFolders`, `FileSets`, `MediaResources`, `NumismaticIssues`, `Playlists`,
`RasterResources`, `ScannedMaps`, `Recording`, `VectorResources`


app/resources/scanned_resources/scanned_resource_change_set.rb
```diff
@@ -46,6 +47,7 @@ def primary_terms
       :title,
       :source_metadata_identifier,
       :member_of_collection_ids,
+      :downloadable,
```
escowles: Please move this just after `pdf_type`

app/resources/scanned_resources/simple_change_set.rb
```diff
@@ -79,6 +80,7 @@ def primary_terms
     [
       :title,
       :sort_title,
+      :downloadable,
```
escowles: Please move this just after `pdf_type`

app/helpers/application_helper.rb
```diff
@@ -178,9 +178,12 @@ def manifest_url(resource)
   end
 
   # Generate the path for the Universal Viewer iframe @src attribute
+  # @param [Valkyrie::Resource]
   # @return [String]
   def universal_viewer_path(resource)
-    "/uv/uv#?manifest=#{manifest_url(resource)}&config=#{root_url}/uv/uv_config.json"
+    config_path = resource.decorate.downloadable? ? "uv_config.json" : "uv_config_downloads_disabled.json"
```
escowles: I think we should show the regular viewer to users who are `admin` or `staff` — e.g., so they can download a PDF or page images to fulfill a user request.

PR #2755, opened 2019-03-20
Add nested numismatic place to numismatic issue

### LINKED ISSUE #2741: Add nested place form to numismatic issue

eliotjordan:
- Creates a numismatic place resource
- Adds nested numismatic place to numismatic issues
- Ensures that all nested resources are populated during changeset syncing.
- Creates a common populator `populate_nested_properties` in ChangeSet for all single-valued nested properties to use.
- Creates a common `prepopulate!` method in ChangeSet so that it doesn't have to be re-implemented in each changeset with a nested resource.

Closed #2741


app/change_sets/change_set.rb
```diff
@@ -76,4 +76,48 @@ def fix_multivalued_keys
...
+      send(populator, fragment: fields, as: name)
+    end
+  end
+
+  # Override prepopulate method to correctly populate nested properties.
+  def prepopulate!(_args = {})
+    schema.each(twin: true) do |property|
+      property_name = property[:name]
+      property_klass = model.class.schema[property_name.to_sym]
+      send(:"#{property_name}=", property_klass.new) unless send(property_name)
```
tpendragon: If `twin: true` finds only the nested ones for some reason (like the above comment), can you link to that documentation?

app/change_sets/change_set.rb
```diff
@@ -76,4 +76,48 @@ def fix_multivalued_keys
...
+    end
+
+    super
+  end
+
+  # Trigger population of nested properties when syncing changeset.
+  # Overrides {Disposable::Twin#sync}
+  def sync
```
tpendragon: Why is this necessary if you're adding the populator to the necessary properties in the ChangeSet?

app/change_sets/change_set.rb
eliotjordan: I found this in reform. It uses it to find nested properties to validate: https://github.com/trailblazer/reform/blob/master/lib/reform/contract/validate.rb#L44-L47

app/change_sets/change_set.rb
eliotjordan: The populator is not run when syncing the resource outside of a controller action. Instantiating a resource with the nested property in FactoryBot is an example. I'm sure why exactly, but I guess this is something buried within reform. Without the populator, when you run `sync` on the changeset, an exception is raised because reform (?) has mapped that nested property to a changeset.

eliotjordan: @tpendragon I created an branch with an example that raises the same exception with the existing letter change set. 

https://github.com/pulibrary/figgy/compare/nested-error-example

tpendragon: Does that still error if for sender you instead set it as just the properties and not the object?

PR #2779, opened 2019-03-26
Import/Export Dependent Objects into Bags

tpendragon:
This does a few things:

1. Adds a test for Ephemera with Vocabularies
2. Stores all referenced objects, recursing deep, into the bag of the parent on export. This bag is considered a preservation artifact, so the hope is that it's self-contained.
3. Don't overwrite objects on import if they already exist.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -203,6 +203,7 @@ RSpec/DescribeClass:
     - 'spec/inputs/**/*'
     - 'spec/requests/**/*'
     - 'spec/resources/**/*routes_spec.rb'
+    - 'spec/adapters/bagit/bag_roundtrip_spec.rb'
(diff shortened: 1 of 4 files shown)


escowles:
I tried this out with an ephemera folder in my dev instance.  Importing an item from an exported bag worked well, except that membership in the box (and therefore project) wasn't imported correctly.

tpendragon:
What do you mean by not correctly? What happened?

escowles:
It was not associated with a box or project at all.  I'm guessing this is because the box links to the folder, not the other way around.

tpendragon:
I think that's accurate behavior. If you export a folder it marks the boundary of association at that folder. If you had exported and then imported the box it should have made the folder as a member. Maybe I'm misunderstanding though.

Issue #2773, opened 2019-03-25
Implement Cloud Fixity Auditing


escowles:
Child of pulibrary/figgy#1876
Depends on pulibrary/figgy#2804

- [ ] Randomly retrieve preservation Bags and fixity check them
  - A javascript function will be executed with a resource ID, which will fetch the object from GCS, figure out the checksum, and put a success or failure message on a Google Cloud pub/sub queue. Figgy will run a worker to pull down those messages and create Events to store that information locally.
- [ ] Add status to Figgy object so you can tell when they were checked and what the outcome was
- [ ] Add any of this kind of failure to the Fixity dashboard.

jrgriffiniii:
Google offers a Ruby client library for interacting with the API on https://cloud.google.com/storage/docs/reference/libraries#client-libraries-install-ruby which I shall be using in an initial approach.

Issue #2554, opened 2019-02-08
Update preservation bags when files or metadata are updated


escowles:
Child of #1876

To efficiently update preservation bags, don't re-upload unchanged files.

tpendragon:
Do we still want to do this? Creating a bag is pretty expensive, both monetarily and in compute resources, and there are a lot of saves that happen when it comes to things like "add a page and put it in the right spot in the book." Maybe we just need a notification if the uploaded date on the bag is less than the modified date of the resource, with a button for "preserve me again please"?

Alternatively we could have some sort of cron job to automatically do a re-persist if a thing hasn't been changed in some span of time, but has been changed since the bag was preserved (like...a week? A couple weeks?)

Issue #911, opened 2018-03-14
Create preservation bags on workflow "complete"


hackmastera:
blocked by https://github.com/pulibrary/figgy/issues/912
Child of #1876

1. Create bag on disk
2. Tar/Gzip Bag
3. Generate CRC32 Checksum of Bag
4. Upload to GCS
5. Attach valkyrie identifier and MultiChecksum to the resource which was bagged up (probably through a FileMetadata object.)

Logic for when to create a bag should be something in the ChangeSetPersister which says: "Am I Complete, do I have my preservation flag marked, and have I never generated a bag - if so, kick off a background job to generate a bag"

hackmastera:
At what level should the bags be created? Use the same logic used to display resources in search results? If so, is that logic easily re-usable? where is it?

Are there other questions about collection bags?

PR #2781, opened 2019-03-26
[WIP] Implements a RemoteChecksumJob and RemoteChecksumService

jrgriffiniii:
This is not complete, but I am interested in clarifying my understanding of #2773 before advancing this.  This would ensure that an original file for a `FileSet` could be uploaded to a Google Cloud Storage Bucket, and have its checksum calculated for any given resource.

Is this a desired pattern?  Or should the job rely upon the ZIP-compressed directory created by the  `BagExporter` as the resource to upload into the cloud?


escowles:
Yes, I think it's good to check individual files in cloud storage — other than Glacier, I think the APIs are such that we don't need to take the extra step of packaging in zip/tar archives.

I think the workflow will be to create a Bag on local disk, and then push all the files in the Bag to cloud storage, and then to check them individually.  It's probably worth exploring whether the local filesystem is required, or if we could explore generating Bags directly to cloud storage.

tpendragon:
I gotta say, I'm liking the idea of zipping the bags.

escowles:
I have mixed feelings about zipping Bags — on one hand it definitely makes it easier to move them around and provides a nice package for users who work with them.  But it adds a layer of hassle when working with them.

Also, I've had trouble in the past with archive libraries and command-line utilities working consistently with archives greater than a certain size (IIRC, 4GB).  I tested a bunch of different tools and found only Gnu Tar could actually create and read archives bigger than that (other options including Java's Tar/Zip support, InfoZip, and BSD Tar all had serious issues that resulted in unreadable archives). Things may have changed in that area since then (this was 5+ years ago), but I think we should be careful to make sure that anything we do here is going to scale up to our largest objects.

tpendragon:
Sorry, when I say zip I mean .tar.gz, it's a bad habit. I hadn't ever heard of problems with it, especially with such low file sizes! We'll definitely have to triple check that it works.

I like the idea of zipping up the bags because it means we have a single file that we can upload, we can associate that file with the metadata (add it to the file_identifiers for a ScannedResource for instance, or even create a FileMetadata node to store the checksum of that zip file), and we can version the bucket so if we keep the name the same we can keep all copies of it forever. Checksumming could become a single check.

jrgriffiniii:
This PR is still unfinished, but there are a number of approaches I can take moving ahead:

`RemoteBagChecksumJob` now creates and appends a `FileSet` for a Bag.  If the bag should be compressed, it compressed it into a single `FileMetadata` node.  Currently ZIP and TAR compression are supported for these.  For uncompressed Bags, the `FileMetadata` nodes are created for each file in the Bag, and these are stored as members of the Bag FileSet.  The checksums here should be stored at the `FileMetadata` layer.

`RemoteBagFileSetChecksumJob` takes my initial approach, where a FileSet is created for a compressed bag, and the checksum is stored at the `Resource` layer.  For uncompressed Bags, each Bag item has a FileSet created, and the checksum is stored on each FileSet (at the `FileSet` layer).

I'm still uncertain as to whether or not compression should be supported, and where best to persist the checksum information (my suspicion is that using the `FileMetadata` would be best).  I can continue to clean this code and improve the tests until this is determined.

escowles:
@jrgriffiniii Let's talk about this on Monday.  I was envisioning two independent processes:
1. When an object with a preservation policy set is created or updated, Figgy spawns a job that Bags the object and persists the Bag to cloud storage.  (#911 / #2554)
2. A GCE instance constantly chooses a random Bag, verifies it, repeating ad infinitum (#2773). These two processes definitely depend on each other, because they need to agree on what the Bags look like, where they are, and how to find them.  But I wasn't thinking that the audit service would create the Bags, just check the ones that already exist.

jrgriffiniii:
> These two processes definitely depend on each other, because they need to agree on what the Bags look like, where they are, and how to find them.

My understanding is that Bags are either directories or compressed files.  The preference seemed to have been for directories, but I don't recall what the final determination was given https://github.com/pulibrary/figgy/pull/2781#issuecomment-477284246

I can tailor this PR to address either case.

As for storing and finding the Bags, I'm simply delegating to the `BagExporter` here: https://github.com/pulibrary/figgy/pull/2781/files#diff-90141f9e31e7124f30bef4f6ba255f22R104

From what I could best determine, this does not provide any approach to compression (if that is desired).

> But I wasn't thinking that the audit service would create the Bags, just check the ones that already exist.

I can adjust this in the PR, https://github.com/pulibrary/figgy/pull/2781/files#diff-90141f9e31e7124f30bef4f6ba255f22R15 does create the `FileMetadata` nodes on a `FileSet` if they don't exist, and that can be removed.

jrgriffiniii:
A strategy for using a separate, cloud-based service is going to be tested as an alternative to the approach proposed in this PR.

PR #2783, opened 2019-03-26
Generate derivatives on import

tpendragon:


app/adapters/bagit/bag_importer.rb
```diff
@@ -42,6 +42,7 @@ def import!
         resource = metadata_adapter.persister.save(resource: bag_resource)
         import_members!
         import_references!
+        RegenerateDerivativesJob.perform_now(resource.id.to_s) if resource.is_a?(FileSet)
```
escowles: Do we want to background the derivatives?

tpendragon: I ran them in line because I wasn't sure how to get it to happen after the commit closes. I can take another look though.

PR #2793, opened 2019-04-01
Add 'Has Structure' facet

### LINKED ISSUE #2775: Facet for structure?
Mentioned as a possibility by @kelea99 — should we have a facet for whether an item has structure or not?  This would allow users to complete items and make them public, but still keep track of which items needed structural metadata to be added.  Right now, some items are kept in `pending` or `metadata_review` states to track which items still need structure.

escowles:
* Adds Yes/No facet for whether an item has logical structure
* Alphabetizes indexer in initializer for readability

Fixes #2775

### DIFF
--- a/spec/indexers/facet_indexer_spec.rb
+++ b/spec/indexers/facet_indexer_spec.rb
@@ -30,4 +30,36 @@ RSpec.describe FacetIndexer do
       end
     end
   end
+
+  context "when the resource has structure" do
+    it "indexes its presence" do
+      file_set = FactoryBot.create_for_repository(:file_set)
+      scanned_resource = FactoryBot.create_for_repository(
+        :scanned_resource,
+        member_ids: file_set.id,
+        thumbnail_id: file_set.id,
+        logical_structure: [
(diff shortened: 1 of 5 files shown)


spec/indexers/facet_indexer_spec.rb
```diff
@@ -30,4 +30,36 @@
...
+  end
+
+  context "when the resource does not have structure" do
+    it "indexes its absence" do
+      file_set = FactoryBot.create_for_repository(:file_set)
+      scanned_resource = FactoryBot.create_for_repository(
+        :scanned_resource,
+        member_ids: file_set.id,
+        thumbnail_id: file_set.id,
+        logical_structure: []
```
tpendragon: Does this test still pass if you don't set this? I want to make sure `logical_structure` is never nil, because the indexer would throw an error.

escowles: The test passes even if there's no logical_structure (or the object doesn't respond to `logical_structure`) — I started out doing a bunch of checking to make sure to avoid errors, but wound up just doing an unqualified `rescue` to make it simpler and catch any error that might happen: https://github.com/pulibrary/figgy/pull/2793/files#diff-2adfa7e07f80ce4d413250621142b804R47

PR #2807, opened 2019-04-02
Architecture Decision #3: Preservation

escowles:


escowles:
One question I have is how these decisions impact #2554 — and how exactly replacing bags will work, and what the cost ramifications are.  If we update all bags monthly, and therefore recreate and re-upload all bags with Voyager metadata, does the storage cost include the old versions of the bag, or is the 90-day minimum count that as a single file?

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,67 @@
...
+objects.
+
+## Decisions
+
+1. Creating bags
+   1. We will periodically scan for objects which meet the following criteria:
+      1. Status is complete
+      2. Not bagged more recently than their `updated_at` timestamp
+      3. Has a `cloud` preservation policy assigned
+   1. In order to avoid redundant processing when there are multiple updates in a short period of time,
```
hackmastera: We do still have a ticket for doing this on create. Remove that ticket and roll everything into this one process?

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,67 @@
...
+      we will not trigger bagging from a ChangeSetPersister or other event-driven mechanism.
+1. Packaging bags
+   1. We will export an object, and all of its members recursively, including the metadata for any Valkyrie
+      object it links to (such as collections, controlled vocabulary terms, etc.).
+   1. We will include both files and metadata (serialized as JSON).
+   1. We will package the bag directory in a Tar archive, compressed with GZip.
+   1. A link to the Tar/GZipped bag file will be added to the object's `file_metadata` property.
```
hackmastera: Does the object already have a file_metadata or are we creating one? It seems like the process of creating the bag is going to change the updated_at date, triggering another bag creation in the following cycle.

escowles: I agree that bagging the object will definitely update it.  But I was thinking we would set the `bagged_at` time to the same value as the `updated_at`, so it wouldn't show up as being updated more recently than it was bagged.

architecture-decisions/0003-preservation.md
escowles: Yes, I think that's where we wound up in our discussion yesterday.

architecture-decisions/0003-preservation.md
hackmastera: right! yes I forgot about the `bagged_at` date. `updated_at` is set by valkyrie, i think. I'm not sure how to get them to align precisely.

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,67 @@
...
+   1. We will calculate the MD5 checksum of the Tar/GZipped bag and include it in the `file_metadata`.
+   1. We will assign the `file_metadata` a PCDM Use value of `pcdm:preservationArchive`.
```
jrgriffiniii: 👍

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,67 @@
...
+      deleted by reducing the number of staff who have access to the bag storage.
+   1. Storing bags in cloud storage will store them in a geographically separate area using a separate
+      technology stack, reducing vulnerability to natural disaster and malicious attack.
+   1. Using MD5 checksums to verify uploading will reduce the possibility of corruption or truncation in 
+      transfer.  There is a very small chance that the archive might be compromised with a malicious file
+      with a colliding MD5 checksum.  But the risk of this kind of attack is very low.
+1. Deleting bags
+   1. Deleting bags manually means the last preserved state of the object will persist after it is deleted.
+   1. Not deleting bags automatically means that we may accrue bags for deleted objects and waste space. But
+      it also means that preservation bags can be used to restore accidentally deleted objects.
```
jrgriffiniii: Does this imply that bags are a separate mechanism for disaster recovery?  Are these given preferential treatment to restoring the entire system state using PostgreSQL backups?

If so, I would also be concerned about this serving the function of proper versioning support.

escowles: I think that our backups are good for certain scenarios, such as corruption/failure of the database or storage, and (assuming it's noticed quickly) recovering from accidental deletion.

The bags are intended for longer-term recovery and larger-scale disaster recovery (up to and including complete destruction of the datacenter, failure of the entire software stack, etc.).

Versioning is a tricky issue, and I haven't really heard our stakeholders ask for it.  But it's a tool that's available and might be appropriate for addressing certain needs.

escowles:
We decided not to do Bags, but instead to serialize to cloud storage in a Bag-like directory structure, including having children nested under their parents, and files nested in the object, but without the manifest overhead.  @tpendragon will write up an alternative architecture decision record for that approach.

Issue #912, opened 2018-03-14
Set up a place to write bags to


hackmastera:
Child of #1876

Create the mechanism and configuration needed to Bag up an object and write it to preservation storage.

Document how the bucket is created and how to access it.

The key file for access should be added to ansible and set up as an environment variable appropriately.

tpendragon:
Child of #1551

tpendragon:
We haven't decided on a strategy for this yet, so we're closing this as premature.

escowles:
I think this should be directly to cloud storage, if possible.

PR #2820, opened 2019-04-02
Architecture Decision #3: Preservation

tpendragon:
Closes pulibrary/figgy#912

### DIFF
--- a/architecture-decisions/0003-preservation.md
+++ b/architecture-decisions/0003-preservation.md
@@ -1,0 +1,114 @@
+# 3. Preservation
+
+Date: 2019-04-02
+
+## Status
+
+Accepted
+
+## Context
+
+We have agreed that we will preserve digital objects by saving resources in
+Google Cloud Storage in a directory structure which preserves both the binaries
+the resource is made up of as well as the JSON serialization of the resource
+itself.
+
+## Decisions
+
+1. Preserving
+   1. We will preserve materials in Google Cloud Coldline Storage with
+      `versioning` enabled. Versions will be kept indefinitely and without
+      limit. All files will go in a single bucket.
+      - Staging bucket is configured with the following command:
+        ```
+        gsutil mb -c regional -l us-west1 -p pulibrary-figgy-storage-1 gs://figgy-staging-preservation
+        echo '{"rule": [{"action": {"type": "Delete"}, "condition": {"age": 2}}]}' > lifecycle.json
+        gsutil lifecycle set lifecycle.json gs://figgy-staging-preservation
+        rm lifecycle.json
+        gsutil bucketpolicyonly set on gs://figgy-staging-preservation
+        gsutil iam ch serviceAccount:figgy-staging@pulibrary-figgy-storage-1.iam.gserviceaccount.com:objectAdmin gs://figgy-staging-preservation
+        ```
+      - Production bucket is configured with the following command:
+        ```
+        gsutil mb -c coldline -l us-west1 -p pulibrary-figgy-storage-1 gs://figgy-preservation
+        gsutil bucketpolicyonly set on gs://figgy-preservation
+        gsutil iam ch serviceAccount:figgy-preservation-production@pulibrary-figgy-storage-1.iam.gserviceaccount.com:objectAdmin gs://figgy-preservation
+        gsutil versioning set on gs://figgy-preservation
+        ```
+   1. When a resource is `complete` and marked with the `cloud` preservation
+      policy it will save itself and all resources contained in `member_ids` in
+      a directory structure in Google Cloud Storage that looks like the
+      following:
+      ```
+      - <resource-id>
+        - data
+          - <child-id>
+            - <child-id>.json
+            - <binary.tif>
+        - <resource-id>.json
+      ```
+   1. Children are preserved on save if their parents are preserved.
+   1. Related objects such as collections, Ephemera Terms, etc. will not be
+      packaged inside the preserved object. If it's important they be preserved,
+      those objects should be marked with the `cloud` preservation policy.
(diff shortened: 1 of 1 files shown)


architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,84 @@
...
+      with the `cloud` preservation policy, it will upload the new binary
+      content to both the repository and to Google Cloud STorage
+   1. If a child is marked to be preserved, but its parent is not, it will still
+      save in a nested directory structure, but will not automatically create
+      backups of its parents.
+   1. This behavior will be attached to the ChangeSetPersister.
+1. Packaging Details
+   1. Resources created in cloud storage will be attached to the appropriate
+      resource in `file_metadata` with a PCDM Use value of
+      `pcdm:PreservationMasterFile`
```
escowles: We're currently using `pcdm:PreservationMasterFile` for audio master files.  Do we want to avoid using the same value, or is the different storage adapter enough to keep them separate?

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,84 @@
...
+      if necessary.
+   1. When a preserved resource is deleted, we will delete its directory from
+      preservation storage. If we need to get it again, we will look at Google
+      Cloud Storage's stored versions.
+
+## Consequences
+
+1. Structure
+   1. The resultant structure will not be in a format that is expected by
+      outside vendors. However, we have a BagIt packager for those use cases.
```
escowles: Do we want to say something here about being a subset of the BagIt spec?  I.e., you could take this format and add the manifests to make it a valid bag.

architecture-decisions/0003-preservation.md
tpendragon: Yeah, when I actually implemented this I ended up using PreservationCopy and PreservedMetadata. Maybe we should block this and see how we like that?

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,84 @@
...
+            - <child-id>.json
+            - <binary.tif>
+        - <resource-id>.json
+      ```
+   1. When a child with a preserved parent (any length up the hierarchy) is
+      `complete` and marked with the `cloud` preservation policy and is updated,
+      it will update itself.
+   1. Related objects such as collections, Ephemera Terms, etc. will not be
+      packaged inside the preserved object. If it's important they be preserved,
+      those objects should be marked with the `cloud` preservation policy.
```
hackmastera: should this be the default for some class/es of objects?

architecture-decisions/0003-preservation.md
hackmastera: What about "PreservationFile" and "PreservationMetadata"? "Copy" feels potentially vague.

hackmastera:
I think we should say something about how we are going to check fixity; I think that discussion / those decisions inform and are informed by this -- should that be a separate ADR or part of this one?

tpendragon:
We could talk about that here, but I don't think we'll know what exactly to say until we try the options and see what works

architecture-decisions/0003-preservation.md
tpendragon: That feels like a different decision. Or it could be at least.

hackmastera:
@tpendragon maybe we could at least add in something about how checksum will be calculated? You said something in channel about excluding the FileMetadata node.

escowles:
I thought about including fixity checking in my ADR draft, but decided it could be in a separate ADR instead.  But I agree that we should document both sides of the process, whether it's in this ADR or a followup.

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,116 @@
...
+      following:
+      ```
+      - <resource-id>
+        - data
+          - <child-id>
+            - <child-id>.json
+            - <binary.tif>
+        - <resource-id>.json
+      ```
+   1. When a child with a preserved parent (any length up the hierarchy) is
```
hackmastera: Can you reword this a bit? It's confusing whether the child or parent is completed, and which is updated.

tpendragon: That better?

architecture-decisions/0003-preservation.md
```diff
@@ -0,0 +1,114 @@
...
+      and then compared to the checksum of the object stored locally.
+   1. In the case of a failure it will be reported to Figgy and displayed in a
+      dashboard for further follow-up and repair.
+
+## Consequences
+
+1. Structure
+   1. The resultant structure will not be in a format that is expected by
+      outside vendors. However, we have a BagIt packager for those use cases,
+      and this structure can be easily converted to a BagIt bag via calculation
```
escowles: "calculation" seems ambiguous here.  We already have the relevant checksums for the data files, right? So we could create a Bag without needing to re-checksum those (which will make it much faster and prevent missing corruption)

architecture-decisions/0003-preservation.md
escowles: 👍

architecture-decisions/0003-preservation.md
tpendragon: K, changed.

escowles: Thanks!

rosy1280:
I'm not sure where to comment this so apologies if this is the wrong place, but the resulting preservation design document sounds a lot like OCFL. Have you considered looking at it? https://ocfl.io

tpendragon:
@rosy1280 Could you make an issue? We did look at OCFL, but I think the discussion deserves something not-closed, so we don't lose track of it.

rosy1280:
@tpendragon #2884

PR #2821, opened 2019-04-02
First pass at persistence of ScannedResources on save.

tpendragon:
This is a draft just to look at. It doesn't save to the Cloud yet, but uses generic Valkyrie methods - so we should be able to sub in a cloud StorageAdapter, maybe with patches to accept our PathGenerator.

The directory structure for a MVW looks like this:
```
b741999c-ea71-40f8-933a-50bd3886447e backup_property % tree
.
├── b741999c-ea71-40f8-933a-50bd3886447e.json
└── data
    └── 985ab752-dbab-4301-bfcd-efdfb2bb31e5
        ├── 985ab752-dbab-4301-bfcd-efdfb2bb31e5.json
        └── data
            └── 81c4fe62-79d3-441f-b06f-b5003be85673
                ├── 81c4fe62-79d3-441f-b06f-b5003be85673.json
                └── example-fc9a980c-9468-4a9b-b3ed-5fb5aab66a31.tif

4 directories, 4 files
```


app/change_set_persisters/change_set_persister/preserve_resource.rb
```diff
@@ -0,0 +1,114 @@
...
+    attr_reader :change_set_persister, :change_set, :post_save_resource
+
+    def initialize(change_set_persister:, change_set:, post_save_resource: nil)
+      @change_set = change_set
+      @change_set_persister = change_set_persister
+      @post_save_resource = post_save_resource
+    end
+
+    def run
+      return unless change_set.try(:preserve?)
```
escowles: I suspect we will eventually wind up with multiple preservation policies, e.g. one for Coldline, one for S3, one for campus storage, etc. Or maybe one for Coldline, and a second policy for Coldline and S3.  Maybe this should lookup a config based on the preservation policy instead of being a binary?

app/resources/scanned_resources/scanned_resource_change_set.rb
```diff
@@ -59,4 +59,8 @@ def primary_terms
       :append_id
     ]
   end
+
+  def preserve?
+    Array.wrap(resource.preservation_policy).include?("cloud") && state == "complete"
+  end
```
escowles: Given the decisions we made today (not preserving linked Collections or EphemeraTerms), I think we probably want almost every kind of resource to allow preservation (everything but FileSets?).

tpendragon: Yeah, this was just to get some tests passing.

tpendragon:
I'm not 100% happy with this, but I think it's ready for review. We might not want to let this into production with the field public, though. I'd be happy to remove the field from the form in that case.

tpendragon:
I've got a branch that stores this in GCS and it seems to work, but I'm running into race conditions when adding a file to a complete & preserved item if it's being added after commit. The job for preserving and the job for derivatives/characterization are running at the same time, which is overriding one another's `file_metadata`.

escowles:
@tpendragon Would adding optimistic locking help fix this race condition?

tpendragon:
It'd stop the overwriting, but a bunch of jobs would be retrying.

app/views/records/edit_fields/_preservation_policy.html.erb
```diff
@@ -0,0 +1,8 @@
+  <%= f.input key,
+              as: :select,
+              collection: ControlledVocabulary.for(key).all,
+              label_method: :label,
+              value_method: :value,
+              include_blank: true,
+              input_html: { class: 'form-control rights-statement' },
+              required: f.object.required?(key) %>
```
escowles: We have a few partials that are basically identical to this (e.g., `_holding_location.html.erb` or `_owners.html.erb`). Do you think it's worth making a helper?

config/initializers/valkyrie.rb
```diff
@@ -74,6 +74,48 @@
...
+      :cloud_backup
+    )
+  else
+    Valkyrie::StorageAdapter.register(
+      Valkyrie::Storage::Disk.new(
+        base_path: Figgy.config["disk_preservation_path"],
+        file_mover: FileUtils.method(:cp),
+        path_generator: NestedStoragePath
+      ),
+      :cloud_backup
```
escowles: Since we anticipate using several different cloud storage options, can you change this to something that references Google and/or Coldline?

README.md
```diff
@@ -131,5 +131,18 @@ By default, Figgy provides users with the ability to upload binaries from the lo
 
 Figgy may also be configured to upload files from hosted storage providers.  Support for users with Google Drive accounts has been tested and verified.  [Please reference the Browse Everything documentation for more details](https://github.com/pulibrary/figgy/blob/master/BROWSE_EVERYTHING.md).
 
+## Preservation Configuration in Development
+
+1. Download and save gcs_pulibrary-staging-credentials.json from Keybase into
```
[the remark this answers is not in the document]
tpendragon: That's funny, LastPass is what I meant and where it is. Got the name wrong.

app/jobs/cleanup_files_job.rb
```diff
@@ -5,7 +5,9 @@ class CleanupFilesJob < ApplicationJob
   def perform(file_identifiers:)
     file_identifiers.each do |file_identifier|
       storage_adapter = Valkyrie::StorageAdapter.adapter_for(id: file_identifier)
+      storage_adapter.delete(id: file_identifier.to_s)
```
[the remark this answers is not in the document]
tpendragon: The dirname delete is abusing a quirk of the disk delete implementation. This deletes the actual ID.

app/services/preserver.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  class NullPreserver
+    def self.preserve!; end
+  end
+
+  attr_reader :change_set, :storage_adapter, :change_set_persister
+  delegate :resource, to: :change_set
+  def initialize(change_set:, storage_adapter: nil, change_set_persister:)
+    @change_set = change_set
+    @storage_adapter = storage_adapter || default_storage_adapter
```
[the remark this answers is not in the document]
tpendragon: I think we'd have another preserver maybe? Hard to tell, seems like something we can handle when we have another one.

app/views/records/edit_fields/_preservation_policy.html.erb
tpendragon: Probably, maybe as another issue though?

app/jobs/preserve_children_job.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+class PreserveChildrenJob < ApplicationJob
+  delegate :metadata_adapter, to: :change_set_persister
+  delegate :query_service, to: :metadata_adapter
+
+  def perform(id:)
+    resource = query_service.find_by(id: id)
```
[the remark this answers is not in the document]
tpendragon: This is true. I don't think we want to pass the member ids, in case there are like a thousand. A member query that takes an ID is a good idea, but seems like a Valkyrie issue yeah?

app/services/preserver.rb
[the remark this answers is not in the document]
escowles: 👍

app/jobs/preserve_children_job.rb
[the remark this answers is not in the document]
escowles: Yes, I'm fine sticking with the current approach here, and I've added a Valkyrie ticket for the member query that takes an ID instead of a resource: https://github.com/samvera/valkyrie/issues/703

app/jobs/cleanup_files_job.rb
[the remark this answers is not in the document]
escowles: That makes sense — and doing something different is beyond the scope of this ticket.  And for that matter, updating the Valkyrie API to have a recursive delete function is probably not a good idea.  So maybe the best thing here is to update this to make it clearer that we're getting the dirname to take advantage of the Disk delete behavior?  Something like:

```
file_id = storage_adapter.kind_of?(Valkyrie::Storage::Disk) ? File.dirname(file_identifier.to_s) : file_identifier.to_s
storage_adapter.delete(id: id)
```

app/views/records/edit_fields/_preservation_policy.html.erb
escowles: 👍 I created a `refactor` ticket for that: https://github.com/pulibrary/figgy/issues/2834

app/jobs/cleanup_files_job.rb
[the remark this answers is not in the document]
tpendragon: I considered that, but the storage adapter is currently decorated so I didn't want to test for the class name. I can easily add a comment though?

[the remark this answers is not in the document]
escowles: That's a good point — a comment here would be fine to make it clear why it's different for the cloud vs. disk.

app/services/preserver.rb
```diff
@@ -31,21 +31,38 @@ def preserve!
...
   end
 
-  def preserve_original_file
-    return unless resource.try(:original_file) && preservation_object.binary_nodes.blank?
-    file_metadata = FileMetadata.new(
-      label: preservation_copy_label,
+  def preserve_binary_content
+    binary_nodes.each do |binary_node|
+      next if binary_node.file_identifiers.blank?
+      next if preservation_object.binary_nodes.find { |x| x.preservation_copy_of_id == binary_node.id }
```
escowles: Shouldn't this be `next unless` to skip binaries that aren't marked as the file to be preserved?

tpendragon: No, it's `next if` so it won't re-upload binaries that it's already uploaded.

escowles: I see the test verifying that both the preservation master and intermediate are saved, so I think I am just misreading this: https://github.com/pulibrary/figgy/pull/2821/commits/40f70fe1e4ee75a6a591200e079f6567f14b061f#diff-2361c599e97a22f4ee2ebef0aba33d40R1557

tpendragon: Maybe it's the variable names? Want me to rename it to `resource_binary_nodes` to differentiate from `preservation_object.binary_nodes`?

escowles: No, I think it's fine — I see how it's working now.

spec/change_set_persisters/change_set_persister_spec.rb
```diff
@@ -1517,4 +1517,107 @@
...
+  describe "preservation" do
+    context "when completing a `cloud` preservation_policy resource" do
+      it "saves to a backup location" do
+        file = fixture_file_upload("files/example.tif", "image/tiff")
+        resource = FactoryBot.create_for_repository(:pending_scanned_resource, preservation_policy: "cloud", files: [file])
+        change_set = DynamicChangeSet.new(resource)
+        change_set.validate(state: "complete")
+
+        output = change_set_persister.save(change_set: change_set)
+        expect(Wayfinder.for(output).preservation_object.metadata_node).not_to be_blank
```
hackmastera: I'm nervous about testing that things aren't blank ever since that update_id bug. Is there something more specific we can check? Also see `be_blank` in other places in this spec.

app/services/preserver.rb
```diff
@@ -0,0 +1,125 @@
+# frozen_string_literal: true
+class Preserver
+  # Provided as a mechanism for switching preservation strategies depending on
+  # the profile. For now, there's only one, so fall down to new.
+  def self.for(change_set:, change_set_persister:, storage_adapter: nil)
+    return NullPreserver unless change_set.try(:preserve?)
+    new(change_set: change_set, change_set_persister: change_set_persister, storage_adapter: storage_adapter)
+  end
+
+  class NullPreserver
```
hackmastera: 👍

app/services/preserver.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def preserve_metadata
+    metadata_node = preservation_object.metadata_node || build_metadata_node
+    uploaded_file = storage_adapter.upload(file: temp_metadata_file.io, original_filename: metadata_node.label.first, resource: resource)
+    metadata_node.file_identifiers = uploaded_file.id
+    preservation_object.metadata_node = metadata_node
+    change_set_persister.metadata_adapter.persister.save(resource: preservation_object)
+  end
+
+  def metadata_checksum
```
hackmastera: Can you move this method below its first use?

app/services/preserver.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def binary_nodes
+    [:original_files, :intermediate_files, :preservation_files].flat_map do |node_type|
+      Array(resource.try(node_type))
+    end
+  end
+
+  def file_metadata_from_binary_node(node)
+    label, splitter, extension = node.label.first.to_s.rpartition(".")
+    label = "#{label}-#{node.id}#{splitter}#{extension}"
```
hackmastera: it's a bit hard to visualize what this will look like. Maybe you could give some comments with examples, either here or in the specs.

app/services/preserver.rb
```diff
@@ -0,0 +1,125 @@
...
+  delegate :resource, to: :change_set
+  def initialize(change_set:, storage_adapter: nil, change_set_persister:)
+    @change_set = change_set
+    @storage_adapter = storage_adapter || default_storage_adapter
+    @change_set_persister = change_set_persister
+  end
+
+  def preserve!
+    preserve_binary_content
+    # Don't preserve children unless this is the first time it's being
```
hackmastera: Can you move this comment above the method? I totally missed the first line of the method when I read it the first time.

app/services/preserver.rb
```diff
@@ -0,0 +1,125 @@
...
+      preserve_metadata
+    else
+      preserve_metadata && preserve_children
+    end
+  end
+
+  def preserve_binary_content
+    binary_nodes.each do |binary_node|
+      next if binary_node.file_identifiers.blank?
+      next if preservation_object.binary_nodes.find { |x| x.preservation_copy_of_id == binary_node.id }
```
hackmastera: the id contains the checksum?

app/services/preserver.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def metadata_checksum
+    @metadata_checksum ||=
+      begin
+        io = StringIO.new(preservation_metadata.to_json)
+        MultiChecksum.for(Valkyrie::StorageAdapter::File.new(io: io, id: "tmp"))
+      end
+  end
+
+  # Don't preserve the PreservedMetadata FileMetadataNode, because it's
```
hackmastera: I feel like this comment might be more helpful somewhere else. I can't really figure out how it's relevant here. Or maybe it needs to be rewritten? I think it may help to have the associated ADR revised / nailed down; this could maybe point there?

tpendragon: Oh, I think this is just an old comment. It's actually pretty irrelevant now.

app/resources/nested_resources/file_metadata.rb
```diff
@@ -29,6 +29,9 @@ class FileMetadata < Valkyrie::Resource
   attribute :fixity_success, Valkyrie::Types::Integer
   attribute :fixity_last_success_date, Valkyrie::Types::DateTime.optional
 
+  # preservation attributes
+  attribute :preservation_copy_of_id, Valkyrie::Types::ID.optional
```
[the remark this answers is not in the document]
hackmastera: Sorry I meant to come back and check on this comment again before submitting!! I'm still not sure what it means, though, and I see I was confused when it was used, also. Maybe an extra comment would help?

escowles:
I am 👍 on removing the dropdown, but I would appreciate a rake task or some other way short of console to tag things for preservation.

tpendragon:
> I am 👍 on removing the dropdown, but I would appreciate a rake task or some other way short of console to tag things for preservation.

What form would you like this to take? A "pass an ID to preserve it" task? We've typically ended up going to the side of those to use the console instead, because of Rails startup times.

escowles:
I agree that "tag one object for preservation" wouldn't be very useful.  Eventually, I expect to add this to the bulk edit form, so adding a "preserve everything in this collection" service/Rake task is probably the better way to start.

tpendragon:
> I agree that "tag one object for preservation" wouldn't be very useful. Eventually, I expect to add this to the bulk edit form, so adding a "preserve everything in this collection" service/Rake task is probably the better way to start.

Would you be against adding that as another follow-on issue?

escowles:
@tpendragon I'm fine with the rake task being in a follow-on issue: https://github.com/pulibrary/figgy/issues/2841

app/resources/nested_resources/file_metadata.rb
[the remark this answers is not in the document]
hackmastera: 👍 that clarifies it. I was wondering about this, actually. From the code it seems that once a file is preserved it will never be preserved again. Is that because of our understanding that a file can never be changed, but instead would be deleted / re-created? I have a vague memory that there is a geo metadata file that doesn't align with this assumption.

app/services/preserver.rb
```diff
@@ -0,0 +1,99 @@
...
+    uploaded_file = storage_adapter.upload(file: temp_metadata_file.io, original_filename: metadata_node.label.first, resource: resource)
+    metadata_node.file_identifiers = uploaded_file.id
+    preservation_object.metadata_node = metadata_node
+    change_set_persister.metadata_adapter.persister.save(resource: preservation_object)
+  end
+
+  def metadata_node
+    @metadata_node ||=
+      begin
+        preservation_object.metadata_node ||
```
hackmastera: It seems like the metadata node always pulls from the existing value; how does it get refreshed?

tpendragon: A new file gets uploaded and the metadata node's file identifiers gets overwritten up on line 68.

eliotjordan: It's true. FGDC metadata files are (for Vector and Raster Resources at this point) are "enhanced" with a URL pointing to Pulmap after the parent resource is saved. It's terrible and I hate it, but it's happening. I do have some ideas on how to avoid this in the future...

https://github.com/pulibrary/figgy/blob/master/app/services/fgdc_update_service.rb

hackmastera: Checksum!

tpendragon:
We're blocking this PR until two edge cases are handled:

1. One volume of a multi-volume ScannedMap is removed from its parent. Its nested directory should get deleted.
2. When binary content is updated in place, as in the case of Fgdc metadata, it should update the binary content in preservation.

tpendragon:
Okay, I think the fgdc use case is handled.

PR #2824, opened 2019-04-03
Don't sync a coin to Ol or build an Ol document  if there is no parent numismatic issue. 

### LINKED ISSUE #2761: [Figgy/staging] NoMethodError: undefined method `first_range' for nil:NilClass
## Backtrace

    line 46 of [PROJECT_ROOT]/app/resources/coins/orangelight_coin_builder.rb: document_hash
    line 11 of [PROJECT_ROOT]/app/resources/coins/orangelight_coin_builder.rb: build
    line 13 of [PROJECT_ROOT]/app/services/orangelight_document.rb: document

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/47329153)

### LINKED ISSUE #2865: Don't sync a coin to OL if there is no parent numismatic issue

christinach:
closes #2865, closes #2761

### DIFF
--- a/app/resources/coins/orangelight_coin_builder.rb
+++ b/app/resources/coins/orangelight_coin_builder.rb
@@ -13,15 +13,27 @@ class OrangelightCoinBuilder
 
   private
 
+    def coin_builder_error
+      { error: "#{decorator.title.first} with id: #{decorator.id} has no parent numismatic issue and cannot build an OL document." }
+    end
+
     def clean_document(hash)
-      hash.delete_if do |_k, v|
-        v.nil? || v.try(:empty?)
+      if hash.nil?
+        coin_builder_error
+      else
+        hash.delete_if do |_k, v|
+          v.nil? || v.try(:empty?)
+        end
       end
     end
 
     # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
     def document_hash
+      document_coin_hash.merge(document_parent_hash) if document_parent_hash.present?
+    end
+
+    def document_coin_hash
       {
         id: decorator.orangelight_id,
         title_display: "Coin: #{decorator.coin_number}",
@@ -47,9 +59,15 @@ class OrangelightCoinBuilder
         size_s: decorator.size,
         technique_s: decorator.technique,
         weight_s: decorator.weight,
+        holdings_1display: holdings_hash
+      }
+    end
+
+    def document_parent_hash
+      return unless parent
+      {
         pub_date_start_sort: parent.first_range&.start&.first.to_i,
         pub_date_end_sort: parent.first_range&.end&.first.to_i,
-        holdings_1display: holdings_hash,
         issue_object_type_s: parent.object_type,
         issue_denomination_s: parent.denomination,
         issue_denomination_sort: parent.denomination&.first,
--- a/app/services/event_generator/orangelight_event_generator.rb
+++ b/app/services/event_generator/orangelight_event_generator.rb
@@ -32,7 +32,8 @@ class EventGenerator
     end
 
     def valid?(record)
-      record.is_a? Coin
+      return true if record.is_a?(Coin) && record.decorate.parents.present?
+      false
     end
 
     private
--- a/spec/services/orangelight_document_spec.rb
+++ b/spec/services/orangelight_document_spec.rb
@@ -157,5 +157,31 @@ describe OrangelightDocument do
         expect(output[:issue_artists_sort]).to eq "artist person, artist role"
       end
     end
+
+    context "with a coin not attached to an issue" do
+      subject(:builder) { described_class.new(coin) }
+      let(:coin) do
+        FactoryBot.create_for_repository(:coin,
+                                         holding_location: "Firestone",
+                                         counter_stamp: "two small counter-stamps visible as small circles on reverse, without known parallel",
+                                         analysis: "holed at 12 o'clock, 16.73 grams",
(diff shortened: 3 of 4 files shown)


eliotjordan:
@christinach I've emailed Alan about coins without issues. If he says that we should account for coins without a parent issue, then I'll merge this. Otherwise, I think we should block syncing of coins without parents.

Maybe here: https://github.com/pulibrary/figgy/blob/master/app/services/event_generator/orangelight_event_generator.rb#L34

eliotjordan:
@christinach Alan says all valid coins should have a parent, so I we should block syncing of coins when that isn't the case.

christinach:
I separated the commits in one preventing to sync the Ol document if the coin has no parent and the other preventing to build the Ol document when visiting the /orangelight url in figgy(this triggered the honeybadger error and issue #2761 ) if the coin has no parent.

spec/services/orangelight_document_spec.rb
```diff
@@ -157,5 +157,37 @@
         expect(output[:issue_artists_sort]).to eq "artist person, artist role"
       end
     end
+
+    context "with a coin not attached to an issue" do
+      subject(:builder) { described_class.new(coin) }
+      let(:file) { fixture_file_upload("files/example.tif", "image/tiff") }
```
eliotjordan: Not sure if we need to attach files if we're just testing a logger output.

spec/services/orangelight_document_spec.rb
```diff
@@ -157,5 +157,37 @@
...
+      let(:coin) do
+        FactoryBot.create_for_repository(:coin,
+                                         files: [file],
```
eliotjordan: Same. Don't think files are needed here.

spec/services/orangelight_document_spec.rb
```diff
@@ -157,5 +157,37 @@
...
+                                         size: "27",
+                                         technique: "Cast",
+                                         weight: "8.26")
+      end
+      before do
+        allow(Rails.logger).to receive(:error)
+      end
+
+      it "will log an error" do
+        MultiJson.load(builder.to_json, symbolize_keys: true)
```
eliotjordan: We can probably just call `builder.to_json` without invoking `MultiJson.load`.

app/services/event_generator/orangelight_event_generator.rb
```diff
@@ -32,7 +32,9 @@ def record_member_updated(record)
     end
 
     def valid?(record)
-      record.is_a? Coin
+      return true if record.is_a?(Coin) && record.decorate.parents.present?
+      Rails.logger.error("#{Time.current} - #{record.title.first} with id: #{record.id} has no parent numismatic issue and cannot be synced to OL")
```
eliotjordan: Because of the event generators are run in a [queue](https://github.com/pulibrary/figgy/blob/master/app/services/event_generator.rb#L14-L19), this error will be logged for many resource types that create message events. (ScannedResources, Geo resources, Ephemera, etc..). I don't think it's necessary or even desired to log in a `valid?` method. I would remove unless there's a good argument for keeping it. And then it should only be called when the record is a Coin.

app/resources/coins/orangelight_coin_builder.rb
```diff
@@ -17,11 +17,18 @@ def clean_document(hash)
       hash.delete_if do |_k, v|
         v.nil? || v.try(:empty?)
       end
+    rescue NoMethodError
```
eliotjordan: I wonder if rescuing an exception here, it would be better from the user's point of view to return an error hash like we do for GeoBlacklight [documents](https://github.com/pulibrary/figgy/blob/master/app/services/geo_discovery/geoblacklight_document.rb#L32-L35). Something like...

```
def build
  if parent
    clean_document(document_hash)
  else
    {error: "#{decorator.title.first} with id: #{decorator.id} has no parent numismatic issue and cannot build an OL document."}
  end
end
```

app/services/event_generator/orangelight_event_generator.rb
christinach: Thanks for that. I did notice that different validations run but didn't review it the other way back for example validating a different resource. It would still go through the coin validation and log the error. I would like to log it but it doesn't bother me to remove it.

app/resources/coins/orangelight_coin_builder.rb
christinach: I think this is only in figgy and I'm not sure if the user would visit ../orangelight to check the OL document. This was my initial thought to rescue it and log it for our eyes and not the user's.

christinach: but I'm happy to check it. Thanks for reviewing!

app/services/event_generator/orangelight_event_generator.rb
christinach: another thought I had following the event generator was not to generate an event at all when the valid? is false

app/resources/coins/orangelight_coin_builder.rb
christinach: Updated with the error message in clean_document(hash).

eliotjordan: I'm concerned that rescuing might mask other errors, and I would consider you and I to be "users" when we manually visit `../orangelight`. For me, getting the feedback directly is way easier than digging through the logs.

christinach: yes, I did update maybe you missed the comment above :)

Issue #2776, opened 2019-03-26
Convert citations and artists to nested forms


eliotjordan:
Add citations and artist nested forms to NumismaticIssue and citation nested forms to Coin.

PR #2829, opened 2019-04-04
Nested numismatic citations

eliotjordan:
- Converts numismatic citations to nested properties in NumismaticIssue , Coin, and NumismaticAccession resources.
- Renames `citation` to `numismatic_citation`. Helps to distinguish nested numismatic citations from non-nested citations in other resource types.

Partially addresses #2776


app/wayfinders/base_wayfinder.rb
```diff
@@ -45,6 +45,30 @@ def self.relationship_by_property(relationship, property:, singular: false, mode
...
+    define_method relationship do
+      return instance_variable_get(:"@#{relationship}") if instance_variable_get(:"@#{relationship}")
+      instance_variable_set(:"@#{relationship}", query_service.custom_queries.find_nested_references_by(resource: resource, nested_property: nested_property, property: property).to_a)
+    end
+    define_method "decorated_#{relationship}" do
+      return instance_variable_get(:"@decorated_#{relationship}") if instance_variable_get(:"@decorated_#{relationship}")
+      instance_variable_set(:"@decorated_#{relationship}", __send__(relationship).map(&:decorate))
+    end
+    define_singular_relation(relationship) if singular
+  end
```
eliotjordan: This is where a new query for relationships on nested properties is added to the wayfinder. An example is resolving NumismaticReferences on a nested NumismaticCitation. In keeping with the spirit of the Wayfinder, I decided to implement this a generic set of methods that call a custom query. Alternatively, we could add methods such as:

```
def numismatic_references
  resource.numismatic_reference_ids.map do |id|
    query_service.find_by(id: id)
  end
end
```

It has to be implemented on a resource's wayfinder each for specific case, but maybe it's more performant? I'm fine with either approach.

eliotjordan: Implemented this instead of using inefficient queries.

app/views/coins/_numismatic_citation_fields.html.erb
```diff
@@ -0,0 +1,31 @@
+<div class='nested-fields'>
```
escowles: Can we put this in views/shared instead of replicating it for each type that has citations?

eliotjordan:
@escowles I moved that partial to shared location.

PR #2831, opened 2019-04-04
Strip spaces from incoming form data

### LINKED ISSUE #2828: Exception is raised when a source metadata identifier has an extra space
Should we trim any accidental spaces?

## Backtrace

    line 44 of [PROJECT_ROOT]/app/services/pul_metadata_services/client.rb: retrieve_from_pulfa
    line 14 of [PROJECT_ROOT]/app/services/pul_metadata_services/client.rb: retrieve
    line 44 of [PROJECT_ROOT]/app/models/remote_record.rb: client_result

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/47767616)

```
URI::InvalidURIError: bad URI(is not URI?): CD-18264 .xml

https://figgy.princeton.edu/concern/scanned_resources/new/recording
```

https://github.com/pulibrary/figgy/blob/master/app/services/pul_metadata_services/client.rb#L44

hackmastera:
closes #2828

### DIFF
--- a/spec/resources/scanned_resources/scanned_resources_controller_spec.rb
+++ b/spec/resources/scanned_resources/scanned_resources_controller_spec.rb
@@ -201,6 +201,21 @@ RSpec.describe ScannedResourcesController, type: :controller do
       end
     end
 
+    context "when an attribute has leading / trailing spaces" do
+      let(:resource) { FactoryBot.create_for_repository(:scanned_resource) }
+      let(:params) do
+        {
+          source_metadata_identifier: " AC044_c0003 "
+        }
+      end
+      it "strips them" do
+        patch :update, params: { id: resource.id.to_s, scanned_resource: params }
+
+        reloaded = find_resource(resource.id)
+        expect(reloaded.source_metadata_identifier.first).to eq "AC044_c0003"
+      end
+    end
+
     context "when a published scanned resource has its title updated" do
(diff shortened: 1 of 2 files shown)


spec/resources/scanned_resources/scanned_resources_controller_spec.rb
```diff
@@ -201,6 +201,21 @@
       end
     end
 
+    context "when an attribute has leading / trailig spaces" do
```
escowles: Typo: "trailig"

spec/resources/scanned_resources/scanned_resources_controller_spec.rb
```diff
@@ -201,6 +201,21 @@
...
+      let(:resource) { FactoryBot.create_for_repository(:scanned_resource) }
+      let(:params) do
+        {
+          source_metadata_identifier: " CD-18264 "
```
escowles: I'm 👍 on the implementation, but not sure about using an invalid source metadata identifier in this test.

hackmastera: hm, good point. I was using the data from the error, but it's incidental that it was bad. I'll sub in something valid.

hackmastera:
@escowles I made the changes you requested and there were also some broken specs which required updates to the implementation. If you could re-review!

Issue #2838, opened 2019-04-05
intermittently failing spec


hackmastera:
Here's a run on CI; it's also been observed in a local dev environment. https://circleci.com/gh/pulibrary/figgy/14905#tests/containers/3

RSpec bisect could not isolate the failure, so it's truly intermittent.

eliotjordan:
@hackmastera What do you think about removing the test that counts the number of records? We're already testing that an error was caught and logged. This is popping up all the time now; to the point where I'm having trouble getting PRs to pass CI.

Added a commit to do that here:
https://github.com/pulibrary/figgy/pull/2850/commits

hackmastera:
@eliotjordan I don't love it. If the record is indexed why would there be an error returned?

Could you put it on a different PR so we can have a larger conversation? If the PR I just pushed doesn't work then maybe it's the best thing to do at least in the short term.

eliotjordan:
I just removed it. How about this?

```
 it "tolerates RSolr::Error::ConnectionRefused, logging bad id" do
      allow_any_instance_of(Valkyrie::Persistence::Solr::Persister).to receive(:save_all).with(resources: resources).and_raise RSolr::Error::ConnectionRefused
      allow_any_instance_of(Valkyrie::Persistence::Solr::Persister).to receive(:save).with(resource: resources[0]).and_raise RSolr::Error::ConnectionRefused

      expect { described_class.reindex_all(logger: logger, wipe: true) }.to change { solr_adapter.query_service.find_all.to_a.length }.by(4)
      expect(logger).to have_received(:error).with("Could not index #{resources[0].id}")
    end
```

hackmastera:
That is an interesting idea!

hackmastera:
resolved in #2852

PR #2849, opened 2019-04-09
Try mocking with wrap_original

hackmastera:
In hopes that the mock wasn't matching on resource due to dependence on instantiation

refs #2838


hackmastera:
didn't work

Issue #2836, opened 2019-04-05
Files duplicated when added to a resource


hackmastera:
Paraphrasing from a slack conversation with Squirrel:

When adding files to a new resource via file manager > upload files from server (I'm uploading them from pudl\rbsc_backup\rbsc_studio.

I'm having an issue uploading images in Figgy. It uploads all of the images, but it also adds a bunch of the files again at the beginning, out of order. Cf https://figgy.princeton.edu/catalog/0c8bba02-a64d-4d91-bae4-65c1fa99d5dc  It's always the same numbers, as you can see here too: https://figgy.princeton.edu/catalog/855efa96-eb8b-4356-aece-e3ee9ba8e465

It's always the same numbers in the files and in the same order. 00000116, 00000056, 00000093, etc.

I noticed it this morning for my last 3 items, which includes C0101_c001964 as well.

May be related to changes in browse everything? Squirrel also notes that they no longer see directories expand when clicking on the checkbox next to them (Squirrel sent a screenshot I can attach if needed of this issue.). Squirrel said this happens on Mac or PC.

hackmastera:
It looks like the files were uploaded 4 times; there are 668 files where there should be 167.

hackmastera:
I've been able to replicate this. If you check the box next to the directory it adds all the files in a non-alphabetic order. If you then open and close the directory it re-adds the files each time.

So currently the only way to add the files you want in the order you need is to open the directory using the little arrow, then check the box next to the directory, and then upload without opening / closing it more. I think this must be related to recent work in browse everything. @jrgriffiniii would you mind giving your opinion when you get a chance?

jrgriffiniii:
I can reenable the directory expansion, but this will require a separate branch for browse-everything.  I can address this.

jrgriffiniii:
This is definitely partly related to the JavaScript for the file-selection widget.  In some cases file entries are not deselected when they are unchecked; I suspect that this might also be involved in the ordering bug.

jrgriffiniii:
https://github.com/pulibrary/figgy/tree/issues-2836-jrgriffiniii-browse-files-duped should resolve this (I've tested this locally, and I'm currently waiting for files being uploaded on staging to finish).  I can issue this as a PR once I've confirmed that it indeed addresses the bugs on staging.

hackmastera:
Here's what I'm seeing in development:
```
[...]

PR #2853, opened 2019-04-09
Resolves issues related to the selection of multiple file entries in the BrowseEverything selection dialogue

jrgriffiniii:
This also ensures that directory file entries are sorted by file name, and as such, advances #2836

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,7 +82,7 @@ gem "rsolr", ">= 1.0"
 
 gem "aasm"
 gem "arabic-letter-connector"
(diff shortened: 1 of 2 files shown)


hackmastera:
Now I get expected behavior when I:
 - just check the directory
 - expand the directory and then check it
 - expand the directory, then check it, then collapse it again
 - expand the directory, then check it, then collapse it, then expand it again again

Unexpected behavior when I:
 - I check the directory, and then expand it (the files are added twice, correct order both times)

hackmastera:
This seems like progress; I'll go ahead and merge

hackmastera:
Oh I forgot this only worked when I removed the methods now overriding upstream methods. I'll issue a follow-up PR.

PR #2857, opened 2019-04-10
Link NumismaticAccessions by reference

### LINKED ISSUE #2856: Link NumismaticAccessions to Coins by reference
- Numismatic accessions are not currently linked to coin resources by reference, they should be.
- Remove the `number_in_accession`  property from Coin. We already have `accession_number`.

eliotjordan:
Closes #2856

### DIFF
--- a/app/resources/coins/coin.rb
+++ b/app/resources/coins/coin.rb
@@ -5,6 +5,7 @@ class Coin < Resource
   include Valkyrie::Resource::AccessControls
   attribute :member_ids, Valkyrie::Types::Array
   attribute :member_of_collection_ids
+  attribute :numismatic_accession_id
 
   # descriptive metadata
   attribute :numismatic_citation, Valkyrie::Types::Array.of(NumismaticCitation).meta(ordered: true)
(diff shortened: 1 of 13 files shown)


app/resources/coins/coin.rb
```diff
@@ -5,11 +5,11 @@ class Coin < Resource
   include Valkyrie::Resource::AccessControls
   attribute :member_ids, Valkyrie::Types::Array
   attribute :member_of_collection_ids
+  attribute :numismatic_accession_id
 
   # descriptive metadata
   attribute :numismatic_citation, Valkyrie::Types::Array.of(NumismaticCitation).meta(ordered: true)
   attribute :coin_number, Valkyrie::Types::Anything
-  attribute :number_in_accession, Valkyrie::Types::Integer
```
escowles: I think we still need this property — my understanding is that items within an accession are numbered, and this lets you say "this coin is item #5 in that accession"

eliotjordan: Ah, right. I mean to remove `accession_number` which was the way that accessions were linked to coins before this PR. I'll switch that around and `number_in_accession` back and get rid of `accession_number`.

eliotjordan:
@escowles Switched out `accession_number ` for `number_in_accession`.

PR #2858, opened 2019-04-10
Deduplicate files before creating PendingUploads

### LINKED ISSUE #2836: Files duplicated when added to a resource

hackmastera:
should be the last piece in order to close #2836

### DIFF
--- a/app/controllers/base_resource_controller.rb
+++ b/app/controllers/base_resource_controller.rb
@@ -46,7 +46,7 @@ class BaseResourceController < ApplicationController
   def selected_files
     files = browse_everything_params.fetch(:selected_files, {})
 
-    files.values.map { |value| BrowseEverything::Resource.new(value) }
+    files.values.uniq.map { |value| BrowseEverything::Resource.new(value) }
   end
 
   # Construct the pending download objects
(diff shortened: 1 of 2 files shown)


app/controllers/base_resource_controller.rb
```diff
@@ -46,7 +46,7 @@ def browse_everything_params
   def selected_files
     files = browse_everything_params.fetch(:selected_files, {})
 
-    files.values.map { |value| BrowseEverything::Resource.new(value) }
+    files.values.map { |value| BrowseEverything::Resource.new(value) }.uniq
```
escowles: Could you do the `uniq` before the `map` and avoid creating the duplicate objects?

hackmastera:
@escowles good call. done!

Issue #2855, opened 2019-04-10
Remove overridden BrowseEverything controller methods and remove the `BrowseEverything::Resource`


jrgriffiniii:
`BulkIngestController#browse_everything_params` and `BulkIngestController#selected_files_params` should be inherited from `BrowseEverythingController` upstream.

tpendragon:
@jrgriffiniii Can you comment on what's necessary to close this ticket now?

PR #2860, opened 2019-04-10
Removes `BrowseEverything::Resource` and `BulkIngestController` browse-everything methods

jrgriffiniii:
Removes BulkIngestController#selected_files_params, BulkIngestController#browse_everything_params, and BulkIngestController#selected_cloud_files? methods, along with BrowseEverything::Resource in favor of requiring these from browse-everything.  Advances #2855


app/controllers/bulk_ingest_controller.rb
```diff
@@ -186,4 +165,10 @@ def workflow_states
     def workflow_class
       @workflow_class ||= DynamicChangeSet.new(resource_class.new).workflow_class
     end
+
+    def process_selected_files_params
+      return unless params[:selected_files]
+
+      params[:browse_everything] = params[:selected_files].fetch(:browse_everything, {})
```
hackmastera: This is used later somewhere else? can you point out where to me?

jrgriffiniii: Sorry, yes, this is related to https://github.com/samvera/browse-everything/compare/master...jrgriffiniii:figgy-issues-2855-jrgriffiniii#diff-5805082baa0b62a6aa50acefa77dcf83R27 and https://github.com/samvera/browse-everything/compare/master...jrgriffiniii:figgy-issues-2855-jrgriffiniii#diff-5805082baa0b62a6aa50acefa77dcf83R12.

`browse-everything` assumes that the request parameter `browse_everything[:selected_files]` contains the values selected by users for upload, and this is indeed valid for the `BaseResourceController`.  However, the `BulkIngestController` assumes that the structure for the parameter is `selected_files[:browse_everything][:selected_files]`: https://github.com/pulibrary/figgy/blob/master/spec/controllers/bulk_ingest_controller_spec.rb#L123

I've just realized that I've forgotten to port this into `BaseResourceController`, I can adjust this and request a second review.

escowles:
@jrgriffiniii One question I still have is whether it makes sense to get your `figgy-issues-2855-jrgriffiniii` branch merged into master.  If you don't think that's realistic in the near term, maybe creating a named branch for tracking what branch we are using in Figgy would be appropriate?

Issue #2830, opened 2019-04-04
Add a switch to make figgy read-only


hackmastera:
To allow for a complete reindex or other semi-downtime requirements

tpendragon:
Maybe related to https://github.com/samvera/valkyrie/issues/270 ?

hackmastera:
The current plan is to see if we can do this via cancancan

PR #2864, opened 2019-04-11
Add READ_ONLY_MODE to Ability

hackmastera:
advances #2830

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,15 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
 
+  before_action :notify_read_only
+
+  def notify_read_only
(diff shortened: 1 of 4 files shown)


eliotjordan:
@hackmastera In general this looks good, but it's a little unclear to me how we set Figgy to read-only mode. Is the thought that we would deploy a branch with the READ_ONLY_MODE set to true?

We could set it with an env variable and add a cap task like this one in OL to do the switch: https://github.com/pulibrary/orangelight/blob/master/config/deploy/staging.rb#L33-L45

hackmastera:
@eliotjordan thanks for the pointer it was just the sort of thing I was fishing for. What do you think of merging this as-is to iterate on? I want to make sure something is in place for this weekend even if it's a bit clunky still.

PR #2871, opened 2019-04-12
Add capistrano switch to read only mode

### LINKED ISSUE #2830: Add a switch to make figgy read-only

hackmastera:
requires provisioning of pulibrary/princeton_ansible#479
fully resolves #2830

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -91,3 +91,42 @@ namespace :deploy do
     end
   end
 end
+
+namespace :env do
+  desc 'Set a Figgy environment variable'
+  task :set do |_task, args|
+    on roles(:all) do
+      abort "Environment variables and values must be specified. `env:set['ENV_VAR=value']`" if args.extras.empty?
+      config_file = '/home/deploy/app_configs/figgy'
+      args.extras.each do |arg|
+        variable, value = arg.split('=', 2)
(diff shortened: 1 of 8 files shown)


config/deploy.rb
```diff
@@ -91,3 +91,42 @@
     end
   end
 end
+
+namespace :env do
+  desc 'Set a Figgy environment variable'
+  task :set do |_task, args|
+    on roles(:all) do
+      abort "Environment variables and values must be specified. `env:set['ENV_VAR=value']`" if args.extras.empty?
+      config_file = '/home/deploy/app_configs/figgy'
```
tpendragon: Do we do this somewhere else? This makes me kind of uncomfortable, since it'll get wiped out on a provision.

hackmastera: Good point.

I pulled it from OL. The code is written to not require this environment variable at all, and it will default to not being in read-only mode. So the potentially bad scenario would be needing to provision during a read-only downtime.

That's not what we're up to this weekend, so maybe it's okay for now?

Also happy to hear other ideas.

tpendragon: For the sake of finishing reindexing, I'm fine with this for now, but I'd like an issue at least to manage the switch in the database somewhere.

hackmastera: here's an issue, feel free to add comments: https://github.com/pulibrary/figgy/issues/2872

PR #2879, opened 2019-04-15
Enables optimistic locking for FileSets

### LINKED ISSUE #2657: intermittent order manager save problems

### LINKED ISSUE #2877: Investigate race condition for Recording Derivative Generation
It seems like sometimes something happens which causes the attached derivatives to be lost.

tpendragon:
This should prevent all race conditions overriding things like derivatives happening in background jobs. It could theoretically cause an error for users too, but we'll have to see if that happens in reality - it should be pretty hard to trigger, since the form doesn't include the lock token.

Tested and made sure this fixes the issue with #2657 and #2877.

Closes #2657 
Closes #2877

### DIFF
--- a/app/adapters/bagit/persister.rb
+++ b/app/adapters/bagit/persister.rb
@@ -44,12 +44,12 @@ module Bagit
         return true if resource.id.blank?
         return true unless resource.optimistic_locking_enabled?
 
-        cached_resource = adapter.query_service.find_by(id: resource.id)
-        return true if cached_resource.blank?
-
         resource_lock_tokens = resource[Valkyrie::Persistence::Attributes::OPTIMISTIC_LOCK]
         resource_value = resource_lock_tokens.find { |lock_token| lock_token.adapter_id == adapter.id }
         return true if resource_value.blank?
+
+        cached_resource = adapter.query_service.find_by(id: resource.id)
+        return true if cached_resource.blank?
 
         cached_value = cached_resource[Valkyrie::Persistence::Attributes::OPTIMISTIC_LOCK].first
         cached_value == resource_value
--- a/app/characterization_services/geo_characterization_service.rb
+++ b/app/characterization_services/geo_characterization_service.rb
@@ -18,11 +18,11 @@ class GeoCharacterizationService
   # @example characterize a file and do not persist the changes
   #   Valkyrie::Derivatives::FileCharacterizationService.for(file_set, persister).characterize(save: false)
   def characterize(save: true)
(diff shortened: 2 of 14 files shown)


app/adapters/bagit/persister.rb
```diff
@@ -53,6 +53,8 @@ def valid_lock?(resource)
 
         cached_value = cached_resource[Valkyrie::Persistence::Attributes::OPTIMISTIC_LOCK].first
         cached_value == resource_value
+      rescue Valkyrie::Persistence::ObjectNotFoundError
```
hackmastera: Is this basically a check for whether the object is persisted? Is there a way to do that in the body of the method instead of via error handling?

eliotjordan:
@tpendragon I tested this with geo material and it seems to work. I did notice that sometimes the FGDC metadata did not extract on a few resources during a bulk ingest, however I couldn't replicate it consistently. Probably an issue unrelated to this PR. My guess is that enabling optimistic locking on Vector and Raster resources would help with this.

app/adapters/bagit/persister.rb
tpendragon: We can't check for persisted (the resource was persisted in a different adapter), but I can check for the lock token first.

tpendragon:
> I am not sure why there are no changes to specs. I think you must have enabled optimistic locking, then made the rest of the changes here due to resulting spec failures?

Yes. I added a quick spec to make sure it's enabled for FileSets.

tpendragon:
@hackmastera I moved this to the `master` branch of `valkyrie-sequel`, if you could re-review.

app/characterization_services/geo_characterization_service.rb
```diff
@@ -35,18 +35,18 @@ def parent
   end
 
   def external_metadata_service
-    @external_metadata_service ||= ExternalMetadataCharacterizationService.new(file_set: file_set, persister: persister)
+    ExternalMetadataCharacterizationService.new(file_set: file_set, persister: persister)
```
jrgriffiniii: Why were these memoized/unmemoized?  If they are only being constructed to have one method call, should `ExternalMetadataCharacterizationService` be refactored to just have a class method in order to avoid the construction of objects?

PR #2880, opened 2019-04-16
update uv-seadragon-extension theme to display fullScreen btn in mobile screen

christinach:
closes https://github.com/pulibrary/orangelight/issues/1630


christinach:
Screen Shot 2019-04-16 at 11 16 54 AM

christinach:
I noticed that the theme.css file is loaded last so any css file I was adding in the uv directory would not overwrite the theme.css. I tried to follow the build.js file for the theme css path but I still could not overwrite it by adding a name-theme file in the themes. The theme which is loading is the uv-seadragon-extension and this is why I update only the theme.css in this directory.

PR #2889, opened 2019-04-18
Turn off search history when in read-only mode

### LINKED ISSUE #2875: read-only mode breaks searching
[Figgy/production] ActiveRecord::StatementInvalid: PG::InsufficientPrivilege: ERROR:  permission denied for relation searches: INSERT INTO "searches" ("query_params", "created_at", "updated_at") VALUES ($1, $2, $3) RETURNING "id"

## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/48219303)

hackmastera:
This allows people to search in read-only mode without getting errors
due to blacklight creating `Search` objects.

fixes #2875

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -21,6 +21,12 @@ class CatalogController < ApplicationController
     }
   end
 
+  # turn off search history during read-only mode
+  def find_or_initialize_search_session_from_params(params)
+    return if Figgy.read_only_mode
+    super
+  end
+
   # enforce hydra access controls
   before_action :set_id, only: :iiif_search
   before_action :enforce_show_permissions, only: [:show, :iiif_search]
--- a/spec/features/catalog_spec.rb
+++ b/spec/features/catalog_spec.rb
@@ -27,4 +27,21 @@ RSpec.feature "Scanned Resources" do
       expect(page).to have_content "New test work"
     end
   end
+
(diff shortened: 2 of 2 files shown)


hackmastera:
Hm. Do not merge it looks like there's another problem here, where the prev / next buttons will still cause errors.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -215,6 +215,24 @@
       get :index, params: { q: "" }
       expect(response).to be_success
     end
+
+    # rubocop:disable RSpec/AnyInstance
+    context "in read-only mode" do
+      let(:emsg) { "PG::InsufficientPrivilege: ERROR:  permission denied for relation searches" }
+      before do
+        allow(Figgy).to receive(:read_only_mode).and_return(true)
+        allow_any_instance_of(Blacklight::SearchContext).to receive(:find_or_initialize_search_session_from_params).and_raise ActiveRecord::StatementInvalid, emsg
```
eliotjordan: Is there a way to do this without `allow_any_instance`, or is there to too much mocking involved? FWIW, I was just poking around blacklight and was astounded by how many levels of nested concerns there are in the controller.

hackmastera:
@eliotjordan I just pushed up something better (deals with the problem closer to the error) which also happens to address your concern!

Issue #2804, opened 2019-04-01
Create a Figgy GraphQL endpoint for events associated with a resource


tpendragon:
The first use case for this is to say "PreservationObject Fixity Verified"
These events should be stored outside the resource, so that updating or creating them don't cause the resource's updated_at to change which may trigger a new bag.

jrgriffiniii:
My understanding of this requirement is that a unidirectional relationship from `Events` linked to both `FileMetadata` nodes and the parent `PreservationObjects` would provide the ability to verify fixity checks at the Bag level and at the level of the individual files.  Is this acceptable, or are their arguments to restrict events just to fixity checks at the level of files?

jrgriffiniii:
https://github.com/pulibrary/figgy/compare/issues-2804-jrgriffiniii-graphql-events introduces a generic `Event` Resource which stores messages and has a relation to modified resources (e. g. `PreservationObjects`).  Further, I've implemented a GraphQL Mutation which permits a client to search for Events by the modified resources, and update the message for any given Events.

Should it be the case that the GraphQL Mutation just creates new Events for PreservationObjects?  If so, should we indefinitely store the Events as a form of an audit trail?

PR #2891, opened 2019-04-19
Implements an Event Resource and a ReportCloudFixity Mutation

jrgriffiniii:
Advances #2804, and please note that this *will* require that UUIDs be generated for FileMetadata IDs before they are persisted.


app/services/preserver.rb
```diff
@@ -53,7 +53,10 @@ def preserve_binary_node(resource_binary_node, file_metadata)
       CleanupFilesJob.perform_later(file_identifiers: file_metadata.file_identifiers.map(&:to_s))
     end
     file_metadata.file_identifiers = uploaded_file.id
-    preservation_object.binary_nodes += [file_metadata] unless file_metadata.persisted?
+    unless file_metadata.persisted?
+      change_set_persister.metadata_adapter.persister.save(resource: file_metadata)
```
tpendragon: I wouldn't do this, I'd just set the ID. I can't believe I forgot to.

app/graphql/mutations/report_cloud_fixity.rb
```diff
@@ -0,0 +1,59 @@
...
+  argument :preservation_object_id, ID, required: true
+  argument :file_metadata_node_id, ID, required: true
+  argument :status, String, required: true
+
+  field :resource, Types::Resource, null: false
+  field :errors, [String], null: true
+
+  def resolve(preservation_object_id, file_metadata_node_id, status)
+    preservation_object = query_service.find_by(id: preservation_object_id)
+    file_metadata_node = query_service.find_by(id: file_metadata_node_id)
```
tpendragon: We don't want to actually save the FileMetadataNodes right? Still leave them as embedded objects?

jrgriffiniii:
Following a meeting earlier today, this needs to be adjusted by having the Mutation removed (Google Cloud Functions are no longer going to asynchronously transmit requests to the GraphQL endpoint exposed for Figgy)

PR #2906, opened 2019-04-22
Google Cloud Fixity Checking

### LINKED ISSUE #2773: Implement Cloud Fixity Auditing

### LINKED ISSUE #2895: Write a Google Cloud Function to perform Fixity Checking
Reports fixity status to a Google Cloud Pub/Sub Topic

tpendragon:
Implements cloud fixity checking via the following workflow:

1. Figgy queues up jobs to be checked in a Google Pub/Sub Topic
2. Cloud function (deployed via `cap staging
   deploy:google_cloud_function`) runs and publishes status to a
   different Google Pub/Sub Topic
3. `rake fixity:run_worker` picks up status jobs on pub/sub and forwards them
   to Sidekiq for processing.

Closes #2895


config/deploy.rb
```diff
@@ -84,6 +84,15 @@
     end
   end
 
+  desc "Deploy Google Cloud Function"
+  task :google_cloud_function do
+    on roles(:db) do
+      run_locally do
+        execute "cd cloud_fixity && gcloud functions deploy checkFixity --runtime nodejs8 --set-env-vars BUCKET=#{fetch(:google_fixity_bucket)},FIXITY_STATUS_TOPIC=#{fetch(:google_fixity_status_topic)} --trigger-topic #{fetch(:google_fixity_request_topic)} --service-account #{fetch(:google_service_account)}"
```
jrgriffiniii: `gcloud` is probably the preferred approach, but I would be curious as to whether or not it is at all possible to integrate something like Grunt or Gulp in order to wrap these types of tasks.

cloud_fixity/index.js
```diff
@@ -0,0 +1,46 @@
+const { Storage } = require('@google-cloud/storage')
+const { PubSub } = require('@google-cloud/pubsub')
+const crypto = require('crypto')
+
+exports.checkFixity = (data, context) => {
+  const attributes = JSON.parse(Buffer.from(data.data, 'base64').toString())
```
jrgriffiniii: Should error handling be introduced here if `data.data` is corrupted or `null`?

config/deploy.rb
tpendragon: I dunno! I'm not too psyched about the idea of adding another tool to our deployment pipeline though.

cloud_fixity/index.js
tpendragon: I think it's pretty reasonable to just let it break. Would that be a problem for a reason I'm not seeing?

jrgriffiniii: I don't see it as a problem, I'm was considering the usage of `console.log` to assist with debugging these types of errors.

app/services/fixity.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def self.pubsub_subscription
+      @pubsub_subscription ||= Figgy.config["fixity_status_topic"]
+    end
+
+    def self.run!
+      Rails.logger.info "Running worker to check for fixity updating"
+
+      topic        = pubsub.topic(pubsub_topic)
```
hackmastera: all these extra spaces make me cry

app/services/fixity.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+module Fixity
```
hackmastera: So the local fixity work happens only in jobs, not services. But it still needs like it would be nice to differentiate between local and cloud stuff in the filename.

Same with the job; maybe `HandleCloudFixityJob` or something?

cloud_fixity/test/index.test.js
```diff
@@ -0,0 +1,61 @@
+// gcloud pubsub topics publish figgy-staging-fixity-request --message '{"preservation_object_id": 1, "file_metadata_node_id": 1, "cloudPath": "example.tif", "md5": "2a28fb702286782b2cbf2ed9a5041ab1"}'
+// gcloud pubsub subscriptions pull --auto-ack figgy-staging-fixity-status
+// gcloud functions deploy checkFixity --runtime nodejs8 --trigger-topic figgy-staging-fixity-request --service-account figgy-staging@pulibrary-figgy-storage-1.iam.gserviceaccount.com
```
hackmastera: Can you expand this? it looks like notes on how to initiate the process but could be fleshed out to be more documentation-y

app/services/fixity.rb
tpendragon: I called it `UpdateFixityJob` assuming we'd spawn the same job for regular fixity checks down the line.

cloud_fixity/test/index.test.js
tpendragon: I'm actually going to delete them. They were just local notes on how to kick things off before the process actually worked.

app/services/fixity.rb
tpendragon: I blame google's example code!

cloud_fixity/test/index.test.js
hackmastera: Okay so then, I'm not sure how everything does get kicked off. I can't find where it pushes to the `figgy-staging-fixity-request` topic. What am I missing?

app/services/fixity.rb
hackmastera: We should make sure this intention is reflected in existing or new tickets

cloud_fixity/test/index.test.js
tpendragon: Oh, yeah, that's a new ticket that exists.

tpendragon: Created #2919

PR #2914, opened 2019-04-23
Implements a Preserver::Importer Class for importing resources from file identifiers

### LINKED ISSUE #2894: Add method of importing from preservation storage

jrgriffiniii:
Resolves #2894

### DIFF
--- a/app/services/preserver/importer.rb
+++ b/app/services/preserver/importer.rb
@@ -37,13 +37,12 @@ class Preserver
     end
 
     def import!
-      fs_attributes = import_metadata(metadata_file_identifier)
+      fs = build_file_set(metadata_file_identifier)
+      fs_change_set = FileSetChangeSet.new(fs)
 
       files = import_binary_nodes(binary_file_identifiers)
-      fs_attributes[:files] = files unless files.empty?
-      fs = FileSet.new
-      fs_change_set = FileSetChangeSet.new(fs)
-      fs_change_set.validate(fs_attributes)
+      fs_change_set.validate(files: files)
+
       persisted = nil
       change_set_persister.buffer_into_index do |buffered_change_set_persister|
         persisted = buffered_change_set_persister.save(change_set: fs_change_set)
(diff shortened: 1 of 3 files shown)


jrgriffiniii:
This accepts `Valkyrie::IDs` from the `FileMetadata` nodes normally stored in `PreservationObjects`, with the IDs being valid `file_identifiers` which are resolved using a `StorageAdapter`.

Is this the desired design?  If not, I can adjust this to import from a URL to Google Cloud resources (or some other value).

tpendragon:
> Is this the desired design? If not, I can adjust this to import from a URL to Google Cloud resources (or some other value).

I think this is probably what we want.

escowles:
@jrgriffiniii This looks good to me, but needs to be rebased.

app/services/preserver/importer.rb
```diff
@@ -0,0 +1,121 @@
...
+        fs_attributes
+      end
+
+      def import_metadata(file_identifier)
+        return {} if file_identifier.nil?
+        file = storage_adapter.find_by(id: file_identifier)
+
+        file_contents = file.read
+        metadata_json = JSON.parse(file_contents)
+        build_file_set_attributes(metadata_json)
```
tpendragon: I think you can use the conversion routines which are a part of Valkyrie. I can't really dig into this right now, but instead of manually parsing the metadata maybe you can look at using https://github.com/samvera/valkyrie/blob/2-0-rc/lib/valkyrie/persistence/shared/json_value_mapper.rb when 2.0 is released? Or even just pass it through the valkyrie-sequel ResourceFactory?

app/services/preserver/importer.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+
+      def build_file_set(file_identifier)
+        return {} if file_identifier.nil?
+        file = storage_adapter.find_by(id: file_identifier)
+
+        file_contents = file.read
+        metadata_json = JSON.parse(file_contents)
+        resource_object = { metadata: metadata_json }
+        Valkyrie.config.metadata_adapter.resource_factory.to_resource(object: resource_object)
```
jrgriffiniii: This delegates to `Valkyrie::Sequel::ResourceFactory#to_resource`, but this does not appear to handle optimistic lock tokens without preprocessing the values.

app/services/preserver/importer.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      def build_file_set(file_identifier)
+        return resource_class.new if file_identifier.nil?
+        metadata_file = storage_adapter.find_by(id: file_identifier)
+
+        metadata_file_contents = metadata_file.read
+        metadata_json = JSON.parse(metadata_file_contents)
+        metadata_json.delete("file_metadata")
+        resource_object = { metadata: metadata_json }
+        file_set = Valkyrie.config.metadata_adapter.resource_factory.to_resource(object: resource_object)
```
jrgriffiniii: While `Valkyrie.config.metadata_adapter.resource_factory.to_resource` from `valkyrie-sequel` (https://github.com/samvera-labs/valkyrie-sequel/blob/master/lib/valkyrie/sequel/resource_factory.rb#L12) does effectively construct the object, there was still a fair amount of processing which I needed to introduce for the metadata both before and after the `FileSet` was constructed.

PR #2921, opened 2019-04-24
small update in the drop down select box

christinach:
Small update in the drop down select. 
I noticed that changing into the same px number the max-height in open and inner class, some information (select options) might not be visible. Now I'm setting the open max-height > than the inner max-height to fix this.


christinach:
can someone check this it's a small fix on a previous commit.

Issue #2918, opened 2019-04-24
"Translation missing: en.models.letter"


escowles:
In the modal to select from all available models to limit a search by work type, I see "translation missing: en.models.letter"

![Screen Shot 2019-04-24 at 8 39 58 AM](https://user-images.githubusercontent.com/856924/56660102-bf177f00-666c-11e9-9cc2-18a8b27bcfbd.png)

PR #2923, opened 2019-04-24
updates locales/en.yml to include Letter in Type of Work

christinach:


christinach:
closes #2918

christinach:
I can see the change locally but not in staging.

escowles:
I'm seeing the error still too — I even restarted Passenger on figgy-staging1 and tried a different browser, with no effect.

eliotjordan:
This works. The resources just needed to be reindexed.

PR #2961, opened 2019-05-01
Updated members_numismatic_issue UI with data-tables

christinach:


christinach:
before: 
Screen Shot 2019-05-01 at 3 13 03 PM

christinach:
with the data-tables
Screen Shot 2019-05-01 at 3 14 12 PM
 update:

christinach:
I added a custom width in the thumbnail column so that if there is no thumbnail image the column will not expand based on the coin url content. This way the 'Actions' column will fit both buttons 'view' and 'edit'.

christinach:
I also removed the panel. I think the data-tables view looks cleaner without a panel.

christinach:
closes #2934

eliotjordan:
This looks great. We should do this with all of our child resource partials.

https://github.com/pulibrary/figgy/issues/2962

PR #2980, opened 2019-05-07
Using Devise::Controllers::StoreLocation

### LINKED ISSUE #2395: Redirecting to previous page after login doesn't always work
When I try to access a private resource (either because of visibility or pending state) when I am not logged in, I get redirected to the home page after logging in with CAS.  I should be redirected back to the private resource after logging in.

Refs #2390

escowles:
* Stores user locations on appropriate GET requests
* Uses stored location to redirect back after login

Fixes #2395


eliotjordan:
Tested and it works. With a small addition, I think it might solve the geo download problem.

PR #2986, opened 2019-05-07
Adds Numismatics UI

### LINKED ISSUE #2929: Home screen for numismatics
Given the number of numismatic resources that can be managed and created, it might be useful to have a home screen just for numismatics.

christinach:
1. Updates the home-screen with the numismatics home url
2. Adds the numismatics home UI
3. The numismatics UI includes -> button for each managed numismatic resource.
4. Updates drop-down _add_content
5. Adds method for querying and counting the managed numismatics in the wayfinders

closes #2929


eliotjordan:
While the datatable and statistics are nice, the home page takes so long to load in staging with real data, that NGNIX times out with a "504 Gateway Time-out" message.

christinach:
@escowles when you have a chance can you check this please?

eliotjordan:
@christinach I just realized that the panels were different sizes for me because I had my window width reduced. Still, I think it would look better it the table name and counts were on seperate rows.

christinach:
1. The panels have the same height. Where is the height difference?

eliotjordan:
@christinach 

Screen Shot 2019-05-10 at 2 58 59 PM

christinach:
This is because based on the bootstrap grid you are in the pixels where the change happens for the responsive.

christinach:
when you say table counts and names in a different row you mean to display something like that so that they don't wrap:
Screen Shot 2019-05-10 at 4 03 11 PM

eliotjordan:
I'm not sure it's worth spending too much time on. There's a narrow range of window widths where this is a problem. Other concerns:

1. I wonder if three panels across would be better? I feel a little overwhelmed with six panels.
1. It would look more pleasing if the buttons were centered in the panel like they are on the Figgy home page.
1. Is `/concern/numismatic_issues/show` the best path for this? This isn't just for Issues, but for all Numismatics resources. I was thinking something simple like `/numismatics`. Definitely open to other opinions on this.
1. I'm seeing the "Manage Numismatics" button and am able to see the Numismatics home screen and even without any permissions.

Screen Shot 2019-05-10 at 3 14 09 PM

escowles:
I agree that `/numismatics` would be a better URL for the dashboard

christinach:
Screen Shot 2019-05-10 at 4 22 35 PM
@eliotjordan I cannot see the numismatics button when I am not logged in.

eliotjordan:
@christinach  Login and then remove yourself as an administrator.

christinach:
1. I'll update the url
2. We can check the 3 panels. 
3. ok, I see thanks I think it's because I check for create permission to display it. 
That brings another question I had; who should be able to view this UI information ?

eliotjordan:
We usually use guards in views like this

```
<% if can?(:create, ::NumismaticIssue) %>
```

christinach:
@eliotjordan I think you are looking a different commit. because I do have this

escowles:
For the permissions, I think it should be visible to `staff` — checking if a user can create a NumismaticIssue is a reasonable way to check that.  I just checked the `Ability` class and none of the Numismatic classes are mentioned directly, so I think that's a bigger issue than this PR.

christinach:
the button in figy-staging should say Manage Numismatics

christinach:
Screen Shot 2019-05-10 at 4 29 46 PM

christinach:
also when I removed my self from admin:
Screen Shot 2019-05-10 at 4 32 08 PM

app/views/catalog/_numismatics_home_panel.html.erb
```diff
@@ -0,0 +1,12 @@
+<% if can?(:create, NumismaticIssue) %>
```
christinach: @eliotjordan

christinach:
https://github.com/pulibrary/figgy/pull/2986/files#diff-8487b4a4288da3e5e658657a20e625bc

app/views/catalog/_numismatics_home_panel.html.erb
eliotjordan: Oh I see. My local copy is not up to date. Let me pull in the latest changes.

eliotjordan:
I see that guard now in the view, that was my mistake. However, the home screen is still accessible without read/write permissions. Will probably need an authorize statement in the controller method:

```
authorize! :read, NumismaticIssue
```

Like the one here:
https://github.com/pulibrary/figgy/blob/master/app/controllers/concerns/resource_controller.rb#L13

christinach:
I can update the ability for the numismatic classes in this PR as well.

eliotjordan:
@christinach Thanks! That would be great. Thanks for all your hard work on this. This is going to work really well when it's finished; as well as set a pattern for other resource types. Like the DataTable you added for Coin children on the Issue show page, grouping together stats, reports, and buttons for creating or managing new resources will also be useful for geospatial data.

christinach:
@eliotjordan thanks for reviewing!

christinach:
Screen Shot 2019-05-22 at 12 01 31 PM

christinach:
Screen Shot 2019-05-22 at 12 14 35 PM

app/views/catalog/numismatics.html.erb
```diff
@@ -0,0 +1,48 @@
+<h3 class="title">Numismatics</h3>
+<div class="row">
+  <div class="col-md-12">
+    <div class="panel-body">
+      <% @numismatic_decorator.metric_labels_first.each_with_index do |name, index| %>
```
eliotjordan: Rather than add this label to the NumismaticIssue decorator, it would probably be better to just enumerate them here:

```
<% ["Accessions", "Firms", "Monograms"].each_with_index do |name, index| %>
```

Or, even better, create a partial to reduce some of the code repetition. Like we do in in the `home_text` view:
https://github.com/pulibrary/figgy/blob/master/app/views/catalog/_home_text.html.erb#L10-L12

app/views/catalog/numismatics.html.erb
```diff
@@ -0,0 +1,48 @@
...
+          </li>
+        </ul>
+      <% end %>
+    </div>
+  </div>
+</div>
+<div class="row">
+  <div class="col-md-12">
+    <div class="panel-body">
+      <% @numismatic_decorator.metric_labels_second.each_with_index do |name, index| %>
```
eliotjordan: Same as above.

app/resources/numismatic_issues/numismatic_issue_decorator.rb
```diff
@@ -135,4 +135,16 @@ def state
...
+    ["Accessions", "Firms", "Monograms"]
+  end
+
+  def metric_labels_second
+    ["People", "Places", "References"]
+  end
+
+  def metric_issues_label
+    ["Issues"]
+  end
```
eliotjordan: See the comments in `numismatics.html.erb`. I don't think the NumismaticIssueDecorator is the right place for these.

app/controllers/catalog_controller.rb
```diff
@@ -174,4 +175,19 @@ def lookup_manifest
       render json: { message: "No manifest found for #{ark}" }, status: 404
     end
   end
+
+  def numismatics
```
eliotjordan: Probable better to move this to it's own controller. The FixityDashboardController is an example of this:
https://github.com/pulibrary/figgy/blob/master/app/controllers/fixity_dashboard_controller.rb

christinach: I suggested yesterday to move it to its own controller when we will refactor numismatics.

app/views/catalog/numismatics.html.erb
christinach: I don't want to have the array in the view this is why I added in the decorator since we use decorators for the views.

christinach: I understand that is a method specific to the NumismaticIssue decorator and not to the NumismaticDecorator since we have not refactored them yet and we dont have a NumismaticDecorator but isn't it better to be outside of the view?

christinach: I will add the partials.

app/controllers/catalog_controller.rb
christinach: the requested change was to be a https://figgy.princeton.edu/numismatics. 
If I move it to a different controller would that be:
https://figgy.princeton.edu/numismatics/home ?
https://figgy.princeton.edu/dashboard/numismatics?
or something else?

christinach: we decided dashboard/numismatics.

app/views/catalog/numismatics.html.erb
christinach: I will also keep the strings in the view as it was suggested from @eliotjordan and @tpendragon

PR #3022, opened 2019-05-14
Fix structure for Audio Files

### LINKED ISSUE #2905: Playlist structure orders unnested ranges wrong
Can be seen in a version of https://figgy.princeton.edu/catalog/128f34e9-af1f-4e65-977b-303c573b41d9.

Basically if you have "A -> [[B -> C], D]", then D will show up before B and C in the viewer.

tpendragon:
Closes #2905

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -111,17 +111,28 @@ class ManifestBuilder
       return default_audio_ranges if logical_structure.blank? || logical_structure.flat_map(&:nodes).blank?
       logical_structure.flat_map do |top_structure|
         top_structure.nodes.map do |node|
-          if node.proxy.present?
-            TopStructure.new(
-              Structure.new(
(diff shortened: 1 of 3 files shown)


hackmastera:
@tpendragon can you say more about what was broken / what fixed it?

tpendragon:
@hackmastera The problem is that when audio files had a structure of
```
- Act 1
  - Act 1a
    - Track 1
    - Track 2
  - Track 3
```

it would show up as 

```
- Act 1
  - Track 3
  - Act 1a
    - Track 1
    - Track 2
```

The reason it did that was because in order for the track names to show up in the manifest we had to build a fake range with the title of the track. Before we did that in the StructureBuilder, but the nodes that get passed to the StructureBuilder separates the ranges it has from the child nodes. Rather than doing that, this PR just wraps the structure up with parent nodes for children before passing to the builder.

PR #3029, opened 2019-05-15
Handles multiple URLs for holding location values

### LINKED ISSUE #3028: Improve error handling for controlled vocabulary term dropdown fields in the resource edit view
## Backtrace

    line 327 of [PROJECT_ROOT]/app/models/controlled_vocabulary.rb: []
    line 327 of [PROJECT_ROOT]/app/models/controlled_vocabulary.rb: block in all
    line 326 of [PROJECT_ROOT]/app/models/controlled_vocabulary.rb: each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/49507284)

jrgriffiniii:
Resolves #3028

### DIFF
--- a/spec/resources/ephemera_folders/ephemera_folder_decorator_spec.rb
+++ b/spec/resources/ephemera_folders/ephemera_folder_decorator_spec.rb
@@ -245,6 +245,51 @@ RSpec.describe EphemeraFolderDecorator do
     it "renders the location label" do
       expect(decorator.rendered_holding_location).to eq(["Firestone Library"])
     end
+
+    context "when a holding location has multiple URLs" do
+      let(:holding_locations) do
+        [
+          {
+            label: "Firestone Library",
+            url: [
+              "https://bibdata.princeton.edu/locations/delivery_locations/15.json",
+              "https://bibdata.princeton.edu/locations/delivery_locations/16.json"
+            ]
+          }
+        ]
+      end
+
+      before do
+        allow(MultiJson).to receive(:load).and_return(holding_locations)
+      end
+
+      it "renders the location label for the selected location" do
+        expect(decorator.rendered_holding_location).to eq(["Firestone Library"])
+      end
+
+      after do
+        allow(MultiJson).to receive(:load).and_call_original
+      end
+    end
+
+    context "when a holding location has no URLs" do
+      let(:holding_locations) do
+        [
+          {
+            label: "Firestone Library",
+            url: []
+          }
+        ]
+      end
+
+      before do
+        allow(MultiJson).to receive(:load).and_return(holding_locations)
+      end
+
+      it "does not render the location" do
+        expect(decorator.rendered_holding_location).to eq([])
+      end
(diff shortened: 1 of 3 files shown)


spec/resources/ephemera_folders/ephemera_folder_decorator_spec.rb
```diff
@@ -245,6 +245,52 @@
...
+            url: []
+          }
+        ]
+      end
+
+      before do
+        allow(MultiJson).to receive(:load).and_return(holding_locations)
+      end
+
+      it "renders the location label for the selected location" do
```
hackmastera: It seems like it renders nothing? Can you update the spec description?

spec/resources/ephemera_folders/ephemera_folder_decorator_spec.rb
```diff
@@ -245,6 +245,52 @@
...
+        expect(decorator.rendered_holding_location).to eq([])
+      end
+
+      after do
+        allow(MultiJson).to receive(:load).and_call_original
```
hackmastera: I don't think I've ever reset a mock like this in an `after`... is that something we need to be doing??

spec/resources/ephemera_folders/ephemera_folder_decorator_spec.rb
```diff
@@ -245,6 +245,52 @@
     it "renders the location label" do
       expect(decorator.rendered_holding_location).to eq(["Firestone Library"])
     end
+
+    context "when a holding location has multiple URLs" do
+      let(:holding_locations) do
+        [
+          {
+            label: "Firestone Library",
+            url: ["https://bibdata.princeton.edu/locations/delivery_locations/15.json"]
```
hackmastera: This looks like only one url. But the context says there are multiple. Should we have a 2nd url? Or are we just testing that it can handle an array?

spec/resources/ephemera_folders/ephemera_folder_decorator_spec.rb
jrgriffiniii: Because I was mocking a Class method, I wanted to be certain that this was explicitly disabled before the next test.

hackmastera: I don't think I've ever done this and I'm pretty sure I've mocked class methods...now I'm worried that the tests are all infected. Do you know of any documentation that this is required?

jrgriffiniii: There wasn't any documentation, I was being cautious being that `MultiJson` having class methods invoked within a different context, and thinking of patterns such as this for the global namespace: https://github.com/samvera/hyrax/blob/44a12d4cb1e96c5c80a489864cdb1acf96d1090e/spec/models/hyrax/work_behavior_spec.rb#L8

I have no strong feelings about keeping this in the PR, hence I will remove it.

PR #3032, opened 2019-05-15
Display in the viewer only the coins that have an image

christinach:
Prevents a sequence 0 error in the UV viewer by building the manifest with only the coins that have an image

Co-authored-by: Trey Pendragon <tpendragon@princeton.edu>

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -28,6 +28,8 @@ class ManifestBuilder
         IndexCollectionNode.new(resource)
       when ScannedMap
         ScannedMapNode.new(resource)
+      when NumismaticIssue
+        NumismaticIssueNode.new(resource)
       when Playlist
         PlaylistNode.new(resource, auth_token)
       else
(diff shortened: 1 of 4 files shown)


christinach:
closes #3026

christinach:
with this change: if a numismatic issue has a list of 3 coins with images and the user by mistake or intentionally removes or don't add an image in the coin that was first created, the UV will not break. This is because the issue manifest is considered a @type sc:Collection. With that said the attached members have a specific sequence and the collection will always read in the manifest sequences key starting from the first member. If the first member misses an image the viewer will break. This is why we exclude it from the manifest. It will still be a numismatic_issue member and listed in the numismatic_issue screen. check: https://figgy-staging.princeton.edu/catalog/3d8e675b-e4a3-4680-81a8-d821914b0126 `coin 13393` was first attached in the issue.

christinach:
(update in the previous comment )before the change: If any other member (not the first attached) misses an image  the viewer will not break but if the user selects in the viewer the coin that misses an image it will show that it's loading without displaying anything. 
with this change: we exclude from the manifest the coins that don't have an image.

app/services/manifest_builder.rb
```diff
@@ -323,6 +325,12 @@ def geo_image?(member)
       end
   end
 
+  class NumismaticIssueNode < CollectionNode
+    def members
+      @members ||= super.select { |x| x.member_ids.present? }
```
hackmastera: Could you replace `x` with a more specific variable name? Will a member of an issue always be a coin? If so maybe use `coin`. Also, it would be nice to have a comment here, like "Only include coins which have file sets; otherwise there's no image for a viewer" or something.

christinach: @hackmastera I updated.

PR #3039, opened 2019-05-16
Make show-page action buttons sticky

### LINKED ISSUE #3009: Sticky action buttons on resource show page
here's the wireframe users voted for during user testing https://docs.google.com/drawings/d/1w0YKSo8s2pz3CFyfiJdCvw5OFCrON4CfN1sdnttK_s0/edit

hackmastera:
fixes #3009


tpendragon:
I like the sticky, I wonder if we should update the background color? The button bar's really easy to ignore.

escowles:
👍 on both points — the stickiness works perfectly, but having the bar be the same color as the shaded table rows makes it kind of disappear.

hackmastera:
Okay here are some colors I found lying around.

First is the original color, for comparison. Pulled from the form heading bars:
Screen Shot 2019-05-16 at 4 40 56 PM

This one is the header / footer background color. I think it is actually different enough to make the button bar pop a tiny bit, maybe just enough:
Screen Shot 2019-05-16 at 4 38 48 PM

Then here's a gray I found in `colors.scss`, cooler in tone than the header / footer to match the color temp of the form but darker to make the button bar stand out:
Screen Shot 2019-05-16 at 4 37 23 PM

Finally, something I found in `colors.scss` called `gainsboro` which I don't know what that is. it's both darker and warmer:
Screen Shot 2019-05-16 at 4 37 04 PM

I also tried the info bar color but i thought it was way too much. but if y'all want some screen shots with dramatically poppier colors i could go that direction.

hackmastera:
I'll also mention I brought the sticky bar all the way down to the bottom of the screen because seeing a tiny bit of content scrolling below the sticky bar was weirding me out / seemed unnecessarily "cool".

tpendragon:
I like the second to last one, the gray you found.

christinach:
I like the last 2. gray and gainsboro.

christinach:
maybe you can check the `axe` tool for accessibility or the`lighthouse` tool in the chrome developer tools -> audits section

christinach:
It will have a result something like 'Elements must have sufficient color contrast' (if there is no sufficient color contrast).

hackmastera:
I've also just noticed it's not playing well with some of the maps stuff z-index-wise.

hackmastera:
@christinach I'm not sure why we need contrast analysis? I didn't change anything that has text against it.

christinach:
It was a suggestion to help decide between the colors but If it only checks the contrast with the text then maybe it's not necessary.

hackmastera:
> The bar is not sticky in IE & Safari. It works for Firefox and Chrome.

Ah, thank you for checking those!

hackmastera:
weird. it is sticky in safari but there are no buttons in it!

tpendragon:
http://dinbror.dk/blog/help-css-position-sticky-doesnt-work-solved/ relevant maybe?

christinach:
I can see the buttons and the sticky in safari. I'll check again browserstack

christinach:
Safari, Chrome and Firefox look ok. (earlier I checked a very old safari version )
IE doesn't like the sticky.

christinach:
https://caniuse.com/#search=position%3Asticky IE has a `red`

tpendragon:
@christinach Does the bar totally go away in IE, such that it's unusable?

christinach:
The bar looks good. It's only at the bottom

christinach:
stickyIE

tpendragon:
I'm for just merging this then.

christinach:
ok!

Issue #2539, opened 2019-02-06
If a derivative fails to generate, it should throw an error.


tpendragon:
Right now it just continued silently, which made the job not retry.

escowles:
To replicate: break ImageMagick and try to generate derivatives.

jrgriffiniii:
I could not find a way to disable this on the CircleCI environment (uninstalling ImageMagick using Homebrew on my local development environment *did* trigger the failures needed to reproduce this), hence I am going to proceed with testing for this by just raising errors within the context of RSpec tests.

jrgriffiniii:
I'm uncertain if retriggering the job explicitly in this manner would be acceptable: https://github.com/pulibrary/figgy/compare/issues-2539-jrgriffiniii-derivative-failures#diff-dde2fa5f290a1e9acf995cfc327145afR15

hackmastera:
@jrgriffiniii hm. it seems like the ticket is asking for the error to make it all the way out so that sidekiq will re-run the job. Is there something up the stack that is stifling the error?

jrgriffiniii:
`MiniMagick::Error: ImageMagick/GraphicsMagick is not installed` is still being propagated up through the stack for `CreateDerivativesJob` (I assume that this is how the job restarts are triggered), I'll look to improve the work on the branch in order to ensure that nothing should be getting stifled.

jrgriffiniii:
This seems to be addressed by the changes proposed in https://github.com/pulibrary/figgy/pull/3040, with the understanding that this is just a different message for `MiniMagick::Error`.  I will proceed by adding this as an additional test case and closing https://github.com/pulibrary/figgy/pull/3048

PR #3040, opened 2019-05-16
Ensures that ImageMagick errors for ScannedMap derivative generation are logged and that derivatives are cleaned up when these errors occur

### LINKED ISSUE #2978: Problems with ScannedMap derivatives
When a ScannedMap thumbnail derivative fails, the derivative job restarts, which then generates a new JP2 derivative. This cycle repeats. To fix this, the scanned map derivative service should rescue errors and cleanup derivatives.

jrgriffiniii:
Resolves #2978 and #2539

### DIFF
--- a/app/derivative_services/scanned_map_derivative_service.rb
+++ b/app/derivative_services/scanned_map_derivative_service.rb
@@ -42,7 +42,24 @@ class ScannedMapDerivativeService
 
   def create_derivatives
     jp2_derivative_service.create_derivatives if jp2_derivative_service.valid?
-    thumbnail_derivative_service.create_derivatives if thumbnail_derivative_service.valid?
+    begin
+      thumbnail_derivative_service.create_derivatives if thumbnail_derivative_service.valid?
+    rescue StandardError => error
+      # Delete the derivative files
+      derivative_files = resource.file_metadata.select(&:derivative?)
+      derivative_files.each do |derivative_file|
+        derivative_file.file_identifiers.each do |file_id|
+          change_set_persister.storage_adapter.delete(id: file_id)
+        end
+      end
+
+      original_files = resource.file_metadata.reject(&:derivative?)
+      # Delete the metadata for the derivative files
+      resource.file_metadata = original_files
+      change_set_persister.persister.save(resource: resource)
+
+      raise error
+    end
   end
 
   # Removes Valkyrie::StorageAdapter::File member Objects for any given Resource (usually a FileSet)
--- a/app/derivative_services/scanned_map_derivative_service.rb
+++ b/app/derivative_services/scanned_map_derivative_service.rb
@@ -43,6 +43,9 @@ class ScannedMapDerivativeService
   def create_derivatives
     jp2_derivative_service.create_derivatives if jp2_derivative_service.valid?
     thumbnail_derivative_service.create_derivatives if thumbnail_derivative_service.valid?
+  rescue MiniMagick::Error => mini_magick_error
+    Rails.logger.error "Failed to generate JP2000 derivatives for #{id}: #{mini_magick_error.message}"
+    cleanup_derivatives
   end
 
   # Removes Valkyrie::StorageAdapter::File member Objects for any given Resource (usually a FileSet)
--- a/app/derivative_services/scanned_map_derivative_service.rb
+++ b/app/derivative_services/scanned_map_derivative_service.rb
@@ -43,10 +43,6 @@ class ScannedMapDerivativeService
   def create_derivatives
     jp2_derivative_service.create_derivatives if jp2_derivative_service.valid?
     thumbnail_derivative_service.create_derivatives if thumbnail_derivative_service.valid?
-  rescue MiniMagick::Error => mini_magick_error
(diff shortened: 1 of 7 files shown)


app/derivative_services/scanned_map_derivative_service.rb
```diff
@@ -42,6 +42,9 @@ def parent
   def create_derivatives
     jp2_derivative_service.create_derivatives if jp2_derivative_service.valid?
     thumbnail_derivative_service.create_derivatives if thumbnail_derivative_service.valid?
+  rescue MiniMagick::Error => mini_magick_error
+    Rails.logger.error "Failed to generate JP2000 derivatives for #{id}: #{mini_magick_error.message}"
```
tpendragon: This will prevent the job from restarting, which means if it fails it will always be broken until we manually kick it. Is that what we want?

jrgriffiniii: Sorry, should the errors *not* be rescued then, and this should propagate the error?  I can still clean up the derivatives and ensure that the error message is logged.

tpendragon: Maybe we should just do this for all derivative errors? If a StaleObjectError happens we could end up with a duplicate too.

app/jobs/create_derivatives_job.rb
```diff
@@ -4,14 +4,19 @@ class CreateDerivativesJob < ApplicationJob
...
     metadata_adapter.persister.save(resource: file_set)
     messenger.derivatives_created(file_set)
     CheckFixityJob.perform_later(file_set_id)
-  rescue Valkyrie::Persistence::ObjectNotFoundError => error
-    Valkyrie.logger.warn "#{self.class}: #{error}: Failed to find the resource #{file_set_id}"
+  rescue Valkyrie::Persistence::ObjectNotFoundError => not_found_error
+    Valkyrie.logger.warn "#{self.class}: #{not_found_error}: Failed to find the resource #{file_set_id}"
+  rescue StandardError => error
+    Valkyrie.logger.error "Failed to generate derivatives for #{file_set_id}: #{error.class}: #{error.message}"
+    derivative_service.cleanup_derivatives
```
jrgriffiniii: This raises a `StaleObjectError` under certain conditions.

tpendragon:
So I've been playing with this. I don't think the tests work - if I remove the stubs and use real stuff I can't get the actual test to fail if I remove the implementation. Further, I'm not sure if this is what we want necessarily. I think what's going on is the ScannedMapDerivativeService runs two derivative services, the first one (the JP2) is succeeding and the second one (the thumbnail) is failing, so every time it re-runs it's recreating the JP2. Do we want to roll back that JP2, or just have it not run the JP2 derivative service if it exists already? I dunno.

app/derivative_services/scanned_map_derivative_service.rb
```diff
@@ -42,7 +42,19 @@ def parent
...
+    begin
+      thumbnail_derivative_service.create_derivatives if thumbnail_derivative_service.valid?
+    rescue StandardError => error
+      # failure_change_set = FileSetChangeSet.new(resource)
+      original_files = resource.file_metadata.reject(&:derivative?)
+      # failure_change_set.validate(file_metadata: original_files)
+      # change_set_persister.save(change_set: failure_change_set)
+
+      # Roll back the JP2000
+      resource.file_metadata = original_files
```
jrgriffiniii: This should delete the metadata for the JP2 derivative when the thumbnail generation fails, but I suspect that the file itself might also need to be explicitly deleted using the storage adapter.

app/jobs/create_derivatives_job.rb
```diff
@@ -4,14 +4,21 @@ class CreateDerivativesJob < ApplicationJob
...
     file_set = query_service.find_by(id: Valkyrie::ID.new(file_set_id))
     file_set.processing_status = "processed"
     metadata_adapter.persister.save(resource: file_set)
     messenger.derivatives_created(file_set)
     CheckFixityJob.perform_later(file_set_id)
-  rescue Valkyrie::Persistence::ObjectNotFoundError => error
-    Valkyrie.logger.warn "#{self.class}: #{error}: Failed to find the resource #{file_set_id}"
+  rescue Valkyrie::Persistence::ObjectNotFoundError => not_found_error
+    Valkyrie.logger.warn "#{self.class}: #{not_found_error}: Failed to find the resource #{file_set_id}"
+  rescue StandardError => error
```
jrgriffiniii: Given https://github.com/pulibrary/figgy/pull/3040#issuecomment-502749459, should this proposed change be removed?  When I remove this but keep https://github.com/pulibrary/figgy/pull/3040/files#diff-575692548a61b37094b8e1c882bca938R47 in place, the tests do fail.

tpendragon:
I just wanted to say I'm sorry for the delay in re-reviewing this. I keep starting and getting lost undoing stubs to make sure the tests work with real data.

jrgriffiniii:
No apologies are needed, I am sorry that I have lost track with this (as well as https://github.com/pulibrary/figgy/pull/3290).

Should I close this for now?  I doubt that I'll have the opportunity to revisit this for at least one week.

tpendragon:
No, please keep it open.

PR #3042, opened 2019-05-16
Keep Track of and display status of FileSet derivative creation

### LINKED ISSUE #2447: Improve User Feedback on Ingest Status
Lots of questions on All-Hands 1/19 about when things are done, if they're done, and background processing that might be happening.

tpendragon:
After browse everything attached, but files not moved to repository:
Screen Shot 2019-05-16 at 3 03 40 PM

Before derivatives done being created:

Screen Shot 2019-05-16 at 3 04 38 PM

Wanted to add a note: I benchmarked this query for production. On a MVW with 1300 total filesets it takes ~ 70 ms to count them.

Closes #2447

### DIFF
--- a/app/views/catalog/_in_process_notification_default.html.erb
+++ b/app/views/catalog/_in_process_notification_default.html.erb
@@ -1,0 +1,6 @@
+<% if @document.wayfinder.try(:in_process_file_sets_count).to_i > 0 %>
+  <div class="alert alert-warning" role="alert">Files currently processing: <%= @document.wayfinder.in_process_file_sets_count %></div>
+<% end %>
+<% if @document.resource.try(:pending_uploads).present? %>
+  <div class="alert alert-warning" role="alert">Files currently attaching: <%= @document.resource.pending_uploads.length %></div>
+<% end %>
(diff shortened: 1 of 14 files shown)


tpendragon:
As per @roelmoon's feedback, it now provides a running tally:
Screen Shot 2019-05-17 at 9 34 46 AM

app/views/catalog/_in_process_notification_default.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if @document.wayfinder.try(:in_process_file_sets_count).to_i > 0 %>
+  <div class="alert alert-warning" role="alert">Files Processed: <%= @document.wayfinder.processed_file_sets_count %> / <%= @document.wayfinder.processed_file_sets_count + @document.wayfinder.in_process_file_sets_count %></div>
+<% end %>
+<% if @document.resource.try(:pending_uploads).present? %>
+  <div class="alert alert-warning" role="alert">Files currently attaching: <%= @document.resource.pending_uploads.length %></div>
```
hackmastera: We should align this heading with the one on the file manager page. I think users will wonder what is the difference between a file attaching and a pending upload.

PR #3043, opened 2019-05-17
after deleting a fileset redirect to the parent

### LINKED ISSUE #3013: File manager -> Deleting a File Set should redirect to the parent

christinach:
closes #3013

### DIFF
--- a/app/resources/file_sets/file_sets_controller.rb
+++ b/app/resources/file_sets/file_sets_controller.rb
@@ -8,9 +8,9 @@ class FileSetsController < ApplicationController
     metadata_adapter: Valkyrie::MetadataAdapter.find(:indexing_persister),
     storage_adapter: Valkyrie::StorageAdapter.find(:disk)
   )
+  before_action :parent_resource, only: [:destroy]
 
   def derivatives
-    file_set = find_resource(params[:id])
     @change_set = change_set_class.new(file_set).prepopulate!
     authorize! :derive, @change_set.resource
     output = RegenerateDerivativesJob.perform_later(params[:id])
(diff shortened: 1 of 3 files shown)


app/resources/file_sets/file_sets_controller.rb
```diff
@@ -53,6 +54,14 @@ def update
     after_update_error error
   end
 
+  def after_delete_success
+    redirect_to solr_document_path(@parent_resource)
+  end
+
+  def parent_resource
+    @parent_resource ||= find_resource(params[:id]).decorate.parent
```
hackmastera: I think we should prefer `Wayfinder.for(file_set).parents` over using the decorator.

It might be nice to extract a method while we're at it like
```
def file_set
  @file_set ||= find_resource(params[:id])
end
```
Since that `find_resource` is called so much.

christinach: ok I'll use the Wayfinder!

christinach:
@hackmastera can you check?

app/resources/file_sets/file_sets_controller.rb
```diff
@@ -8,6 +8,7 @@ class FileSetsController < ApplicationController
     metadata_adapter: Valkyrie::MetadataAdapter.find(:indexing_persister),
     storage_adapter: Valkyrie::StorageAdapter.find(:disk)
   )
+  before_action :parent_resource, only: [:destroy]
 
   def derivatives
     file_set = find_resource(params[:id])
```
hackmastera: there's one more here! I think you can just delete this line.

christinach: Thanks!

christinach: @hackmastera removed!

PR #3048, opened 2019-05-17
Ensures that MiniMagick errors are handled within CreateDerivativesJobs and RegenerateDerivativesJob by logging the error message and explicitly restarting the jobs

### LINKED ISSUE #2539: If a derivative fails to generate, it should throw an error.

jrgriffiniii:
Resolves #2539

### DIFF
--- a/spec/jobs/create_derivatives_job_spec.rb
+++ b/spec/jobs/create_derivatives_job_spec.rb
@@ -28,5 +28,25 @@ RSpec.describe CreateDerivativesJob do
     it "does not error with a non-existent file_set_id" do
       expect { described_class.perform_now("blabla") }.not_to raise_error
     end
+
+    context "when an ImageMagick error is raised" do
+      let(:derivatives_service) { instance_double(Valkyrie::Derivatives::DerivativeService) }
+
+      before do
+        allow(Rails.logger).to receive(:error)
+        allow(derivatives_service).to receive(:create_derivatives).and_raise(MiniMagick::Error)
+        allow(derivatives_service).to receive(:cleanup_derivatives)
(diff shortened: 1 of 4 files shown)


spec/jobs/create_derivatives_job_spec.rb
```diff
@@ -28,5 +29,37 @@
...
+
+  context "with ImageMagick uninstalled on your MacBook" do
+    let(:file) { fixture_file_upload("files/abstract.tiff", "image/tiff") }
+    let(:scanned_resource) { FactoryBot.create_for_repository(:scanned_resource, files: [file]) }
+    let(:file_set) { scanned_resource.decorate.file_sets.first }
+    let(:adapter) { Valkyrie::MetadataAdapter.find(:indexing_persister) }
+    let(:query_service) { adapter.query_service }
+
+    it "propagates an ImageMagick error", run_real_derivatives: true do
+      expect { described_class.perform_now(file_set.id.to_s) }.to raise_error(MiniMagick::Invalid)
```
jrgriffiniii: This does successfully propagate when I uninstall ImageMagick within my local development environment, thus I am uncertain as to how this would be failing silently if it the error is being raised by `MiniMagick`.

hackmastera: Yeah.. that makes me think we haven't found the right trigger for the silent fail.

jrgriffiniii:
This issue which this was to address should be resolved by https://github.com/pulibrary/figgy/pull/3040 (please see https://github.com/pulibrary/figgy/issues/2539#issuecomment-500568999)

Issue #2441, opened 2019-01-17
Update metadata for everything


tpendragon:
Requires #2440 

The job for automatically refreshing Voyager metadata has been broken (fixed when #2440 is fixed), so we need to refresh the metadata on all the voyager items.

hackmastera:
Write a rake task that queues up voyager updater jobs for everything with a source metadata identifier

tpendragon:
The rake task is done and all the jobs are queued up. I'll close this when they're all done.

tpendragon:
These are done, but this didn't close #2443 because it only updates ark metadata if the identifier, state, or title has changed.

PR #3049, opened 2019-05-17
Rake task to refresh remote metadata on all resources

tpendragon:
Work towards #2441

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -279,3 +279,6 @@ RSpec/ScatteredSetup:
   Exclude:
     - 'spec/resources/vector_resources/vector_resource_derivative_service_spec.rb'
     - 'spec/derivative_services/external_metadata_derivative_service_spec.rb'
+Style/PreferredHashMethods:
+  Exclude:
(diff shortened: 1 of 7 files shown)


app/queries/find_ids_with_property_set.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+class FindIdsWithPropertySet
```
hackmastera: `Property Set` is a little confusing because it could mean a `Set` like the noun. I know it's longer but maybe we could call it `FindIdsWithPropertyNotNil`? Or something else that's more specific? I think `FindIdsWithProperty` might even work.

spec/queries/find_ids_with_property_set_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe FindIdsWithPropertySet do
+  it "returns all resource ids with the given property set" do
```
hackmastera: maybe "returns all resource ids where the given property has a value"

PR #3051, opened 2019-05-20
Remove File manager from numismatic issue

### LINKED ISSUE #3046: Remove File Manager & Order Manager from Numismatic Issue
Issues aren't displayed publicly. File should be attached to the coin.

christinach:
closes #3046

### DIFF
--- a/app/resources/numismatic_issues/numismatic_issue_decorator.rb
+++ b/app/resources/numismatic_issues/numismatic_issue_decorator.rb
@@ -78,10 +78,14 @@ class NumismaticIssueDecorator < Valkyrie::ResourceDecorator
   end
 
   def manageable_files?
-    true
+    false
   end
 
   def manageable_structure?
+    false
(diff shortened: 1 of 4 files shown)


christinach:
Removing the File manager -> we upload a file through the coin File manager.
Removing the Order manager -> I think the only reason to keep it, would be to set a coin image as the 'resource thumbnail' for the numismatic issue.

Let me know if we want the Order Manager and I will update the code and put it back.

Maybe there are some details about the concept on how we use the Order Manager and how we use the File Manager on a resource, that I don't know.

escowles:
I'm not sure about the order manager.  Setting a thumbnail and reordering the coins seem like good things to do, but I'm not sure if they care about that, or if we should just use the first coin as the thumbnail.

christinach:
Thank you @escowles! I will put the Order Manager back then.

PR #3055, opened 2019-05-21
Update ability with numismatic classes- updates classes with depositor

### LINKED ISSUE #3004: Update Ability class with the numismatic classes

christinach:
closes #3004

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -64,7 +64,8 @@ class Ability
   end
 
   def curation_concerns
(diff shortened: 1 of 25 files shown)


christinach:
Besides Coin, NumismaticIssue and NumismaticMonogram do I need the depositor in the rest of the classes? I added it but I don't know what is the rule of thump for the depositor and how we decide when we want it and when not.

escowles:
@christinach I think it's fine to have depositor on all of the classes — it would help figure out who created what reference, people, place, etc. records if we wanted to figure that out.

christinach:
Thank you for letting me know @escowles

PR #3056, opened 2019-05-21
Require underscores to separate PULFA collection/component IDs

escowles:
* Refactor PULFA id validation and parsing
* Require underscores to separate PULFA collection/component IDs (instead of allowing either slash or underscore)
* Add form hints on acceptable ID formats

### DIFF
--- a/app/models/remote_record.rb
+++ b/app/models/remote_record.rb
@@ -18,19 +18,19 @@ class RemoteRecord
 
   def self.pulfa?(source_metadata_identifier)
     return false if source_metadata_identifier.match?(/\//)
-    source_metadata_identifier.match?(/^([A-Z][a-zA-Z0-9\.-]+)(_[c0-9]+)?/)
+    source_metadata_identifier.match?(/^([A-Z][a-zA-Z0-9\.-]+)(_[a-z0-9]+)?/)
   end
 
   def self.pulfa_collection(source_metadata_identifier)
     return if source_metadata_identifier.match?(/\//)
-    m = source_metadata_identifier.match(/^([A-Z][a-zA-Z0-9.-]+)[_]?.*/)
+    m = source_metadata_identifier.match(/^([A-Z][a-zA-Z0-9.-]+)([_][a-z0-9]+)?/)
     m[1] if m
   end
 
   def self.pulfa_component(source_metadata_identifier)
     return if source_metadata_identifier.match?(/\//)
     return unless source_metadata_identifier.match?(/_/)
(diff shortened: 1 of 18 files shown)


app/models/remote_record.rb
```diff
@@ -16,9 +16,31 @@ def self.bibdata?(source_metadata_identifier)
     PulMetadataServices::Client.bibdata?(source_metadata_identifier)
   end
 
+  def self.pulfa?(source_metadata_identifier)
+    return false if source_metadata_identifier.match?(/\//)
+    source_metadata_identifier.match?(/^([A-Z][a-zA-Z0-9\.-]+)(_[c0-9]+)?/)
```
tpendragon: Are we super confident this'll work? One reason we had the fallback from what I remember was that things were kind of inconsistent in component ID land.

app/models/remote_record.rb
```diff
@@ -16,9 +16,31 @@ def self.bibdata?(source_metadata_identifier)
...
+  end
+
+  def self.valid?(source_metadata_identifier)
+    bibdata?(source_metadata_identifier) || pulfa?(source_metadata_identifier)
+  end
+
   def self.source_metadata_url(id)
     return "https://bibdata.princeton.edu/bibliographic/#{id}" if bibdata?(id)
-    "https://findingaids.princeton.edu/collections/#{id.tr('_', '/')}.xml?scope=record"
+    return "https://findingaids.princeton.edu/collections/#{id.tr('_', '/')}.xml?scope=record" if pulfa?(id)
```
tpendragon: Does this error without the explicit return?

app/models/remote_record.rb
escowles: I just looked at the finding aids EAD XML and there are some variations (only numbers, or starting with "i" instead of "c").  I'll update the regex to be more lenient — probably any alphanumeric string.

tpendragon:
@escowles Do you have a plan for changing the existing "slash" items to underscores?

escowles:
@tpendragon Sorry, I forgot to add a migration.  I can add that.

db/migrate/20190521200107_update_source_metadata_identifier.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class UpdateSourceMetadataIdentifier < ActiveRecord::Migration[5.1]
+  def change
```
tpendragon: Do we do migrations for this? I think we've done rake tasks lately. I personally like this, but I know there's been some difference in opinion. If we keep it like this, this should just be a `def up` with an empty `def down`.

tpendragon:
@escowles This is good, but needs rebased.

PR #3082, opened 2019-05-22
updates numismatic_controllers_place test depositor

### LINKED ISSUE #3081: Periodically failing numismatic_places_controller test
https://github.com/pulibrary/figgy/blob/master/spec/resources/numismatic_places/numismatic_places_controller_spec.rb#L39

christinach:
closes #3081


christinach:
I updated the valid params so that when I query I find the specific issue that was created. The problem was that I was looking for city: "city" which is also created by default because of the factory in the another example with a nil user;

PR #3087, opened 2019-05-23
adds file name in numismatics manager

### LINKED ISSUE #2983: Monograms need more info the label
- Many of the monograms have the same description / title. We should probably add file name and monogram number.
- Also add these same columns to the index page so it's easier to sort.

tampakis:
Closes #2983


app/views/numismatic_monograms/index.html.erb
```diff
@@ -12,6 +13,7 @@
       <% @numismatic_monograms.each do |monogram| %>
         <tr>
           <td><%= render_thumbnail_tag monogram, { onerror: default_icon_fallback }, suppress_link: true %></td>
+          <td><%= monogram.decorated_file_sets&.first&.titles&.join(', ') %> </td>
```
christinach: Do you mind moving the logic 'decorated_file_sets&.first&.titles&.join(', ')' in a method for example 'def decorated_filename' in https://github.com/pulibrary/figgy/blob/master/app/resources/numismatic_monograms/numismatic_monogram_decorator.rb 
and also add a test in https://github.com/pulibrary/figgy/blob/master/spec/resources/numismatic_monograms/numismatic_monogram_decorator_spec.rb

Issue #3080, opened 2019-05-22
Convert MediaResource to be Recordings


tpendragon:

hackmastera:
this is done; the last outstanding ticket doesn't block closing this epic.

Issue #3021, opened 2019-05-14
Determine and document process for archival media bag ingest error correction


hackmastera:
Currently, when there are files the ingest script couldn't match to any particular component id, a media resource is created and they are all just stuck in there together.

When this happens, collections professionals can sort out the problems and this often involves fixing up the EAD so that files are matched to the correct component id. We need to determine the next course of action for the ingest.

Should we delete the media resources which ingested incorrectly and re-run the import? If so, that work flow should be tested for idempotence.

Should we delete and re-ingest the entire bag at once (so everything has the same upload set)?

Are there other ways of proceeding that make sense?

Once this is figured out, #3016 #3017 and #3018 can be used to test approaches, since they all need fixes.

hackmastera:
Implementing a way for collections professionals to resolve these manually could advance the use case of reorganizing a finding aid.

escowles:
Decision: for an entire Archival Media Collection, re-run the matching rules and move/combine/split MediaResources and/or FileSets according to the current rules and finding aid data.

hackmastera:
We'll accomplish this by making the following structural changes:
1. Files will be grouped into resources by physical origin (aka files that share a barcode / thumbnail image / `.xml` metadata file will be filesets on the same resource)
1. Additional resources will be created as proxies for the archival description. Component IDs will live here and metadata will be pulled from finding aids.

Here is a diagram:
https://docs.google.com/drawings/d/1kR_jV-84jXZNI3GfV8enDqv5rMHMizBxZx4-3GszggE/edit

hackmastera:
Then the ingest process will be updated to first load all the files out of the bag, then create the description proxies and organize the resources into them. The 2nd step of this process should be repeatable / idempotent.

tpendragon:
We're assuming that the structure will never go deeper than ArchivalMediaCollection -> MediaResource (component_id) -> MediaResource (barcode) -> FileSets. Where should we document that?

escowles:
@tpendragon That could be an ADR — is there somewhere better to document that?

hackmastera:
It seems like it would be good to have some user-facing documentation. Maybe near whatever UI we end up creating to close this ticket?

PR #3089, opened 2019-05-24
Rewrite IngestArchivalMediaBagJob and related classes

### LINKED ISSUE #3020: Archival media bag ingest should accommodate `_pt1`, etc.
In several cases the vendor had to split one side of a recording into multiple parts.

example:
https://github.com/pulibrary/figgy/issues/3018

tpendragon:
Implement new hierarchical model structure as per #3021
closes #3020
Generally clean up and refactor IngestArchivalMediaBagJob and
ArchivalMediaBagParser

We tried this out, and it ingests correctly, but there's no way to get from the descriptive proxy to the MediaResource with the FileSets. We expect we'll fix this as part of converting to using Recordings (#3080)


hackmastera:
To the reviewer: For IngestArchivalMediaBagJob and ArchivalMediaBagParser this is enough of a rewrite that the diffs might not be super helpful and you may want to look at the files directly.

spec/fixtures/files/pulfa/C0652/c0383.xml
```diff
@@ -1,7 +1,7 @@
...
-               Collections</corpname><subarea>Manuscripts Division</subarea><address><addressline>One Washington Road</addressline><addressline>Princeton, New Jersey 08544 USA</addressline></address></repository><container type="item" parent="C0652_i22">8</container><physdesc><extent type="computed" unit="folders">1 folder</extent></physdesc></did><altformavail><p>32101047382484_1_a.mp3</p><p>32101047382484_2_a.mp3</p><p>32101047382492_2_a.mp3</p></altformavail><scopecontent><p>2 audiocassettes; Tape 1, Side A written description on tape: Fitas / Monegal
+               Collections</corpname><subarea>Manuscripts Division</subarea><address><addressline>One Washington Road</addressline><addressline>Princeton, New Jersey 08544 USA</addressline></address></repository><container type="item" parent="C0652_i22">8</container><physdesc><extent type="computed" unit="folders">1 folder</extent></physdesc></did><altformavail><p>32101047382492_2_a.mp3</p><p>32101047382484_1_a.mp3</p><p>32101047382484_2_a.mp3</p></altformavail><scopecontent><p>2 audiocassettes; Tape 1, Side A written description on tape: Fitas / Monegal
```
escowles: Are there any real changes to these fixture files? One change is an added linebreak, but I can't see what the change is for the other two lines.

tpendragon: Yeah, it's hard to see. The order of the altformavails are switched to make the ordering test fail.

hackmastera: We had to switch the order of the barcodes to adequately test the order of the resulting resource members in one of our tests; we needed to manipulate the order in the file vs. the filesystem sort order.

escowles: 👍 OK, that makes sense and lines up with the test that uses it

escowles:
@hackmastera Also, thanks for the prompt to go look at the files instead of the diff — that was indeed much easier to read.

PR #3090, opened 2019-05-24
[wip] Do not display an empty viewer if the members have no file_sets

### LINKED ISSUE #2825: Don't display an empty UV when attaching a resource with no FileSet

christinach:
closes #2825

### DIFF
--- a/app/decorators/valkyrie/resource_decorator.rb
+++ b/app/decorators/valkyrie/resource_decorator.rb
@@ -20,6 +20,11 @@ class Valkyrie::ResourceDecorator < ApplicationDecorator
     @wayfinder ||= Wayfinder.for(object)
   end
 
+  # Do not display an empty viewer if the members have no file_sets
+  def member_filesets
+    wayfinder.members.select { |m| m.is_a?(FileSet) || m.is_a?(ProxyFileSet) ? m : m.member_ids.present? }
+  end
+
   # resource decorators will use this method if they define :member_of_collections
   #   in self.display_attributes
   def member_of_collections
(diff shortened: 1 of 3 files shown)


app/decorators/valkyrie/resource_decorator.rb
```diff
@@ -20,6 +20,11 @@ def wayfinder
     @wayfinder ||= Wayfinder.for(object)
   end
 
+  # Do not display an empty viewer if the members have no file_sets
+  def member_filesets
+    wayfinder.members.select { |m| m.is_a?(FileSet) || m.is_a?(ProxyFileSet) ? m : m.member_ids.present? }
```
hackmastera: I think this select encodes an assumption of the depth of nested resources, i.e., there won't be nesting beyond 2 resources before hitting `FileSets`. Is that true for all types of materials? I'm not sure, it's hard to keep track of them. If it is, it would be nice to have a comment indicating that this assumption is relied on here, and actually it would be nice to have an ADR stating that we intend to enforce that structure.

Also, I think the select would read a tiny bit easier if you said `? true :` instead of `? m :`; I spent some time trying to figure out how you were going to use the `m` from there.

christinach: @hackmastera the PR is in WIP for the reason that I wanted to check more the rest (all) of the resources and their nested depth. I would certainly add a comment to explain any assumption I make. 
Specifically I want to check the existence of file_sets or proxy_file_sets that might error when the universal_viewer loads.

christinach: the difficulty with this PR is that not all the resources have the same nested depth and it can have different cases (if conditions)

christinach:
Closing because we decided it didn't worth the processing time to hide the UV if there are not uploaded file_sets.

PR #3096, opened 2019-05-28
Recordings instead of MediaResources

### LINKED ISSUE #3079: IngestArchivalMediaBagJob should create Recordings instead of MediaResources

tpendragon:
I got a little ahead of myself. This PR does three things:

IngestArchivalMediaBagJob ingests Recordings instead.
Regenerating HLS derivatives works when there's an intermediate file.
Initial Manifest generation works

Closes #3079

### DIFF
--- a/app/derivative_services/default_derivative_service.rb
+++ b/app/derivative_services/default_derivative_service.rb
@@ -31,6 +31,7 @@ class DefaultDerivativeService
   end
 
   def valid?
+    return false unless original_file
     ["image/tiff", "image/jpeg"].include?(mime_type.first) && !parent.is_a?(ScannedMap)
   end
 
--- a/spec/jobs/ingest_archival_media_bag_job_spec.rb
+++ b/spec/jobs/ingest_archival_media_bag_job_spec.rb
@@ -43,12 +43,12 @@ RSpec.describe IngestArchivalMediaBagJob do
       expect(file_set.transfer_notes.first).to start_with "Side A"
     end
 
-    it "creates one MediaResource for the barcode and one for the component ID" do
-      expect(query_service.find_all_of_model(model: MediaResource).size).to eq 2
+    it "creates one ScannedResource for the barcode and one for the component ID" do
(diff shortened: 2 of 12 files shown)


tpendragon:
There's a bug which I think could be a new ticket: Audio files added from this job don't get derivatives created.

spec/jobs/ingest_archival_media_bag_job_spec.rb
```diff
@@ -43,12 +43,12 @@
       expect(file_set.transfer_notes.first).to start_with "Side A"
     end
 
-    it "creates one MediaResource for the barcode and one for the component ID" do
-      expect(query_service.find_all_of_model(model: MediaResource).size).to eq 2
+    it "creates one ScannedResource for the barcode and one for the component ID" do
```
hackmastera: In some of these comments it would maybe be nice to say "Recording" instead of "ScannedResource". Which leads me to wonder, should we be testing that these have the "recording" change set?

app/derivative_services/default_derivative_service.rb
```diff
@@ -31,6 +31,7 @@ def change_set
   end
 
   def valid?
+    return false unless original_file
```
hackmastera: What was happening that you needed to add all these?

tpendragon: There's no `original_file`, so it delegates `mime_type` to original_file and then `.first` can't be called on nil.

spec/jobs/ingest_archival_media_bag_job_spec.rb
tpendragon: Added a couple checks for that, and changed instances of ScannedResource in descriptions to Recording.

app/derivative_services/default_derivative_service.rb
hackmastera: right, of course. and i was making use of that fact for the rest of my morning 😀  thanks

PR #3104, opened 2019-05-30
Descriptive Proxy Recordings display as Recordings

### LINKED ISSUE #3100: Recording Descriptive Proxies are ingesting as MultiVolume Works
Screen Shot 2019-05-29 at 1 13 38 PM

tpendragon:
1. This makes the change set human readable model take priority over the fact that it has ScannedResource children.
2. Hides "Create a Playlist" button for descriptive proxies
3. Displays the member list if it's a descriptive proxy.

Closes #3100

### DIFF
--- a/app/views/catalog/_admin_controls_recording.html.erb
+++ b/app/views/catalog/_admin_controls_recording.html.erb
@@ -12,7 +12,7 @@
         <%= link_to "Order Manager", main_app.polymorphic_path([:order_manager, resource]), class: 'btn btn-default' %>
       <% end %>
     <% end %>
-    <% if can?(:update, resource) && Wayfinder.for(resource).scanned_resources_count == 0 %>
+    <% if can?(:update, resource) && Wayfinder.for(resource).file_sets_count > 0 %>
       <%= link_to "Create Playlist", main_app.playlists_path(recording_id: resource.id), method: :post, class: 'btn btn-primary' %>
     <% end %>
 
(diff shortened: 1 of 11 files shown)


app/views/catalog/_admin_controls_recording.html.erb
```diff
@@ -12,7 +12,7 @@
         <%= link_to "Order Manager", main_app.polymorphic_path([:order_manager, resource]), class: 'btn btn-default' %>
       <% end %>
     <% end %>
-    <% if can?(:update, resource) %>
+    <% if can?(:update, resource) && Wayfinder.for(resource).scanned_resources_count == 0 %>
```
hackmastera: I wonder if it makes sense to check that there are files, instead of there are not scanned resources?

tpendragon: I like this for consistency, and it seems reasonable to check to see if the object has any of the things that are being displayed rather than that it doesn't have something it wouldn't display. Are you concerned that one day we might have another model type as a nested resource? I have some concern that we might have both a FileSet and another type, and we'd still want to display this if so.

hackmastera: Your reasoning is why I was thinking of checking for the files. if there are files, then we could create a playlist out of them, so display the playlist link. It seems like you are thinking about the member list maybe? what is the thread of consistency?

PR #3108, opened 2019-05-31
Numismatics namespace refactor

eliotjordan:
Group all numismatics resources into a single directory and Refactor into a Numismatics namespace.

```
app/resources/numismatics
├── accessions
│   └── numismatics
│       ├── accession.rb
│       ├── accession_change_set.rb
│       ├── accession_decorator.rb
│       ├── accession_wayfinder.rb
│       └── accessions_controller.rb
├── artists
│   └── numismatics
│       ├── artist.rb
│       ├── artist_change_set.rb
│       ├── artist_decorator.rb
│       └── artist_wayfinder.rb
```


config/routes.rb
```diff
@@ -121,50 +121,45 @@
...
+          get :file_manager
+          get :order_manager
+          get :manifest, defaults: { format: :json }
+          post :browse_everything_files
+        end
       end
+      resources :references
     end
-    get "/numismatic_issues/:parent_id/coin" => "coins#new", as: :parent_new_coin
-    get "/parent/:parent_id/numismatic_artist" => "numismatic_artists#new", as: :parent_add_numismatic_artist
```
escowles: I'm not seeing Numismatics::Artist in the routes — is this not needed (only used nested)?

eliotjordan: Artist is nested. I think line 167 was leftover from an earlier refactor.

PR #3123, opened 2019-06-05
Display cloud fixity status on FileSets

### LINKED ISSUE #3111: Display cloud fixity on resources
Following on from #3023 — we should display cloud fixity status next to local fixity status on resources.

escowles:
Towards #3111


tpendragon:
This looks good to me.

Issue #3075, opened 2019-05-22
Recording manifests should display an attached image FileSet if there is one as the background of the viewer


tpendragon:
This means a Recording will have member_ids which are four FileSets: XML File/Image/Two sound Files. The manifest builder will say "I'm a recording, I have an image FileSet, make it my background. Don't do anything with the XML file."

jrgriffiniii:
https://github.com/pulibrary/figgy/compare/issues-3075-jrgriffiniii-background-av-fileset updates the `ManifestBuilder` to generate manifests for `FileSets` with images on `Recordings`, but this still seems to be broken in the UV.  I'll need to look towards https://universalviewer.io/examples/#?c=&m=&s=&cv=&manifest=http%3A%2F%2Fwellcomelibrary.org%2Fiiif%2Fb17307922%2Fmanifest for the manifest structure.

jrgriffiniii:
Overriding the `CanvasBuilder` to use the `thumbnail` property did not seem to yield any fruitful results:
![image](https://user-images.githubusercontent.com/1443986/58846780-7fba5600-864e-11e9-8490-8ae21279935f.png)

Overriding the `ContentBuilder` (https://github.com/samvera-labs/iiif_manifest/blob/master/lib/iiif_manifest/v3/manifest_builder/content_builder.rb#L4) in order to provide the `thumbnail` property at the level of the Content Resource (https://iiif.io/api/presentation/3.0/#57-content-resources) might be more effective.

jrgriffiniii:
This didn't prove effective, not did attempting to pass the `thumbnail` to the `Annotation`.  https://github.com/IIIF-Commons/iiif-av-component/blob/master/src/AVComponent.ts#L2533 seems to be where the background for audio resources would be set, I'll be tracing the functionality from this point.

jrgriffiniii:
This has been successfully tested with a fixture for the UV, I can proceed by looking to use the needed image `FileSet`:

![image](https://user-images.githubusercontent.com/1443986/58962104-788f6700-8778-11e9-80f4-fac9427ff19f.png)

jrgriffiniii:
https://github.com/samvera-labs/iiif_manifest/blob/master/lib/iiif_manifest/v3/manifest_builder/canvas_builder.rb#L49 seems to be problematic when `record.display_content` is just empty (but not `nil`).

jrgriffiniii:
[...]

PR #3129, opened 2019-06-10
Generates the `posterCanvas` property for IIIFv3 Manifests where Recordings have image FileSets

jrgriffiniii:
Advances #3075

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -339,11 +339,25 @@ class ManifestBuilder
 
   class RecordingNode < RootNode
     def leaf_nodes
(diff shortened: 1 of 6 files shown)


app/services/manifest_builder.rb
```diff
@@ -605,7 +620,7 @@ def manifest_image_thumbnail_path(resource)
     end
 
     def manifest_image_medium_path(resource)
-      "#{manifest_image_path(resource)}/full/!1000,/0/default.jpg"
+      "#{manifest_image_path(resource)}/full/1000,/0/default.jpg"
```
jrgriffiniii: This was necessary for me when I was testing using `riiif`, and I could not find a solution for resolving this within my local development environment.

PR #3131, opened 2019-06-10
Ensures that ScannedResourcesController#change_set_class returns DynamicChangeSet when the change_set parameter is invalid

### LINKED ISSUE #2722: [Figgy/production] NameError: uninitialized constant NewChangeSet
## Backtrace

    line 12 of [PROJECT_ROOT]/app/change_sets/dynamic_change_set.rb: class_from_param
    line 26 of [PROJECT_ROOT]/app/resources/scanned_resources/scanned_resources_controller.rb: change_set_class
    line 109 of [PROJECT_ROOT]/app/resources/scanned_resources/scanned_resources_controller.rb: ingest_folder

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/46908296)

jrgriffiniii:
Resolves #2722 (previously this was only handled at the level of `ResourceController#new`, but this is invoked from other contexts; please see https://app.honeybadger.io/projects/53391/faults/46908296)

### DIFF
--- a/app/resources/scanned_resources/scanned_resources_controller.rb
+++ b/app/resources/scanned_resources/scanned_resources_controller.rb
@@ -22,15 +22,11 @@ class ScannedResourcesController < BaseResourceController
   end
 
   def change_set_class
-    if params[:change_set].present? || (resource_params && resource_params[:change_set].present?)
-      change_set_name = params[:change_set] || resource_params[:change_set]
-      DynamicChangeSet.class_from_param(change_set_name)
+    if change_set_param.present?
+      DynamicChangeSet.class_from_param(change_set_param)
     else
       DynamicChangeSet
     end
-  rescue NameError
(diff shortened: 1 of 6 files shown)


app/resources/scanned_resources/scanned_resources_controller.rb
```diff
@@ -23,10 +23,14 @@ def after_create_success(obj, change_set)
 
   def change_set_class
     if params[:change_set].present? || (resource_params && resource_params[:change_set].present?)
-      DynamicChangeSet.class_from_param(params[:change_set] || resource_params[:change_set])
+      change_set_name = params[:change_set] || resource_params[:change_set]
+      DynamicChangeSet.class_from_param(change_set_name)
     else
       DynamicChangeSet
     end
+  rescue NameError
```
escowles: If the problem is that this error occurs in several different contexts, maybe it would be better to handle it in `DynamicChangeSet` instead of in the controller?

PR #3134, opened 2019-06-17
Preserve in a background job, always.

tpendragon:
Updates were way too slow for preserved resources that were updating all
children.


jrgriffiniii:
https://circleci.com/gh/pulibrary/figgy/18025#tests/containers/2 has failures which appear to be unrelated, I'm going to restart the build.

tpendragon:
@jrgriffiniii Those were real failures. Pushed up a fix.

PR #3137, opened 2019-06-18
Restructures Archival Media Collections as ChangeSets

### LINKED ISSUE #3078: Refactor ArchivalMediaCollections to use a ChangeSet on a Collection object
This should have a bag_path property which shows up on the edit screen, and it should index as both a Collection and an ArchivalMediaCollection

jrgriffiniii:
Resolves #3078


app/resources/collections/collections_controller.rb
```diff
@@ -43,6 +43,22 @@ def ark_report
...
+    if params[:change_set].present? || (resource_params && resource_params[:change_set].present?)
+      DynamicChangeSet.class_from_param(params[:change_set] || resource_params[:change_set])
+    else
+      DynamicChangeSet
+    end
+  end
+
+  def render_new
+    if change_set_param
+      redirect_to polymorphic_path([:new, change_set_param, self.class.resource_class])
```
jrgriffiniii: This is still a problem, but `render :new` does not preserve the state of the `@change_set`.

jrgriffiniii: I was incorrect, the value of the `change_set` and the state captured with `@change_set` are preserved, but the request parameters do not contain the ChangeSet name (which are what is parsed in https://github.com/pulibrary/figgy/blob/master/app/helpers/metadata_form_helper.rb#L11).

app/helpers/metadata_form_helper.rb
```diff
@@ -1,10 +1,19 @@
...
-      "#{params['action'].capitalize} #{params['change_set'].humanize.singularize.downcase} resource"
+    controller_name = params["controller"]
+    model_name = controller_name.singularize.downcase
+    action = params["action"].capitalize
+    action = "New" if action == "Create"
+
+    model_params = params[model_name]
+
+    if model_params && form_custom_title(model_params)
+      "#{action} #{model_params['change_set'].humanize.singularize.downcase} resource"
```
jrgriffiniii: Without this, when `create` actions raise errors, the title for the form will just be `Collection` (even when the `change_set` is set properly to `archival_media_collection`)

app/indexers/human_readable_type_indexer.rb
```diff
@@ -8,11 +13,28 @@ def initialize(resource:)
...
+    def index_change_set?(change_set)
+      change_set_class_name = "#{change_set.camelize}ChangeSet"
+      change_set_class = change_set_class_name.constantize
+      self.class.change_sets.include?(change_set_class)
+    rescue NameError
+      Valkyrie.logger.warn("#{change_set} is not a valid resource type.")
+      false
+    end
+
+    def human_readable_type
```
jrgriffiniii: I'm not quite certain about this approach, but I felt that it was more acceptable to attempt something similar to this than to create an additional Indexer which just overrides the `human_readable_type_ssim` field values.

Issue #3041, opened 2019-05-16
Add preservation support to a/v resources


hackmastera:
Do we want a whole Archival Media Collection package to be the unit of preservation, or each Media Resource?

Implement by adding the preserve property.

jrgriffiniii:
https://github.com/pulibrary/figgy/compare/issues-3041-jrgriffiniii-av-preservation introduces preservation support for `MediaResourceChangeSets`, but it seems to be constructing `PreservationObjects` for both the `MediaResource` and `FileSet` separately.

tpendragon:
It's correct behavior that it's constructing `PreservationObjects` for both `MediaResource` and `FileSet`.

tpendragon:
To close this ticket we need to make sure that on import the Recording is set to be preserved and test it in staging.

tpendragon:
Tested. Works!

PR #3140, opened 2019-06-19
Introduces preservation support for MediaResources

jrgriffiniii:
Advances #3041

### DIFF
--- a/app/characterization_services/mediainfo_characterization_service.rb
+++ b/app/characterization_services/mediainfo_characterization_service.rb
@@ -152,7 +152,7 @@ class MediainfoCharacterizationService
     # Retrieves the master binary file in this FileSet
     # @return [FileNode]
     def preservation_file
-      if parent.try(:image_resource?)
+      if parent.try(:image_resource?) || parent.try(:media_resource?)
         @file_set.original_file
       else
         @file_set.preservation_file
--- a/app/characterization_services/mediainfo_characterization_service.rb
+++ b/app/characterization_services/mediainfo_characterization_service.rb
@@ -46,6 +46,7 @@ class MediainfoCharacterizationService
   # Determines if the parent of the FileSet is a MediaResource
   # @return [TrueClass, FalseClass]
   def valid?
+    return false if preservation_file.nil?
     (parent.try(:media_resource?) || parent.try(:image_resource?)) && supported_format?
   end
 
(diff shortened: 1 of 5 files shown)


app/characterization_services/mediainfo_characterization_service.rb
```diff
@@ -152,7 +152,7 @@ def file_object
     # Retrieves the master binary file in this FileSet
     # @return [FileNode]
     def preservation_file
-      if parent.try(:image_resource?)
+      if parent.try(:image_resource?) || parent.try(:media_resource?)
         @file_set.original_file
       else
         @file_set.preservation_file
```
jrgriffiniii: Given that the `MediainfoCharacterizationService` is only configured for the media types `audio/mpeg` and `audio/x-wav`, I'm uncertain as to whether this condition would be reached.

jrgriffiniii: No, this is not the case, https://github.com/pulibrary/figgy/blob/master/app/characterization_services/mediainfo_characterization_service.rb#L49 will never return `true` if another resource (e. g. `ScannedResource`) is provided with A/V FileSets.

app/characterization_services/mediainfo_characterization_service.rb
```diff
@@ -152,10 +153,6 @@ def file_object
     # Retrieves the master binary file in this FileSet
     # @return [FileNode]
     def preservation_file
-      if parent.try(:image_resource?)
-        @file_set.original_file
-      else
-        @file_set.preservation_file
-      end
+      @file_set.original_file
```
tpendragon: So wait, A/V bags don't use this characterizer?

jrgriffiniii: I'm still investigating this, but for the case with https://github.com/pulibrary/figgy/blob/master/spec/jobs/ingest_archival_media_bag_job_spec.rb#L23, while it has files with the necessary media types (https://github.com/pulibrary/figgy/tree/master/spec/fixtures/av/la_c0652_2017_05_bag/data), it does not fulfill the `(parent.try(:media_resource?) || parent.try(:image_resource?))` condition in https://github.com/pulibrary/figgy/blob/master/app/characterization_services/mediainfo_characterization_service.rb#L49

tpendragon: Maybe stubbed characterization is causing it?

jrgriffiniii: I've restored this, I suspect that it may have been https://github.com/pulibrary/figgy/pull/3140/files#diff-2b6680eb7380002602ad9e9f157a6942R49 which was needed.

PR #3141, opened 2019-06-19
Don't display inaccessible manifests in Collection manifest.

tpendragon:
Fixes display for IIIF Universe

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -4,8 +4,8 @@ class ManifestBuilder
 
   ##
   # @param [Resource] resource the Resource subject
-  def initialize(resource, auth_token = nil)
-    @resource = RootNode.for(resource, auth_token)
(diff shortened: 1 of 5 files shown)


app/services/manifest_builder.rb
```diff
@@ -4,8 +4,8 @@ class ManifestBuilder
 
   ##
   # @param [Resource] resource the Resource subject
-  def initialize(resource, auth_token = nil)
-    @resource = RootNode.for(resource, auth_token)
+  def initialize(resource, auth_token = nil, current_ability = nil)
```
jrgriffiniii: I can create a separate issue for this, but I suspect that it might be easier to offer keyword arguments for the `ManifestBuilder` constructor.

tpendragon: :+1: I just didn't wanna change a bunch right now.

PR #3144, opened 2019-06-20
 Shell out to shasum/md5sum rather than doing a full bag validation

tpendragon:
Should be significantly faster.


tpendragon:
I closed this because while it's a faster, a bunch of the spec sha files aren't properly formatted for `shasum` - it only has one space instead of two.

PR #3175, opened 2019-07-02
Fix file set cloud fixity display.

tpendragon:
For IDs which aren't in a PreservationObject it was returning all
events.

### DIFF
--- a/app/resources/file_sets/file_set_decorator.rb
+++ b/app/resources/file_sets/file_set_decorator.rb
@@ -49,7 +49,9 @@ class FileSetDecorator < Valkyrie::ResourceDecorator
   end
 
   def cloud_fixity_events_for(file_id)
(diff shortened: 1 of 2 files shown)


app/resources/file_sets/file_set_decorator.rb
```diff
@@ -49,7 +49,9 @@ def custom_queries
   end
 
   def cloud_fixity_events_for(file_id)
-    custom_queries.find_by_property(property: :child_id, value: preservation_id_of(file_id))
+    preservation_id = preservation_id_of(file_id)
+    return [] unless preservation_id.present?
+    custom_queries.find_by_property(property: :child_id, value: preservation_id)
```
jrgriffiniii: I'm curious as to why this was returning all `Events` as opposed to `[]` (I supposed because `null` would be valid for all JSON-serialized metadata for `Events`), but that's just a documentation issue.

PR #3180, opened 2019-07-02
Extends the ManifestBuilder using MultiVolumeRecordingNode for handling cases where Recordings do not have FileSets but have FileSets within member Recordings

### LINKED ISSUE #3156: UV can't play a collection of recordings
This is a problem for our AV bags because they're typically a grouping of barcodes under one component ID. We can probably fix this by inlining all the child file sets for generation of the manifest (like we do for Sammelbands.)

jrgriffiniii:
Fully resolves #3156

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -18,6 +18,17 @@ class ManifestBuilder
   ##
   # Presenter modeling the Resource subjects as root nodes
   class RootNode
+    def self.multi_volume_recording?(resource)
+      decorated = resource.decorate
+      return unless decorated.respond_to?(:volumes) && !decorated.volumes.empty?
+
+      volumes = decorated.volumes
+      volume_file_sets = volumes.map(&:file_sets)
+      volume_file_sets.flatten!
+      audio_file_sets = volume_file_sets.select(&:audio?)
+      !audio_file_sets.empty?
+    end
+
     def self.for(resource, auth_token = nil, current_ability = nil)
       case resource
       when Collection
@@ -40,7 +51,11 @@ class ManifestBuilder
       else
         case DynamicChangeSet.new(resource)
         when RecordingChangeSet
-          RecordingNode.new(resource)
+          if multi_volume_recording?(resource)
+            MultiVolumeRecordingNode.new(resource)
+          else
+            RecordingNode.new(resource)
+          end
         else
           new(resource, auth_token)
         end
@@ -115,9 +130,14 @@ class ManifestBuilder
     end
 
     def audio_manifest?
-      file_set_presenters.find do |fs_presenter|
+      audio_file_sets = file_set_presenters.select do |fs_presenter|
         fs_presenter.display_content.present?
-      end.present?
+      end
+
+      work_presenter_nodes = Array.wrap(work_presenters)
+      work_presenters = work_presenter_nodes.select(&:audio_manifest?)
+
+      !audio_file_sets.empty? || !work_presenters.empty?
     end
 
     def audio_ranges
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/figgy_boot.es6
```diff
@@ -18,7 +18,6 @@ import BoundingBoxSelector from "bounding_box_selector"
 export default class Initializer {
   constructor() {
     this.server_uploader = new ServerUploader
-    this.collection_bag_uploader = new CollectionBagUploader
```
jrgriffiniii: This needs to be extracted and moved to a separate PR.

jrgriffiniii: These changes are now addressed with https://github.com/pulibrary/figgy/pull/3181

jrgriffiniii:
The Range labels are currently generated from the FileSet labels for each audio resource in the UV, but these are rendered as a shallow list:
image


There should be nested in structure, with the parent resource as a parent node on the tree display.

app/services/manifest_builder/manifest_service_locator_v3.rb
```diff
@@ -70,6 +70,36 @@ def canvas_builder
...
+        IIIFManifest::ManifestBuilder::ChildManifestBuilderFactory.new(
+          composite_builder: composite_builder,
+          child_manifest_builder: child_manifest_builder
+        )
+      end
+
+      def child_manifest_builders
+        composite_builder_factory.new(
+          shallow_child_manifest_builder_factory,
+          structure_builder,
```
jrgriffiniii: When attempting an appraoch where `Collections` are generated (as opposed to `Manifests`) for multi-volume `Recordings`, this needs to be injected into the `child_manifest_builder` used to generate the Manifests in https://github.com/samvera-labs/iiif_manifest/blob/master/lib/iiif_manifest/manifest_builder/child_manifest_builder_factory.rb#L13.  Otherwise the `Ranges` within each member `Manifest` level simply aren't generated.

jrgriffiniii:
https://github.com/samvera-labs/iiif_manifest/blob/master/lib/iiif_manifest/manifest_builder/child_manifest_builder_factory.rb#L12 is using the `work_presenters` to generate the Manifests, so I'll attempt to generate the ranges by overriding the `child_manifest_builder` here.

app/services/manifest_builder/manifest_service_locator_v3.rb
```diff
@@ -70,6 +70,33 @@ def canvas_builder
...
+          composite_builder: composite_builder
+        )
+      end
+
+      def child_manifest_builder
+        IIIFManifest::ManifestServiceLocator::InjectedFactory.new(
+          ManifestBuilder,
+          builders: composite_builder_factory.new(
```
jrgriffiniii: Overriding this upstream in a local fork of the `iiif_manifest` Gem did successfully generate the `Ranges` for member Manifests, but they still are not rendered in the UV properly.

jrgriffiniii:
The `Structures` are properly generated within the `Collection` (one `Range` within a `Structure` for each `Manifest`), and each member `Manifest` seems to have exactly one `Canvas` within `items`.  However, there are still errors in the nested display:

image

jrgriffiniii:
This nesting is actually not featured for `Playlists`, so I'm going to need to find external examples for reference when looking to support this from within the Universal Viewer:

image

jrgriffiniii:
One example that appears to work follows this structure:
image

image

jrgriffiniii:
https://github.com/pulibrary/figgy/pull/3180/files#diff-55f651679ae90d845871424d7d4e9bd3R393 ensures that each member `Recording` is rendered without the `FileSet`:
image

The `FileSets` probably aren't rendered given that the `Range` `Canvases` should have `labels` (but, do not):
image

jrgriffiniii:
https://github.com/pulibrary/figgy/pull/3180/commits/3c0610c6e1874bbccce40d911133b99143fca44d adds `labels` to the Canvases, but these still don't change the content in the `Index` tab for the `Content` pane:
image
image

`Ranges` within each top-level `Range` must be what is required.

jrgriffiniii:
https://github.com/pulibrary/figgy/pull/3180/commits/066160edf3f7fe020ece6133ba2862821c4984d9 has these Ranges rendered, just testing this in the staging environment now.

tpendragon:
I rebased this in a different branch which I'll deploy to staging to test as soon as the Archival Media Recordings get re-ingested there (working on that now.)

PR #3197, opened 2019-07-15
Ensuring that only image files in the TIFF and JPEG are imported by the NumismaticsImportService

### LINKED ISSUE #3025: Only search for .tif and .TIF coin images during import
There are also jpegs, pngs, and psds in the directories. Filter these out and only import files with `.tif` and `.TIF` extensions.

https://github.com/pulibrary/figgy/blob/master/app/services/numismatics_import_service.rb#L307

jrgriffiniii:
Resolves #3025


eliotjordan:
@jrgriffiniii The PR looks good, however, it turns out we're going to need to import JPGs as well. There are many missing coin TIFFs that we only have derivative images for.

PR #3210, opened 2019-07-17
Ensuring that JSON-LD expressions are available for resources which are complete and non-private

### LINKED ISSUE #3204: JSON/LD for records need to be accessible for all complete items that are not private
So that dpul can index all records even though we lock down manifests.

jrgriffiniii:
Also adds response media type handling for linked data formats in ApplicationController.  Resolves #3204

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -42,6 +42,7 @@ class ApplicationController < ActionController::Base
           head :forbidden
         end
       end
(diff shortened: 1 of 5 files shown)


tpendragon:
@hackmastera I agree. I think this was just an unfortunate case where we were implementing something at the same time and didn't realize their implementation requirements were the same.

jrgriffiniii:
Duplicates https://github.com/pulibrary/figgy/pull/3215

PR #3215, opened 2019-07-18
Open Items in GraphQL Response

### LINKED ISSUE #3206: Update GraphQL to return more records
GraphQL should return non-private complete items, but provide an authorization message property which Orangelight or Finding Aids should display if there’s some other way to access the material (login or be on campus/VPN or reading room or whatever). It should also not return the thumbnail URL. May also need an authorization code.

This will support orangelight logic in determining how to display links to digital objects.

tpendragon:
Closes #3206

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -34,7 +34,7 @@ class Types::QueryType < Types::BaseObject
 
   def resource(id:)
     resource = query_service.find_by(id: id)
-    return unless ability.can? :read, resource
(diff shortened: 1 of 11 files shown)


tpendragon:
Orangelight will need to use the correct viewer link, and it's not grabbing cookies appropriately to render thumbnails.

tpendragon:
When https://github.com/pulibrary/orangelight/pull/1738 is deployed here's an example: https://catalog-staging.princeton.edu/catalog/823463

PR #3231, opened 2019-07-26
Add Release Notes to Figgy

### LINKED ISSUE #3225: Create a changelog
with process

Requested at ADAPT+ meeting as well as github presentation.

### LINKED ISSUE #3228: CI should fail a step if we don't add to release notes
This step shouldn't block us from merging, but should force us to think about if something should be added there.

tpendragon:
Closes #3225 
Closes #3228 
Screen Shot 2019-07-26 at 5 48 16 PM

### DIFF
--- a/app/assets/stylesheets/components/header.scss
+++ b/app/assets/stylesheets/components/header.scss
@@ -34,10 +34,6 @@ header {
 }
 
 @media (min-width: $bp-medium) {
-  .l-branding__pul {
-    height: 43px;
-  }
-
   .l-branding__pul .logo__pul {
     text-align: left;
   }
--- a/app/assets/stylesheets/components/header.scss
+++ b/app/assets/stylesheets/components/header.scss
@@ -258,3 +258,8 @@ header {
     }
   }
 }
+.menu--level-1,
+.logo__pul {
+  height: 40px;
+  position: relative;
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,5 +1,7 @@
 <% provide :body do %>
-<%= render 'shared/header' %>
+<div class="lux">
+  <%= render 'shared/header' %>
+</div>
 <%= render 'shared/flash_message' %>
 <main>
   <div id="main" role="main" class="<%= container_type %> <%= yield(:page_class) if content_for?(:page_class)%> lux">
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/components/header.scss
```diff
@@ -258,3 +254,18 @@ header {
...
+  }
+
+  .menu--level-1>ul {
+    display: inline-block;
+
+    @media (min-width: 645px) {
+      margin-top: 0;
+      @include vertical-align;
+    }
+  }
```
hackmastera: do we still need the changes in this file with the lux header?

app/views/layouts/application.html.erb
```diff
@@ -1,5 +1,7 @@
 <% provide :body do %>
-<%= render 'shared/header' %>
+<div class="lux">
```
hackmastera: @sdellis we added a second lux load... is this bad? There was one already for the order manager

config/initializers/extra_extra.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+ExtraExtra.renderer_options = {
+  header_level_offset: 1, # h1's are h1
```
hackmastera: My understanding is that the only `h1` should be your home link, and page headings should start at h2.

hackmastera:
@tpendragon if you want me to mess with this further I'm happy to; I know you're on another sprint this week.

app/assets/stylesheets/components/header.scss
tpendragon: We don't need the ul bit, but do need to fix the height

config/initializers/extra_extra.rb
tpendragon: Fixed.

app/assets/stylesheets/components/header.scss
tpendragon: Fixed.

app/views/layouts/application.html.erb
tpendragon: They shouldn't run into each other, but yeah @sdellis would know better.

hackmastera: 👍 confirmed via slack: "yes you can load more than one lux mounting element. I think there may be an issue with multiple mounting elements  in the static build"

PR #3239, opened 2019-07-30
Don't add @princeton.edu to UIDs when they are email addresses already

### LINKED ISSUE #1482: Email addresses are wrong for non-Princeton users
It's possible for non-Princeton users to get accounts, and their NetIDs are their (external) email addresses.  We currently just add `@princeton.edu` to make their email addresses, resulting in malformed email addresses like `foo@bar.com@princeton.edu`

escowles:
Fixes #1482

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,12 @@ Release notes template:
 ## Removed
 
 -->
+# 2019-08-02
+
+## Fixed
+
+* Email addresses for external users no longer have "@princeton.edu" added to the end of valid email addresses.
+
 # 2019-07-26
 
 ## Added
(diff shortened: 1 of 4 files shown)


tpendragon:
Should we have a migration for existing users?

escowles:
@tpendragon Yep, just added that as a rake task.

tpendragon:
@escowles Can you rebase this? I expect it to fail CI now, since this should be in release notes.

escowles:
👍 I rebased and added a release notes entry (also opened https://github.com/pulibrary/figgy/pull/3248 with a straight rebase to make sure it failed, and it does)

RELEASE_NOTES.md
```diff
@@ -8,6 +8,8 @@ Release notes template:
 
 ## Fixed
 
+* Email addresses for external users no longer have "@princeton.edu" added to the end of valid email addresses.
```
tpendragon: I think you added this to the documentation template instead of as a new entry.

escowles: 😞 yes, that's exactly what I did.

PR #3253, opened 2019-08-05
Remove site-title in favor of lux header

### LINKED ISSUE #3241: Remove one of the "Figgy" links from header/s
Now that we're using lux we should probably remove the Figgy link from the search bar area.

hackmastera:
fixes #3241

Screen Shot 2019-08-05 at 3 11 09 PM


tpendragon:
Release notes?

hackmastera:
Sure, I guess so! It's good we have that check there!

PR #3274, opened 2019-08-08
Update visibility badge notices with new rules

### LINKED ISSUE #3199: Update sync message when using new visibility values
reading_room and on_campus visibility should not prevent an item from syncing to dpul, etc. Currently the help text says they won't sync. Once new sync logic is put into place in dpul, correct the help text.

hackmastera:
fixes #3199

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -45,18 +45,24 @@ en:
       warning_html: ''
       class: "label-danger"
   computed_visibility:
-    public:
+    open:
       class: "alert-success"
-      note_html: "This item will sync to the Catalog, DPUL, Maps Portal, and/or LAE."
-    suppressed_visibility:
-      class: "alert-warning"
-      note_html: "This item will not sync to the Catalog, DPUL, Maps Portal, or LAE due to the visiblity setting."
+      note_html: "Users will be able to view this digital object on discovery sites."
+    authenticated:
+      class: "alert-info"
+      note_html: "Users will be able to view this digital object on discovery sites if they log in with CAS."
+    on_campus:
+      class: "alert-info"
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -45,18 +45,24 @@ en:
...
+      note_html: "Users will be able to view this digital object on discovery sites."
+    authenticated:
+      class: "alert-info"
+      note_html: "Users will be able to view this digital object on discovery sites if they log in with CAS."
+    on_campus:
+      class: "alert-info"
+      note_html: "Users will be able to view this digital object on discovery sites if they are on campus or logged in to campus VPN."
+    reading_room:
+      class: "alert-info"
+      note_html: "Users will be able to view this digital object on discovery sites if they are on campus or logged in to campus VPN."
```
escowles: Shouldn't this say something about being in the reading room?

spec/decorators/valkyrie/resource_decorator_spec.rb
```diff
@@ -223,32 +223,79 @@ class MyResource < Resource
...
+
+      it "has a warning about the workflow" do
+        expect(decorator.visibility.first).to have_selector("div.alert-warning", text: "Users will not be able to view this digital object on discovery sites due to the workflow status.")
+      end
+    end
+
+    context "complete reading room resource" do
+      let(:resource) { FactoryBot.build(:complete_reading_room_scanned_resource) }
+      it "has a permissions / location notice" do
+        note = "Users will be able to view this digital object on discovery sites if they are on campus or logged in to campus VPN."
```
escowles: Same comment here as above — shouldn't this reference being in the reading room?

hackmastera:
@escowles thank you for catching that!! fixed.

PR #3289, opened 2019-08-13
Add migration to fix PreservationObjects with no FileMetadata IDs

tpendragon:
This generates IDs for those PreservationObjects which don't have them in their FileMetadata nodes (a result of a bug which has since been fixed), and also deletes any events which try to associate with child_ids that don't exist.

Edit: A fair warning that since most of our events are from the pilot, a vast majority of our events are going to get deleted (all but 553 of them.) However, there's no reasonable way to determine which binary_node an Event was meant for, so it can't really be helped.

### DIFF
--- a/spec/services/migrations/add_preservation_object_ids_migrator_spec.rb
+++ b/spec/services/migrations/add_preservation_object_ids_migrator_spec.rb
@@ -19,6 +19,23 @@ RSpec.describe Migrations::AddPreservationObjectIdsMigrator do
       expect(after_result.flat_map(&:binary_nodes).map(&:id).compact.length).to eq 2
     end
 
+    it "deletes all events that don't have a child ID" do
+      FactoryBot.create_for_repository(
+        :event,
+        child_id: Valkyrie::ID.new("")
+      )
+      FactoryBot.create_for_repository(
+        :event,
+        child_id: Valkyrie::ID.new(SecureRandom.uuid)
+      )
+
+      expect(query_service.custom_queries.count_all_of_model(model: Event)).to eq 2
+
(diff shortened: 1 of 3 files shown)


spec/services/migrations/add_preservation_object_ids_migrator_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      before_count = query_service.custom_queries.find_by_property(property: :metadata_node, value: { id: {} }).size
+      expect(before_count).to eq 1
+
+      described_class.call
+
+      after_result = query_service.custom_queries.find_by_property(property: :metadata_node, value: { id: {} })
+      expect(after_result.size).to eq 2
+      expect(after_result.map(&:metadata_node).map(&:id)).to include preservation_object1.metadata_node.id
+      expect(after_result.flat_map(&:binary_nodes).map(&:id).compact.length).to eq 2
```
jrgriffiniii: 👍

spec/services/migrations/add_preservation_object_ids_migrator_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      FactoryBot.create_for_repository(
+        :event,
+        child_id: Valkyrie::ID.new(SecureRandom.uuid)
+      )
+
+      expect(query_service.custom_queries.count_all_of_model(model: Event)).to eq 2
+
+      described_class.call
+
+      expect(query_service.custom_queries.count_all_of_model(model: Event)).to eq 1
```
jrgriffiniii: 👍

PR #3301, opened 2019-08-19
Check that change in release notes adds a new message

### LINKED ISSUE #3267: Improve reliability of release note check
it fails to block if the something was merged since the last was created from master

hackmastera:
to prevent false positives on this check
fixes #3267


hackmastera:
This should be ready; tested all 3 cases locally. Doesn't seem like it should have a release notes entry.

PR #3302, opened 2019-08-19
Bulk Ingest Updates

escowles:
* Support the `replaces` attribute
* Move BibID-guessing to BulkIngestController, to allow command-line ingest of items without BibIDs

### DIFF
--- a/spec/controllers/bulk_ingest_controller_spec.rb
+++ b/spec/controllers/bulk_ingest_controller_spec.rb
@@ -66,6 +66,10 @@ RSpec.describe BulkIngestController do
     end
 
     context "with one single-volume resource" do
+      before do
+        stub_bibdata(bib_id: 4609321)
+      end
+
       let(:attributes) do
         {
           workflow: { state: "pending" },
@@ -78,7 +82,7 @@ RSpec.describe BulkIngestController do
       let(:selected_files) do
         {
           "0" => {
-            "url" => "file:///base/resource1/1.tif",
+            "url" => "file:///base/4609321/1.tif",
             "file_name" => "1.tif",
             "file_size" => "100"
           }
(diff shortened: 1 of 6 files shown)


tpendragon:
I'm kind of confused by this. Does this mean that bulk ingest from the command line can't do source metadata identifier guessing anymore?

escowles:
Yes, command-line ingest doesn't try to use the directory name, it requires that you specify it separately using the `BIB` environment variable.  I think we should make the change because otherwise there is no way to ingest items without bib records, when their directories happen to be the same as an existing bib number (e.g., `007`).

hackmastera:
If we have both use cases maybe a flag?

escowles:
Do we have both use cases? I thought using the directory as the bibid was added for the bulk ingest GUI — is adding the extra env variable to the command-line ingest a problem?

tpendragon:
I think the command line use case is probably for us bulk ingesting a directory for the studio or something. We haven't really done it much lately...

spec/controllers/bulk_ingest_controller_spec.rb
```diff
@@ -60,38 +60,62 @@
...
+        }
+      }
+    end
+
     before do
       allow(IngestFolderJob).to receive(:perform_later)
       allow(IngestFoldersJob).to receive(:perform_later)
     end
 
     context "with one single-volume resource" do
```
hackmastera: Can you update this and/or the test name to mention the bib id dir? To set it in context against line 105.

spec/controllers/bulk_ingest_controller_spec.rb
```diff
@@ -60,38 +60,62 @@
...
-        {
-          "0" => {
-            "url" => "file:///base/4609321/1.tif",
-            "file_name" => "1.tif",
-            "file_size" => "100"
-          }
-        }
+    end
+
+    context "when the directory looks like a bibid, but isn't valid" do
```
hackmastera: This one is a replacement for the "June 31" test that was removed in the other spec?

escowles: Yes, though the scenario is a little different (invalid bibid vs. directory name that doesn't match the pattern).  I'll add an example that uses "June 31" like the old test.

spec/controllers/bulk_ingest_controller_spec.rb
escowles: 👍

escowles:
@hackmastera I've updated the tests to clarify the wording and add back a "June 31" test.

PR #3322, opened 2019-08-20
Support import from GCS structure

### LINKED ISSUE #3297: Allow import of resources from Preservation Storage without a PreservationObject
PreservationObjects get deleted when the resource does, so we should be able to just point at a directory in GCS and import it.

tpendragon:
Closes #3297 

Confirmed that this can import A/V materials as well as ScannedResources.

### DIFF
--- a/app/services/preserver/blind_importer.rb
+++ b/app/services/preserver/blind_importer.rb
@@ -24,7 +24,7 @@ class Preserver::BlindImporter
     member_ids = importing_resource.try(:member_ids) || []
     member_ids.map! do |member_id|
       begin
-        member = self.class.import(id: member_id, importing_metadata_adapter: importing_metadata_adapter.with_parent(parent: importing_resource))
+        member = self.class.import(id: member_id, importing_metadata_adapter: importing_metadata_adapter.with_context(parent: importing_resource))
         importing_change_set.created_file_sets += [member] if member.is_a?(FileSet)
         member.id
       rescue Valkyrie::Persistence::ObjectNotFoundError
--- a/app/services/preserver/blind_importer.rb
+++ b/app/services/preserver/blind_importer.rb
@@ -1,46 +1,48 @@
 # frozen_string_literal: true
 class Preserver::BlindImporter
-  def self.import(id:, change_set_persister:, importing_metadata_adapter: default_importing_metadata_adapter)
-    new(id: id, change_set_persister: change_set_persister, importing_metadata_adapter: importing_metadata_adapter).import!
+  def self.import(id:, change_set_persister:, source_metadata_adapter: default_source_metadata_adapter)
+    new(id: id, change_set_persister: change_set_persister, source_metadata_adapter: source_metadata_adapter).import!
   end
 
-  def self.default_importing_metadata_adapter
-    FileMetadataAdapter.new(storage_adapter: importing_storage_adapter)
+  def self.default_source_metadata_adapter
+    FileMetadataAdapter.new(storage_adapter: source_storage_adapter)
   end
 
-  def self.importing_storage_adapter
+  def self.source_storage_adapter
     Valkyrie::StorageAdapter.find(:google_cloud_storage)
   end
 
-  attr_reader :id, :importing_metadata_adapter, :change_set_persister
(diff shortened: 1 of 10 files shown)


app/services/preserver/blind_importer.rb
```diff
@@ -0,0 +1,66 @@
...
+        member.id
+      rescue Valkyrie::Persistence::ObjectNotFoundError
+        nil
+      end
+    end.compact!
+    output = change_set_persister.save(change_set: importing_change_set)
+    output
+  end
+
+  def import_file_identifiers
```
hackmastera: looks like we're importing the actual files as part of this method,  not just getting the ids. rename?

app/services/preserver/blind_importer.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  attr_reader :id, :importing_metadata_adapter, :change_set_persister
+  delegate :storage_adapter, to: :change_set_persister
+  def initialize(id:, importing_metadata_adapter:, change_set_persister:)
+    @id = id
+    @importing_metadata_adapter = importing_metadata_adapter
+    @change_set_persister = change_set_persister
+  end
+
+  def import!
```
hackmastera: comment that it's recursive might be helpful

app/services/preserver/blind_importer/file_metadata_adapter/query_service.rb
```diff
@@ -0,0 +1,75 @@
...
+    def initialize(adapter:)
+      @adapter = adapter
+    end
+
+    def find_by(id:)
+      file = storage_adapter.find_by(id: storage_id_from_resource_id(id))
+      json = JSON.parse(file.read)
+      attributes = Valkyrie::Persistence::Shared::JSONValueMapper.new(json).result.symbolize_keys
+      resource_processor.call(resource: Valkyrie::Types::Anything[attributes], adapter: adapter)
+    # Rescue NoMethodError because it's what the GCS shrine adapter is throwing
```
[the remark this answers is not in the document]
tpendragon: https://github.com/pulibrary/figgy/issues/3325

tpendragon:
> I need to look at it a bit more to sort out how the context / parent stuff works. maybe some descriptive comments about that at a class level somewhere would be good.

Yeah. I wish I didn't have to do this and could instead attach some kind of hint to the resource itself that the adapter could use, so that rather than an addition to the adapter it could just implement `find_members`.

app/services/preserver/blind_importer/file_metadata_adapter.rb
```diff
@@ -0,0 +1,19 @@
...
+      @parents = parents
+      @resource_processor = resource_processor
+    end
+
+    def query_service
+      @query_service ||= QueryService.new(adapter: self)
+    end
+
+    def with_context(parent:, **_args)
+      self.class.new(storage_adapter: storage_adapter, parents: parents + [parent])
```
hackmastera: why is parents an array? We have a max of 1 parent.

app/services/preserver/blind_importer.rb
```diff
@@ -0,0 +1,68 @@
...
+    @source_metadata_adapter = source_metadata_adapter
+    @change_set_persister = change_set_persister
+  end
+
+  # Imports the given ID from the source metadata adapter as well as all its
+  # children.
+  def import!
+    import_binary_files
+    member_ids = source_resource.try(:member_ids) || []
+    member_ids.map! do |member_id|
```
hackmastera: So after mapping from the figgy ids to the cloud storage ids in the query service, now we're mapping back to figgy ids. 👍

app/services/preserver/blind_importer/file_metadata_adapter/query_service.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+class Preserver::BlindImporter::FileMetadataAdapter
+  class QueryService
```
hackmastera: A typical valkyrie query service will reference an external converter but b/c of the limited scope here it's all rolled into this class. not just in the form of ConvertLocalStorageIDs but also via `Valkyrie::Types::Anything`

hackmastera:
oh, also related to this review I created #3326

PR #3348, opened 2019-09-03
Quiet tests

escowles:
Silences the vast majority of logging, deprecation warnings, etc. output during test suite execution.  There are still a few HoneyBadger warnings about `ActionDispatch::TestProcess`, but everything else is silent.


spec/resources/ephemera_boxes/ephemera_boxes_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 require "rails_helper"
-include ActionDispatch::TestProcess
+include FixtureFileUpload
```
jrgriffiniii: 👍

Issue #3349, opened 2019-09-03
I should be able to restore a resource from its tombstone


escowles:
## Success Criteria

There's a job which can take a resource tombstone and restore it. Should work similarly to FileSet restore.

jrgriffiniii:
I'm currently able to restore a `FileSet`, but without it's `FileMetadata` (please see https://github.com/pulibrary/figgy/compare/issues-3349-jrgriffiniii-restore-tombstones?expand=1).

I can certainly try to reconstruct the `FileMetadata` for the original file from the `PreservationObject` `FileMetadata`, but I just wish to confirm that this would be desireable.

tpendragon:
I think a FileSet is handled by https://github.com/pulibrary/figgy/pull/3351. A resource would probably use the BlindImporter.

jrgriffiniii:
I've started with this, but I needed to get tombstones generated with: https://github.com/pulibrary/figgy/commit/21fc9f28954564fb67ea9a8b46f7d54b8832c651

eliotjordan:
Closed by https://github.com/pulibrary/figgy/pull/5609

Issue #3347, opened 2019-08-30
Resources should create a Tombstone when they're deleted


tpendragon:
## Success Criteria

On delete, a tombstone is made, it includes:

Deleter, deleted_at, resource, and a cache of the preservation object. Propagate functionality from FileSet.

jrgriffiniii:
https://github.com/pulibrary/figgy/pull/3335 seems to have advanced this pretty far.  Should this be generalized for all Figgy `Resources`?

PR #3356, opened 2019-09-04
Ensures that tombstones are generated for deleted ScannedResources and can be used for reinstating ScannedResources

jrgriffiniii:
Advances #3349 (I have not tested this beyond ScannedResources, but that may not be necessary).  This also may relate to #3347.

### DIFF
--- a/app/change_set_persisters/change_set_persister/create_tombstone.rb
+++ b/app/change_set_persisters/change_set_persister/create_tombstone.rb
@@ -7,19 +7,9 @@ class ChangeSetPersister
       @change_set_persister = change_set_persister
     end
 
-    def file_set_with_parent?
-      if resource.is_a?(FileSet)
-        resource.try(:original_file) && parent
-      else
-        file_sets = resource.decorate.try(:file_sets)
-        return if file_sets.blank?
-
-        file_sets.first.try(:original_file)
-      end
-    end
-
     def run
-      return unless file_set_with_parent?
+      return unless resource.is_a?(FileSet) || resource.decorate.respond_to?(:file_sets)
+
       tombstone = Tombstone.new
       tombstone_change_set = DynamicChangeSet.new(tombstone)
       tombstone_change_set.validate(attributes)
@@ -31,10 +21,13 @@ class ChangeSetPersister
       def file_set
         return resource if resource.is_a?(FileSet)
 
-        resource.decorate.file_sets.first
+        file_sets = resource.decorate.file_sets
+        file_sets.first
       end
 
       def original_filename
+        return unless file_set && file_set.original_file
+
         file_set.original_file.original_filename
       end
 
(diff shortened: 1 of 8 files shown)


spec/change_set_persisters/change_set_persister_spec.rb
```diff
@@ -1721,6 +1721,72 @@
...
+        change_set = DynamicChangeSet.new(resource)
+        change_set.validate(state: "complete")
+
+        change_set_persister.save(change_set: change_set)
+
+        change_set = DynamicChangeSet.new(resource)
+        change_set_persister.delete(change_set: change_set)
+
+        tombstones = change_set_persister.query_service.find_all_of_model(model: Tombstone)
+        expect(tombstones).not_to be_empty
```
tpendragon: Can we make some more assertions here to make sure there's just one tombstone and it points to the old resource?

app/services/preserver/blind_importer.rb
```diff
@@ -33,6 +33,14 @@ def import!
     end
     # Get rid of non-preserved members.
     source_change_set.try(:member_ids=, member_ids)
+    # Delete any tombstones
+    source_change_set.created_file_sets.each do |_created_file_set|
```
tpendragon: Why do you iterate over created file sets? It seems like you don't use the result.

app/change_set_persisters/change_set_persister/create_tombstone.rb
```diff
@@ -7,8 +7,19 @@ def initialize(change_set_persister: nil, change_set:, post_save_resource: nil)
...
+        file_sets = resource.decorate.try(:file_sets)
+        return if file_sets.blank?
+
+        file_sets.first.try(:original_file)
+      end
+    end
+
     def run
-      return unless resource.is_a?(FileSet) && resource.try(:original_file) && parent
+      return unless file_set_with_parent?
```
tpendragon: Why do we check for if there are file sets if it's a parent? Wouldn't it just create a tombstone no matter what now?

app/change_set_persisters/change_set_persister/create_tombstone.rb
```diff
@@ -8,7 +8,8 @@ def initialize(change_set_persister: nil, change_set:, post_save_resource: nil)
     end
 
     def run
-      return unless resource.is_a?(FileSet) && resource.try(:original_file) && parent
+      return unless resource.is_a?(FileSet) || resource.decorate.respond_to?(:file_sets)
```
hackmastera: So this would not create a tombstone for any resource whose children are other resources? Like a multi-volume work. is that correct?

tpendragon: I think we can change this to check if it should be preserved. Something like `DynamicChangeSet.new(resource).preserve?`

jrgriffiniii:
#3349 does not appear to be resolved, but I'm wary of letting this PR remain stale.  Would anyone object if I closed this, saved the branch, and resumed progress in 1-2 weeks please?

Issue #3350, opened 2019-09-03
I should be able to list and search for resource tombstones


escowles:
Use cases this interface would have to support:

“I deleted everything in the grounds and buildings collection on accident”
 - implies multi-select or some other notion of bulk restore.
 - Implies that you can search tombstones by what collection they used to be a member of.

“I deleted the book/resource on accident two weeks ago” 
 - Implies the interface should show when it was deleted, let you facet by deleted_at.
 - Implies the interface should show + allow facet on who deleted a thing.
 - Implies if there's a rake task to delete things it should put a NetID in, or there should be a default value.

Options:
 - Index tombstones, default filter them out.
    - Have some way to show deleted items.
    - Bulk edit when filtered to tombstones could bulk restore?

## Success Criteria

Add a "Deleted Resources" facet that will display tombstones only (normally they'd be filtered out.)
Facets for "Deleter", and a start/end date facet for "deleted_at"

[New Ticket] Add a "Bulk Edit" button to restore all tombstones in the search results.

jrgriffiniii:
Does what is requested in this issue require additional UI functionality on the "File Manager" interface (i. e. something comparable to the viewing and list searching capabilities of Recordings for Playlist Views)?

tpendragon:
I don't think so, I think there's probably just a dashboard or something. We can probably wireframe it when we get to this ticket

hackmastera:
FileSet Tombstones will be accessible from the FileManager already. This UI would just be for Resource Tombstones.

Create a dashboard page which presents a searchable datatable. Initially populate with most recently deleted; otherwise max out at N search results. Each row will have a "restore" button.

PR #3357, opened 2019-09-04
Implements a TombstoneIndexer

jrgriffiniii:
Advances #3350


jrgriffiniii:
This is no longer needed given that an approach using DataTables is going to be taken (as outlined in https://github.com/pulibrary/figgy/issues/3350).

PR #3373, opened 2019-09-13
15x parallel just for fun.

tpendragon:


tpendragon:
This lowered test time by like, 1 minute.

PR #3401, opened 2019-09-30
Use global variable to store order manager fetching state

### LINKED ISSUE #3399: Order Manager Apply Changes Button doesn't unlock after a save
To Reproduce:

1. Go into order manager, edit label, "Apply Changes"
2. Edit label again, "Apply Changes" will be disabled (stop sign symbol when you try to click it)

Reported by Dan. Current workaround is to only do one set of changes and then reload, but he's scared to do a bunch of work that's wasted.

eliotjordan:
Order manager fetch state is never returned to true.

The context of `this` in the `isDisabled` method:

https://github.com/pulibrary/figgy/blob/f4d6b17e6c0dc422c8e541a97727fb7a43376cf9/app/javascript/components/OrderManagerControls.vue#L100

is different from the context of `this` in the `SAVED_STATE` method:
https://github.com/pulibrary/figgy/blob/f4d6b17e6c0dc422c8e541a97727fb7a43376cf9/app/javascript/store/resource/index.js#L36

The result is that the fetching state is never returned to `false`. This PR uses a global variable to store this state. 

Closes #3399

### DIFF
--- a/app/javascript/components/OrderManagerControls.vue
+++ b/app/javascript/components/OrderManagerControls.vue
@@ -97,7 +97,7 @@ export default {
       return this.resource.isMultiVolume
     },
     isDisabled: function () {
-      if (this.$store.getters.stateChanged && !this.fetching) {
+      if (this.$store.getters.stateChanged && !window.fetching) {
         return false
       } else {
         return true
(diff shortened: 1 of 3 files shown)


eliotjordan:
`this` is in the first case refers to `VueComponent`, while in the other case it refers to an `Observer` object.

PR #3412, opened 2019-10-03
Use VIPS to remove compression from Tiffs

tpendragon:
It's more memory efficient, we're hoping it fixes the problem with large
TIFFs.

We'll need to install libvips and libvips-dev on the machines if we
deploy this.


tpendragon:
Disk space is the problem, so I'm closing this.

PR #3415, opened 2019-10-04
Add cap [staging/production] solr:console

tpendragon:
No more ssh -L for opening Solr console.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -123,3 +123,19 @@ namespace :rabbitmq do
     end
   end
 end
+
+namespace :solr do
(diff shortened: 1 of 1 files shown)


config/deploy.rb
```diff
@@ -123,3 +123,19 @@
     end
   end
 end
+
+namespace :solr do
+  desc 'Opens RabbitMQ Console'
```
hackmastera: here and lower down rabbitmq needs to say solr

tpendragon:
@hackmastera Fixed

hackmastera:
🎉

PR #3433, opened 2019-10-14
Changes target of Hathi SIP from a zip file to a directory.

cwulfman:
The zip files were too big to handle, so this PR drops the zipping and just creates a directory in which to deposit the files in the SIP.

### DIFF
--- a/app/exporters/hathi/submission_information_package.rb
+++ b/app/exporters/hathi/submission_information_package.rb
@@ -55,8 +55,8 @@ module Hathi
 
       def deposit_checksums
         File.open(File.join(@export_dir, "checksum.md5"), "w") do |f|
-          # iterate over k,v in checksums
-          checksums.each { |k, v| f.write format("%s %s\n", v, k) }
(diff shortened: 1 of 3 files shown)


app/exporters/hathi/submission_information_package.rb
```diff
@@ -12,52 +14,60 @@ def initialize(package:, base_path:)
...
-        zipfile.get_output_stream("checksum.md5") do |f|
-          # iterate over k,v in checksums
-          checksums.each { |k, v| f.write format("%s %s\n", v, k) }
-        end
+    end
+    
+    def deposit_checksums()
+      File.open(File.join(@export_dir, "checksum.md5"), "w") do |f|
+        # iterate over k,v in checksums
+        checksums.each { |k, v| f.write format("%s %s\n", v, k) }
```
escowles: Out of curiosity, does Hathi allow more than one space in the checksum file? If they allow it, I'd recommend using two spaces, since that allows the standard Unix `md5sum` command to verify the checksums.

PR #3434, opened 2019-10-14
Add External Contributors to Ephemera Projects.

### LINKED ISSUE #3398: Contributor Role for Ephemera Projects
A user with the Contributor Role should be able to:

1. View/Create/Edit Ephemera Folders/Boxes for their Projects.
2. Not be able to see private materials outside of their Project.
3. Be unable to delete in all cases

Implementation Notes:

Three options:

1. Users have a list of Projects they're a member of.
2. Ephemera Projects have a list of "contributor_ids" which are UserIDs. Being part of this list gives one the "contributor" role.
3. Add a "scope" column to the Role model, so that a user can be a contributor for a specific subset of Projects.

Prefer Option 2, but report back if it's difficult.

Caveats:

Implement without worrying about Solr Discovery for now, as it's likely to be difficult and may not be necessary. The majority of the Ephemera use cases are handled via the dashboard, which is solely powered by Postgres. If this turns out to be a problem, we can look more into this.

tpendragon:
Closes #3398

### DIFF
--- a/app/controllers/concerns/resource_controller.rb
+++ b/app/controllers/concerns/resource_controller.rb
@@ -15,11 +15,14 @@ module ResourceController
       redirect_to new_scanned_resource_path
     else
       @change_set = change_set_class.new(new_resource, append_id: params[:parent_id]).prepopulate!
-      # If there's a parent, authorize updating it instead.
       authorize_create!(change_set: @change_set)
     end
   end
 
+  # For new/create if a resource is going to be appended to a parent, the
+  # permissions should be based on the ability to update the parent it's going
+  # to be appended to. Enables users who only have permission to add to a single
+  # Ephemera Project.
   def authorize_create!(change_set:)
     if change_set.append_id.present?
       authorize! :update, query_service.find_by(id: Array(change_set.append_id).first)
--- a/app/controllers/concerns/resource_controller.rb
+++ b/app/controllers/concerns/resource_controller.rb
@@ -15,6 +15,15 @@ module ResourceController
       redirect_to new_scanned_resource_path
     else
       @change_set = change_set_class.new(new_resource, append_id: params[:parent_id]).prepopulate!
+      # If there's a parent, authorize updating it instead.
+      authorize_create!(change_set: @change_set)
+    end
+  end
+
+  def authorize_create!(change_set:)
+    if change_set.append_id.present?
+      authorize! :update, query_service.find_by(id: Array(change_set.append_id).first)
+    else
       authorize! :create, resource_class
     end
   end
@@ -25,8 +34,9 @@ module ResourceController
 
   def create
     @change_set = change_set_class.new(resource_class.new)
-    authorize! :create, @change_set.resource
-    if @change_set.validate(resource_params.merge(depositor: [current_user.uid]))
+    @change_set.validate(resource_params.merge(depositor: [current_user&.uid]))
+    authorize_create!(change_set: @change_set)
+    if @change_set.valid?
       @change_set.sync
       obj = nil
       change_set_persister.buffer_into_index do |buffered_changeset_persister|
--- a/app/resources/ephemera_projects/ephemera_project.rb
+++ b/app/resources/ephemera_projects/ephemera_project.rb
@@ -13,6 +13,6 @@ class EphemeraProject < Resource
 
   # Append contributor_uids to edit_users
   def edit_users
-    self[:edit_users] | contributor_uids
+    (self[:edit_users] + contributor_uids).uniq
   end
 end
--- a/app/resources/ephemera_projects/ephemera_project.rb
+++ b/app/resources/ephemera_projects/ephemera_project.rb
@@ -5,6 +5,7 @@ class EphemeraProject < Resource
   attribute :title, Valkyrie::Types::Set
   attribute :slug, Valkyrie::Types::Set
   attribute :top_language, Valkyrie::Types::Set
+  attribute :contributor_uids, Valkyrie::Types::Set
 
   def logical_structure
     []
--- a/app/resources/ephemera_projects/ephemera_project.rb
+++ b/app/resources/ephemera_projects/ephemera_project.rb
@@ -10,4 +10,9 @@ class EphemeraProject < Resource
   def logical_structure
     []
   end
+
+  # Append contributor_uids to edit_users
+  def edit_users
+    self[:edit_users] | contributor_uids
+  end
 end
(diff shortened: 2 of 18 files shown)


app/resources/ephemera_projects/ephemera_project.rb
```diff
@@ -5,8 +5,14 @@ class EphemeraProject < Resource
...
   attribute :top_language, Valkyrie::Types::Set
+  attribute :contributor_uids, Valkyrie::Types::Set
 
   def logical_structure
     []
   end
+
+  # Append contributor_uids to edit_users
+  def edit_users
+    self[:edit_users] | contributor_uids
```
escowles: I was just saying yesterday that I'd never actually used `|` instead of `||` — is there a reason for `|` here?

hackmastera: @escowles I just looked it up; it's defined on array as a set union https://ruby-doc.org/core-2.4.1/Array.html#method-i-7C

escowles: @hackmastera 👍 thanks for looking that up, I was misreading it.  Now I see that it's a union of the edit users and contributor users.

spec/models/ability_spec.rb
```diff
@@ -80,6 +80,13 @@
   let(:other_staff_file) { other_staff_scanned_resource.decorate.members.first }
   let(:admin_file) { FactoryBot.build(:file_set, user: admin_user) }
 
+  let(:contributor_ephemera_project) do
+    FactoryBot.create_for_repository(:ephemera_project, member_ids: [contributor_ephemera_folder.id, contributor_ephemera_box.id], contributor_uids: [current_user&.uid])
+  end
+  let(:contributor_ephemera_folder) { FactoryBot.create_for_repository(:ephemera_folder) }
+  let(:contributor_ephemera_box) { FactoryBot.create_for_repository(:ephemera_box, member_ids: contributor_ephemera_folder_in_box.id) }
+  let(:contributor_ephemera_folder_in_box) { FactoryBot.create_for_repository(:ephemera_folder) }
+
```
escowles: The additions here all make sense.  But I was expecting to see a set of tests that verified that an external contributor could create/update the folders in their project, but not create/update those in a different project, and also that they couldn't see a private non-ephemera item.

app/controllers/concerns/resource_controller.rb
```diff
@@ -15,6 +15,15 @@ def new
       redirect_to new_scanned_resource_path
     else
       @change_set = change_set_class.new(new_resource, append_id: params[:parent_id]).prepopulate!
+      # If there's a parent, authorize updating it instead.
```
hackmastera: Maybe move this comment to the method itself, adding a tiny bit more about why. I think elsewhere in this PR the logic exists that if the parent can be edited then the folder can be created, but it's not clear here that the update authorization also confers the create authorization.

spec/models/ability_spec.rb
```diff
@@ -297,6 +312,25 @@
       is_expected.not_to be_able_to(:discover, pending_scanned_resource)
       is_expected.not_to be_able_to(:discover, reading_room_scanned_resource)
       is_expected.to be_able_to(:discover, campus_ip_scanned_resource)
+
+      # Ephemera Project Contributors
```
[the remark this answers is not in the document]
escowles: Yes, they will have NetIDs that are their external email addresses.

app/resources/ephemera_projects/ephemera_project.rb
tpendragon: Sorry, a math background makes me reach for set operators. 😆 I can make this `+`

spec/models/ability_spec.rb
tpendragon: I can add a set of tests for a different project. There's already tests for non-ephemera items - all the other tests under campus users weren't changed.

spec/resources/ephemera_folders/ephemera_folders_controller_spec.rb
```diff
@@ -44,6 +44,19 @@
   describe "new" do
     it_behaves_like "an access controlled new request"
 
+    context "when they have permission to edit the parent" do
```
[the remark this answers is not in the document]
tpendragon: This seems like it's already covered by the regular "doesn't have permission" tests to me.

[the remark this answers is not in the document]
hackmastera: I tried to find that... :( big test I guess.

spec/views/catalog/_admin_controls_ephemera_box.html.erb_spec.rb
```diff
@@ -25,4 +27,11 @@
       expect(rendered).to have_link "Attach Hard Drive", href: attach_drive_ephemera_box_path(box.id)
     end
   end
+  context "as an external contributor" do
+    let(:user) { FactoryBot.create(:user) }
+    let(:project) { FactoryBot.create_for_repository(:ephemera_project, member_ids: box.id, contributor_uids: user.uid) }
+    it "displays a button to attach an Ephemera Folder" do
```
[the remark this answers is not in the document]
tpendragon: We can. We never had that test before though - it's just a side effect of not being an admin. Do you still want me to add it?

app/controllers/concerns/resource_controller.rb
tpendragon: Oh, yeah. This is just left behind from before a refactor. 😅

app/resources/ephemera_projects/ephemera_project.rb
hackmastera: @tpendragon the set operator deduplicates, which seems like a nice safeguard. Will `contributor_uids` never get saved to `edit_users`?

spec/resources/ephemera_folders/ephemera_folders_controller_spec.rb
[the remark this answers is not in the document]
tpendragon: I do. Is there something I'm missing?

app/resources/ephemera_projects/ephemera_project.rb
tpendragon: Yes, I know it de-dups. I was going to just add `.uniq` at the end. :)

spec/resources/ephemera_folders/ephemera_folders_controller_spec.rb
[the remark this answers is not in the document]
hackmastera: Works for me!

spec/views/catalog/_admin_controls_ephemera_box.html.erb_spec.rb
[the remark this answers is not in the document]
hackmastera: nah, not if it's unrelated.

PR #3436, opened 2019-10-16
Preserve everything

### LINKED ISSUE #3283: Preserve everything except Music Reserves
This should be the default case, and we need to go back and preserve everything that hasn't been yet.

There could be a couple of different ways to do this, the important parts are:
1. Don't preserve Audio Reserves objects (new or future)
2. Other objects should be preserved by default
3. All current non-Audio Reserves objects should be preserved if they haven't been already

eliotjordan:
- Makes preservation the default for resources.
- Removes the `preservation_policy` property to simplify implementation.
- Adds test to ensure collections do not preserve members.

Closes #3283

### DIFF
--- a/spec/change_set_persisters/change_set_persister_spec.rb
+++ b/spec/change_set_persisters/change_set_persister_spec.rb
@@ -1532,7 +1532,7 @@ RSpec.describe ChangeSetPersister do
   end
 
   describe "preservation" do
-    context "when a completed resource is updated with a `cloud` preservation_policy outside of the change_set_persister" do
+    context "when a completed preserved resource is updated outside of the change_set_persister" do
       it "backgrounds any child preserving" do
         allow(Preserver).to receive(:new).and_call_original
         allow(PreserveChildrenJob).to receive(:perform_later).and_call_original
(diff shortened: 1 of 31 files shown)


eliotjordan:
Next steps are to build the preservation logic for music reserves.

spec/change_set_persisters/change_set_persister_spec.rb
```diff
@@ -1541,21 +1541,20 @@
...
         file_set.read_groups = []
-        resource.preservation_policy = "cloud"
         resource = change_set_persister.metadata_adapter.persister.save(resource: resource)
         change_set_persister.metadata_adapter.persister.save(resource: file_set)
         change_set = DynamicChangeSet.new(resource)
 
         change_set_persister.save(change_set: change_set)
         expect(PreserveChildrenJob).to have_received(:perform_later).exactly(1).times
-        expect(Preserver).to have_received(:new).exactly(2).times
+        expect(Preserver).to have_received(:new).exactly(5).times
```
eliotjordan: One consequence of this preserve everything approach is that preservation is triggered multiple times for FileSets added to completed resources. Once for characterization and at least once for derivative generation.

Calls to Preserver service in this test:
1. Saving the scanned resource as a completed object
1. Characterizing the attached file
1. Generating the derivative for the attached file
1. Saving the scanned resource again
1. Saving the FileSet again

tpendragon: That should be rare, and only result in re-saving the JSON file yeah?

eliotjordan: I'd say it's probably rare. Just wanted to point it out.

eliotjordan:
Updated with logic to not preserve Audio Reserves.

PR #3453, opened 2019-10-29
Puts jp2 derivatives instead of tiffs into the SIP.

cwulfman:


app/exporters/hathi/submission_information_package.rb
```diff
@@ -29,10 +29,12 @@ def export
       def deposit_files
         package.pages.each do |page|
           digester.reset
-          FileUtils.cp page.tiff_path,
-                       File.join(@export_dir, page.name + ".tif")
-          digester.file page.tiff_path
-          checksums[page.name + ".tif"] = digester.hexdigest
+          # FileUtils.cp page.tiff_path,
```
tpendragon: Can you delete the commented code?

cwulfman: done.

PR #3461, opened 2019-10-31
Merge branch 'master' into remove-zip

cwulfman:


escowles:
@cwulfman I think we can close this PR — the substantive changes (using a directory instead of a zip file) were merged already, and we don't really need to document changes in the command-line tools with release notes entries.

PR #3468, opened 2019-11-04
Viewer uses dynamic config instead of requiring each embedding app to set it themselves

### LINKED ISSUE #3464: Downloads still available even when set to "none"
The viewer for this item is still showing a download button and allowing downloads when Downloads is set to "none" on the edit page:

https://findingaids.princeton.edu/collections/C1567/c2
https://figgy.princeton.edu/catalog/60a951da-3b3d-4cc3-899d-8b89d64e6c83

escowles:
Fixes #3464


tpendragon:
Seems fine to me.

PR #3470, opened 2019-11-05
addresses comments by Hathi about meta.yml

cwulfman:
This patch addresses comments from HathiTrust about our sample meta.yml file.


app/exporters/hathi/image_metadata.rb
```diff
@@ -2,47 +2,49 @@
...
+      original_image["%z"] == 1
     end
 
     def resolution
-      repr_image["resolution"].first
+      original_image["resolution"].first
     end
 
     def scanning_order
-      "left_to_right"
```
escowles: We do have some right-to-left items, so maybe checking the `viewing_direction` property would be good here?

app/exporters/hathi/image_metadata.rb
```diff
@@ -2,47 +2,49 @@
...
+      "left-to-right"
     end
 
     def reading_order
-      "left_to_right"
```
escowles: Ditto about `viewing_direction`

app/exporters/hathi/image_metadata.rb
cwulfman: Here is Hathi's sample meta.yaml file: https://drive.google.com/file/d/0B0EHs5JWGUMLWjU2OHVhQzN5WEk/view

PR #3472, opened 2019-11-07
Revert "Prohibit anonymous downloads when a resource is not publicly downloadable"

hackmastera:
needs to be tested on staging.
This reverts commit cbf748712d4997ec136cc150c386f12b62393865.


hackmastera:
close in favor of #3478

PR #3476, opened 2019-11-07
Updated the Gemfile to pin to Webpacker 4.0.z releases

jrgriffiniii:
Figgy was previously using a prerelease version of `Webpacker`, and this triggered errors in certain development environments.


hackmastera:
Does webpacker use semver? If so it seems like we should be able to leave it at `"~> 4.0"` but still update the `Gemfile.lock` to the preferred version.

PR #3484, opened 2019-11-11
Ensures that PendingUpload constructors receive Hashes with symbolized keys

jrgriffiniii:
This was triggering failures for BrowseEverything uploads, as the file metadata was not actually being stored in the PendingUpload Resource.

### DIFF
--- a/app/controllers/bulk_ingest_controller.rb
+++ b/app/controllers/bulk_ingest_controller.rb
@@ -163,7 +163,7 @@ class BulkIngestController < ApplicationController
 
       @new_pending_uploads = []
       selected_files.each do |selected_file|
-        file_attributes = selected_file.to_h
+        file_attributes = selected_file.to_h.symbolize_keys
 
         # This is needed in order to ensure that files are authorized for the
         # download from the Cloud Service provider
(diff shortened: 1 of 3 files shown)


app/controllers/bulk_ingest_controller.rb
```diff
@@ -163,7 +163,7 @@ def new_pending_uploads
 
       @new_pending_uploads = []
       selected_files.each do |selected_file|
-        file_attributes = selected_file.to_h
+        file_attributes = selected_file.to_h.symbolize_keys
```
tpendragon: Can you update the test to pass strings?

jrgriffiniii: I am sorry for that, https://github.com/pulibrary/figgy/blob/master/spec/controllers/bulk_ingest_controller_spec.rb#L201 passes the keys as strings, and I now check to ensure that these are symbolized with https://github.com/pulibrary/figgy/pull/3484/files#diff-1248e98671a3d761ac436ef1b0447f78R239

jrgriffiniii:
Apologies for the failing tests, this should be ready for another review please.

Issue #3488, opened 2019-11-11
Delete resources stuck in a bad "pending uploads" state


tpendragon:
[Figgy/production] ActionView::Template::Error: undefined method `first' for nil:NilClass

## Backtrace

    line 9 of [PROJECT_ROOT]/app/models/pending_upload.rb: original_filename
    line 13 of [PROJECT_ROOT]/app/views/base/_pending_uploads.html.erb: block in _app_views_base__pending_uploads_html_erb__3568314398988254467_70184313985700
    line 10 of [PROJECT_ROOT]/app/views/base/_pending_uploads.html.erb: each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/57256942)

Resource https://figgy.princeton.edu/catalog/3caad915-5768-4d51-ad21-f85f944105e8 says it has a pending upload but jobs queue is empty.

hackmastera:
This was a result of #3487

hackmastera:
There are a bunch of these, looks like 4 clumps of 30-ish so ~120.

hackmastera:
I think we can find all these with `qs.custom_queries.find_by_property(property: "title", value: nil)`; haven't confirmed that retrieves the one in question, though. Also not sure if it will retrieve more than we're looking for. may need to further limit by created today.

tpendragon:
Querying for nil with that query doesn't work unfortunately. It finds everything, weird side effect of the contains operator.

hackmastera:
😆 no wonder it got hung up. I guess we should just look for resources with the pending uploads status. Maybe use read-only mode to ensure nothing is actually being ingested.

jrgriffiniii:
Resources which have `PendingUploads` as members have these removed with https://github.com/pulibrary/figgy/blob/master/app/services/file_appender.rb#L175. An error should be raised with https://github.com/pulibrary/figgy/blob/master/app/services/file_appender.rb#L68 in order to prevent this from occurring, but this is not occurring (hence, while HoneyBadger notifications do not report these failures).

I would propose that `PendingUploads` should not be removed if there are such failures, and such failures should be explicitly logged.

hackmastera:
I think these bad resources are still there; they show up in a blank search result if you page back far enough. I'm not sure how to retrieve them, though, and I'm not sure how far back you'd have to page at this point.

tpendragon:
We haven't had any new ones since 2020, and most of them were from the same date. I did this to get rid of them:

[...]

PR #3493, opened 2019-11-12
Implements a CleanPendingUploadsJob, a FindPendingUploadsFailures query, and the Rake task clean:pending_uploads

jrgriffiniii:
These provide one with the ability to delete resources which have failed file uploads as members. Introduced in response to #3488

### DIFF
--- a/app/jobs/clean_pending_uploads_job.rb
+++ b/app/jobs/clean_pending_uploads_job.rb
@@ -6,11 +6,11 @@ class CleanPendingUploadsJob < ApplicationJob
         if resource.respond_to?(:pending_uploads) && resource.pending_uploads.empty? && resource.state.include?("pending")
           change_set = DynamicChangeSet.new(resource)
           if dry_run
-            Valkyrie.logger.info("Found #{resource.id} as an uploaded resource without any FileSets - this would normally be deleted by CleanPendingUploadsJob")
+            Valkyrie.logger.info("Found #{resource.decorate.first_title} (#{resource.id}) as an uploaded resource without any FileSets - this would normally be deleted by CleanPendingUploadsJob")
           else
             buffered_changeset_persister.delete(change_set: change_set)
+            Valkyrie.logger.info "Deleted a resource with failed uploads with the title: #{resource.decorate.first_title} (#{resource.id})"
           end
-          Valkyrie.logger.info "Deleted a resource with failed uploads with the ID: #{resource.id}"
         end
       end
     end
--- a/app/jobs/clean_pending_uploads_job.rb
+++ b/app/jobs/clean_pending_uploads_job.rb
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
 class CleanPendingUploadsJob < ApplicationJob
-  def perform
+  def perform(dry_run: false)
     query_service.custom_queries.find_pending_upload_failures.each do |resource|
       change_set_persister.buffer_into_index do |buffered_changeset_persister|
         if resource.respond_to?(:pending_uploads) && resource.pending_uploads.empty? && resource.state.include?("pending")
           change_set = DynamicChangeSet.new(resource)
-          buffered_changeset_persister.delete(change_set: change_set)
+          if dry_run
+            Valkyrie.logger.info("Found #{resource.id} as an uploaded resource without any FileSets - this would normally be deleted by CleanPendingUploadsJob")
+          else
+            buffered_changeset_persister.delete(change_set: change_set)
+          end
           Valkyrie.logger.info "Deleted a resource with failed uploads with the ID: #{resource.id}"
         end
       end
--- a/app/jobs/clean_pending_uploads_job.rb
+++ b/app/jobs/clean_pending_uploads_job.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+class CleanPendingUploadsJob < ApplicationJob
+  def perform
+    query_service.custom_queries.find_pending_upload_failures.each do |resource|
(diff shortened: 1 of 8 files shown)


jrgriffiniii:
> Can we add a check so that we don't delete anything that has filesets? All the ones we want to delete have pending up loads but no file sets; I'm not sure what else may be out there with stale pending uploads.

I have added an additional test for this at https://github.com/pulibrary/figgy/pull/3493/files#diff-242bcc50b5849636074e348c21bbb49bR56

jrgriffiniii:
> And maybe a dry-run flag that just returns them so the task can list them by title and date modified?

This was added with https://github.com/pulibrary/figgy/pull/3493/files#diff-242bcc50b5849636074e348c21bbb49bR73

app/jobs/clean_pending_uploads_job.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+class CleanPendingUploadsJob < ApplicationJob
+  def perform(dry_run: false)
+    query_service.custom_queries.find_pending_upload_failures.each do |resource|
+      change_set_persister.buffer_into_index do |buffered_changeset_persister|
+        if resource.respond_to?(:pending_uploads) && resource.pending_uploads.empty? && resource.state.include?("pending")
+          change_set = DynamicChangeSet.new(resource)
+          if dry_run
+            Valkyrie.logger.info("Found #{resource.id} as an uploaded resource without any FileSets - this would normally be deleted by CleanPendingUploadsJob")
```
hackmastera: Can we also print the decorated title?

app/jobs/clean_pending_uploads_job.rb
```diff
@@ -0,0 +1,32 @@
...
+          else
+            buffered_changeset_persister.delete(change_set: change_set)
+          end
+          Valkyrie.logger.info "Deleted a resource with failed uploads with the ID: #{resource.id}"
```
hackmastera: This should be inside the else

lib/tasks/clean.rake
```diff
@@ -16,4 +16,10 @@ namespace :clean do
   task :dead_queues do
     CleanDeadQueuesJob.set(queue: :low).perform_later
   end
+
+  desc "Clean failed PendingUpload Resources."
+  task :pending_uploads, [:dry_run] => [:environment] do |_t, args|
+    args.with_defaults(dry_run: false)
+    CleanPendingUploadsJob.set(queue: :low).perform_later(dry_run: args[:dry_run].present?)
```
hackmastera: Perhaps the job should take a logger and dry-run should do perform_now with a stdout logger.

app/queries/find_pending_upload_failures.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  attr_reader :query_service
+  delegate :resources, to: :query_service
+  delegate :resource_factory, to: :query_service
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def find_pending_upload_failures
+    relation = { "pending_uploads" => [], "state" => ["pending"] }
```
jrgriffiniii: This should be querying for cases where the `file` uploads have been removed (https://github.com/pulibrary/figgy/blob/master/app/services/file_appender.rb#L175), but the FileSet itself has not had the necessary `FileMetadata` nodes added.

jrgriffiniii:
https://github.com/pulibrary/figgy/blob/master/app/services/file_appender.rb#L5 and https://github.com/pulibrary/figgy/blob/master/app/services/file_appender.rb#L8 should be raised within `FileAppender`, but I'm not certain where it is.

app/queries/find_pending_upload_failures.rb
jrgriffiniii: https://app.honeybadger.io/projects/53391/faults/57256942 shows that https://github.com/pulibrary/figgy/blob/master/app/models/pending_upload.rb#L13 raises an error, but https://github.com/pulibrary/figgy/blob/master/app/services/file_appender.rb#L178 should remove `PendingUploads` given that https://github.com/pulibrary/figgy/blob/master/app/services/file_appender.rb#L112 captures and logs errors. This query should probably just query for Resources and update the `pending_uploads` attribute when they are missing the `file_name` attribute.

jrgriffiniii: `PendingUpload` seems to have been persisted with a random UUID for some years (https://github.com/pulibrary/figgy/blame/dd40b22a78e011b18d942364297d271eca1f605b/app/controllers/base_resource_controller.rb#L34), so I will proceed with trying to just query for cases where these have been persisted with invalid metadata.

jrgriffiniii:
https://github.com/pulibrary/figgy/pull/3493/commits/951909cdd93023238586f4f770c0c518c7dabb38 and https://github.com/pulibrary/figgy/pull/3493/commits/2cd14d927dd2d11d9d58b22fcbbe574d9f77626c ensure that `pending_uploads` on Resources are just deleted, while the Resources themselves remain persisted. `pending_uploads` will *not* be deleted for cases where the `file_name` on the `PendingUpload` is not `nil`.

tpendragon:
We've redone how pending uploads work since this PR, and the issue it was solving hasn't cropped up again. Closing.

tpendragon:
@jrgriffiniii @hackartisan Anyone remember why I reopened this?

hackartisan:
There are a bunch of messed up resources. The issue is still valid. Last comment: https://github.com/pulibrary/figgy/issues/3488#issuecomment-648346651 -- we can talk about this on a Monday if you like.

bess:
Looked at this in team meeting -- look again in August 2021

PR #3494, opened 2019-11-13
Enable test coverage calculation by default

escowles:


hackmastera:
Interesting. Is this related to the segfault problem? Can we just set this in the test environment for the app?

escowles:
I don't remember why we made that optional — but it's been like that since the beginning of Figgy: https://github.com/pulibrary/figgy/blob/master/spec/spec_helper.rb#L2

I would be fine with just making that happen by default instead of having a conditional.

tpendragon:
It's a part of the suspenders default setup. We should just remove the conditional

PR #3525, opened 2019-12-10
Hathi metadata refinements

cwulfman:
Based on feedback from HathiTrust, this PR modifies the format of some fields in `meta.yml`; it also generates a `pagedata` field of labels for each page.


escowles:
This looks fine to me — I just restarted Circle because the test failure looks like it's unrelated to the changes here.

cwulfman:
Withdrawing this PR and replacing it with #3578

PR #3540, opened 2019-12-19
Deposit files error

cwulfman:
Tests for ocr? and hocr? were broken because not dealing with empty string.

### DIFF
--- a/spec/exporters/hathi/content_package_spec.rb
+++ b/spec/exporters/hathi/content_package_spec.rb
@@ -104,6 +104,17 @@ RSpec.describe Hathi::ContentPackage do
       expect(page.derivative_path.ftype).to eq "file"
     end
 
+    it "has a copy of the original image" do
+      image = package.pages.first.original_image
+      expect(image.mime_type).to eq "image/tiff"
+    end
+
+    # fix this: figure out how to get FactoryBot to create derivatives
+    it "has a copy of the derivative image" do
+      image = package.pages.first.derivative_image
+      expect(image.mime_type).to eq "image/tiff"
+    end
+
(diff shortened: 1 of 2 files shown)


spec/exporters/hathi/content_package_spec.rb
```diff
@@ -104,6 +104,17 @@
       expect(page.derivative_path.ftype).to eq "file"
     end
 
+    it "has a copy of the original image" do
+      image = package.pages.first.original_image
+      expect(image.mime_type).to eq "image/tiff"
+    end
+
+    # fix this: figure out how to get FactoryBot to create derivatives
```
escowles: I believe you can use the `run_real_derivatives` directive like https://github.com/pulibrary/figgy/blob/master/spec/adapters/bagit/bag_exporter_spec.rb#L19

cwulfman: Not sure what it's doing; the mime_type is still image/tiff, but I'd expect it to be image/jp2

escowles:
@cwulfman That's very strange — I assume it's generating derivatives correctly when you run it instead of them being TIFFs?

cwulfman:
That's right; they seem to be jp2s.

hackmastera:
it's probably https://github.com/pulibrary/figgy/blob/master/spec/support/stub_tika.rb#L23

cwulfman:
Well that would certainly explain it.  Do we care about this enough to start mucking around with this mocking code?  Or is there another way?

escowles:
I don't think it's important to test that here, I think you can just remove the mime type check and this is fine.

PR #3548, opened 2019-12-31
Set node version to match production

hackmastera:
Currently my dev environment is messed up in the javascript. But I do think we want this. Perhaps someone whose lux is working correctly locally could pull this and confirm it's good.


jrgriffiniii:
I'm not certain if this is related (it may well be just my dev. environment), but I'm experiencing some unexpected behavior with LUX components when switching to use this Node version. I will look to diagnose this.

hackmastera:
@jrgriffiniii I was looking at the lux readme yesterday and it says node 10.16 is required to develop. I'm not sure what's required to run it, though. We don't seem to be having problems on staging / prod... but I wonder if we ought to update them.

jrgriffiniii:
I was just looking at trying to work with LTS 13.5.0 (given what I'm seeing for https://nodejs.org/en/about/releases/), but this breaks with our version of `node-sass`.

hackmastera:
Yeah, that sounds like the problem @sdellis hit yesterday. maybe i should just close this until we get to #3556

jrgriffiniii:
I'm also trying to determine why precisely `fsevents` and `node-gyp` are breaking for me with earlier LTS releases. If sdellis was encountering these or similar problems, then I can postpone looking for a solution for this.

hackmastera:
@jrgriffiniii He didn't mention those. It's up to you! It seems like it's something we'll need to get a handle on but I'm not sure how much time you want to put toward it! Thanks for looking at it, regardless.

jrgriffiniii:
I'm really struggling to find a more elegant solution than `yarn upgrade`, but it does work with LTS Erbium (currently 12.14.0), and restores my build of LUX locally (albeit, with some warnings):

```
WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely chan
ge in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the foll
owing commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3

⚠ ｢wdm｣: Hash: 3d98b3966f6cbef2f45c
Version: webpack 4.41.5
Time: 18627ms
Built at: 01/03/2020 10:26:22 AM
                                             Asset       Size               Chunks                         Chunk Names
            js/application-1b8985bf8e8101757ace.js   3.13 MiB          application  [emitted] [immutable]  application
        js/application-1b8985bf8e8101757ace.js.map   3.45 MiB          application  [emitted] [dev]        application
    js/auth_link_clipboard-a7dc74c5ec6143199b54.js    422 KiB  auth_link_clipboard  [emitted] [immutable]  auth_link_clipboard
js/auth_link_clipboard-a7dc74c5ec6143199b54.js.map    477 KiB  auth_link_clipboard  [emitted] [dev]        auth_link_clipboard
                                     manifest.json  759 bytes                       [emitted]

WARNING in ./app/javascript/components/OrderManagerResourceForm.vue
Module Warning (from ./node_modules/eslint-loader/index.js):

...figgy/app/javascript/components/OrderManagerResourceForm.vue
  14:38  warning  Directive "if" should be on a new line               vue/max-attributes-per-line
  14:38  warning  Attribute "v-if" should go before "variation"        vue/attributes-order
  15:1   warning  Expected indentation of 8 spaces but found 0 spaces  vue/html-indent

✖ 3 problems (0 errors, 3 warnings)
  0 errors and 3 warnings potentially fixable with the `--fix` option.

ℹ ｢wdm｣: Compiled with warnings.
```

jrgriffiniii:
I suspect that https://github.com/pulibrary/figgy/pull/3557/files#diff-8ee2343978836a779dc9f8d6b794c3b2 addresses the `yarn upgrade`, I will review that first

tpendragon:
Closing because we've got #3752 now.

PR #1646, opened 2018-08-06
[WIP/POC] Try optlocks

hackmastera:
Add optimistic locking to scanned resources

resolves samvera-labs/valkyrie#458

PR #3549, opened 2019-12-31
[WIP - do not merge] Optimistic locking on all resources

hackartisan:
This PR pulls out and builds on work done in https://github.com/pulibrary/figgy/pull/1646 to extend optimistic locking to all resources that are updated via forms.

It's not quite ready yet as graphql / order manager functionality needs to be reviewed and probably patched up.

ChangeSets were updated if they inherited from`ChangeSet` or their corresponding resource inherited from `Resource`, unless the resource had no controller (`PreservationObject`, `DateRange`, and `WorkflowNote`). Resources inheriting from `Valkyrie::Resource` whose change set inherit from `Valkyrie::ChangeSet` were mostly internal-use resources or nested resources; these do not have optimistic locking turned on.


tpendragon:
@hackartisan Can I close this?

PR #3560, opened 2020-01-03
Removing lib-proc4 and lib-proc5 from the deployment hosts

jrgriffiniii:
This is needed in order to restore production deployments from the `master` branch


jrgriffiniii:
@hackmastera I'm sorry for that, the note has been removed

PR #3564, opened 2020-01-06
Fix slow query on Catalog Show Page.

tpendragon:
This index speeds up the query from 11s to 20ms.


hackmastera:
oh. there's a test failure though :(

tpendragon:
Yeah, looks like we populate some bad data for a test.

PR #3573, opened 2020-01-08
Adds a Rake task for reporting Numismatic::Coins which do not have members

### LINKED ISSUE #3448: Create a list of coins that have no tiff and no jpeg

jrgriffiniii:
This was necessary in order to resolve #3448, and it should be noted that a new query (`FindCoinsMissingMembers`) was needed in order to provide this Task.


tpendragon:
@jrgriffiniii @eliotjordan Did you two run this SQL query in prod in the dbconsole? I wanna make sure it works if there are no members for some resources.

jrgriffiniii:
While we were pairing yesterday @eliotjordan was testing this using pgAdmin, and we did test this on the staging environment (please see https://github.com/pulibrary/figgy/issues/3448#issuecomment-571814693)

PR #3578, opened 2020-01-09
Refactors pageclass

cwulfman:
Creates original and derivative subclasses for Page class.

Removes pagedata method, because we don't use Hathi's controlled vocabulary for the labels.


cwulfman:
Rubocop is complaining about a non-existent problem.

PR #3608, opened 2020-01-31
Set Ruby to 2.6.5/NodeJS to 8.17.0

tpendragon:
We're testing and deploying to 2.6.5 now.


hackmastera:
> does rubocop need to be updated?

Yeah, looks like 2.6 support was added in https://github.com/rubocop-hq/rubocop/releases/tag/v0.60.0

tpendragon:
@hackmastera I killed the postfixes.

PR #3609, opened 2020-01-31
use Lando for solr

tpendragon:
Just playing with Lando as an option for dev dependencies.

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -1,0 +1,19 @@
+name: figgy
+services:
+  figgy_test_solr:
+    type: solr:7
+    portforward: true
+    core: figgy-core-test
+    config:
+      dir: solr/config
+  figgy_development_solr:
+    type: solr:7
+    portforward: true
+    core: figgy-core-dev
+    config:
+      dir: solr/config
+proxy:
+  figgy_test_solr:
+    - figgy.test.solr.lndo.site:8983
+  figgy_development_solr:
+    - figgy.dev.solr.lndo.site:8983
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ Remember you'll need to run `bundle install` and `yarn install` on an ongoing ba
 
 ## Setup server
 
+### Manual Setup
+
 1. For test:
    - `RAILS_ENV=test rake db:setup`
    - `rake figgy:test`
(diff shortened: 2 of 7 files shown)


jeremyf:
Seems legible, though as someone that continues to develop bare metal, its a bit beyond my common usage. Though this next computer I'm getting will probably push me towards standardized development environments.

README.md
```diff
@@ -101,6 +103,23 @@ Remember you'll need to run `bundle install` and `yarn install` on an ongoing ba
      - If you run into problems with `solr_wrapper`, you can also start Solr with `rake figgy:development`
    - Access Figgy at http://localhost:3000/
 
+### Lando
+
+1. Uninstall Docker
+2. Install Lando DMG from [[https://github.com/lando/lando/releases]]
```
hackmastera, line 109: What version of lando?

eliotjordan: I used the latest `v3.0.0-rrc.4` and it works well.

tpendragon:
This used to have redis and I removed it. Maybe a follow-up?

.lando.yml
```diff
@@ -0,0 +1,19 @@
+name: figgy
+services:
+  figgy_test_solr:
+    type: solr:7
```
christinach, line 4: is it still solr 7 for figgy?

hackmastera: yep

README.md
```diff
@@ -101,6 +103,23 @@ Remember you'll need to run `bundle install` and `yarn install` on an ongoing ba
...
+3. `lando start`
+
+1. For test:
+   - `RAILS_ENV=test rake db:setup`
+   - In a separate terminal: `bundle exec rspec`
+   - Run jest tests: `yarn test`
+2. For development:
+   - ``export SECRET_KEY_BASE=`rake secret` ``
+   - `rake db:setup`
+   - In a separate terminal: `foreman start` (you can close the one launching solr)
```
hackmastera, line 119: Procfile still includes solr wrapper. This will conflict with the lando port so it should probably be removed.

hackmastera:
I left a couple of comments that can be addressed by documentation. Otherwise, the lando setup is nice but docker use is definitely taxing my system. I might need to work with it for a bit to gauge impact.

I don’t think I’d want to move other services for now (e.g. redis)

christinach:
@tpendragon lando ran smoothly. My only concern is that docker consumes a high amount of CPU but maybe this is something I can improve in my local docker installation.

README.md
tpendragon: It doesn't conflict with the lando port, but it isn't necessary so I'll remove it.

hackmastera: Hm. Not sure why foreman failed for me then. I don't generally use it so there may have been a pre-existing or unrelated issue. anyway 👍

PR #3632, opened 2020-02-05
Combine two PUO recordings with the same bib-id into one Recording

### LINKED ISSUE #3569: Performances for multiple days should show up in the same Recording.
In the database two days are two recordings, but in Figgy two days is one bib record. So we should shove all the audio files for each day, in order, into one recording.

tpendragon:
Closes #3569

### DIFF
--- a/app/services/music_import_service.rb
+++ b/app/services/music_import_service.rb
@@ -288,7 +288,8 @@ class MusicImportService
         st << { nodes: nodes, label: date }
       end
       change_set = DynamicChangeSet.new(output)
-      change_set.logical_structure = [{ label: "By Date", nodes: structure }]
+      change_set.logical_structure[0].label = "By Date"
+      change_set.logical_structure[0].nodes += structure
       buffered_change_set_persister.save(change_set: change_set)
     end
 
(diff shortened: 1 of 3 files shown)


app/services/music_import_service.rb
```diff
@@ -288,7 +288,11 @@ def create_logical_structure(output, buffered_change_set_persister)
         st << { nodes: nodes, label: date }
       end
       change_set = DynamicChangeSet.new(output)
-      change_set.logical_structure = [{ label: "By Date", nodes: structure }]
+      if change_set.logical_structure[0][:label] == ["By Date"]
+        change_set.logical_structure[0].nodes += structure
```
hackmastera: It's a bit confusing to use dot notation in this line but hash access in the line above.

app/services/music_import_service.rb
```diff
@@ -288,7 +288,11 @@ def create_logical_structure(output, buffered_change_set_persister)
         st << { nodes: nodes, label: date }
       end
       change_set = DynamicChangeSet.new(output)
-      change_set.logical_structure = [{ label: "By Date", nodes: structure }]
+      if change_set.logical_structure[0][:label] == ["By Date"]
```
hackmastera: Can you add a comment explaining why this check is made? I'm not sure what makes this label special or what other labels there may be.

hackmastera: It seems like it's just a matter of whether there's already a value?

tpendragon:
@hackmastera Refactored that conditional a bit. It should be ready for review again.

Issue #1836, opened 2018-09-12
Staff user cannot view private object they created


hackartisan:
a staff user should be able to view an object they created regardless of visibility

hackmastera:
permissions. not sure there's enough here to reproduce

tpendragon:
Tested, works.

Issue #1367, opened 2018-06-04
Staff users can't delete FileSets they have created


escowles:
The `Ability` class is checking whether the FileSet's `depositor` matches the user's id, but the FileSet's `depositor` property is `nil`.

hackmastera:
What should it check? `depositor` on the FileSet's parent?

escowles:
Checking the parent's `depositor` sounds good — it means there is only one definition of who deposited an item for purposes of checking who can delete things.

hackmastera:
permissions

PR #3644, opened 2020-02-07
allows staff users to use vue select in numismatics issue form

### LINKED ISSUE #3617: Staff users don't have access to numismatics dropdowns
When editing numismatics issues/coins, the vue-select ajax lookup mechanism only works for admin users

tampakis:
Closes #3617

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,7 @@ Release notes template:
 ## Removed
 
 -->
+
 # 2020-02-11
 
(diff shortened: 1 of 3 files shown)


hackmastera:
I wonder if this could be the cause also of #1367 and #1836. The "staff users can't do everything we'd expect them to be able to do" has been a longstanding bug.

christinach:
@tampakis thank you for fixing this issue. can you rebase and push again?

PR #3668, opened 2020-02-12
Allow download of HLS partials if open/complete.

### LINKED ISSUE #3652: Recording Downloads are always requiring login, no matter what.
This is breaking anonymous viewing in DPUL.

tpendragon:
Closes #3652


tpendragon:
@hackmastera and I discovered in looking through this that it wouldn't work for authenticated users anyways. In fact, NetID-auth'd downloads don't work at all.

PR #3675, opened 2020-02-13
Ensuring that the Universal Viewer is not rendered for Issues where Coins do not have FileSets

### LINKED ISSUE #3658: Remove UV from Issue show page and move to Coin show page
On the Issue show page, there is a big black box for the UV which never renders if there aren't images attached to the issue's coins.

Move the UV rendering to the coin show page instead because it shows a preview of what the viewer will look like in the public catalog UI.

jrgriffiniii:
Resolves #3658

### DIFF
--- a/app/views/catalog/_universal_viewer_issue.html.erb
+++ b/app/views/catalog/_universal_viewer_issue.html.erb
@@ -1,4 +1,4 @@
-<% if resource.class.can_have_manifests? && resource.try(:member_ids).present? && !Wayfinder.for(resource).coin_file_sets.empty? %>
+<% if resource.class.can_have_manifests? && !Wayfinder.for(resource).coin_file_sets.empty? %>
   <div class="intrinsic-container intrinsic-container-16x9 uv-container">
     <%# note: UV html file is '/viewer' %>
     <iframe src="<%= universal_viewer_path(resource) %>" allowfullscreen="true"></iframe>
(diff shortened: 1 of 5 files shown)


app/views/catalog/_universal_viewer_issue.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if resource.class.can_have_manifests? && resource.try(:member_ids).present? && !resource.decorate.decorated_coins.map(&:decorated_file_sets).flatten.empty?  %>
```
tpendragon: Rather than using the decorator can we do something like `Wayfinder.for(resource).coin_file_sets.empty?`?

app/views/catalog/_universal_viewer_issue.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if resource.class.can_have_manifests? && resource.try(:member_ids).present? && !Wayfinder.for(resource).coin_file_sets.empty? %>
```
tpendragon: Can you change `!Wayfinder.for(resource).coin_file_sets.empty?` to `Wayfinder.for(resource).coin_file_sets.present?`. The double negative spins me for a bit of a loop.

spec/resources/numismatics/issues/numismatics/issue_wayfinder_spec.rb
```diff
@@ -18,4 +18,21 @@
       expect(numismatic_issue_wayfinder.issues_count).to eq 1
     end
   end
+
+  describe "#coin_file_sets" do
```
tpendragon: :+1:

app/views/catalog/_universal_viewer_issue.html.erb
tpendragon: Oh, also you should be able to just rip out `resource.try(:member_ids).present?` and just put that in the wayfinder.

PR #3680, opened 2020-02-14
Ensuring that Vue ajax-select Components are mounted when added in multiple select fields for Numismatic Issues

### LINKED ISSUE #3666: Numismatics reference ajax dropdown doesn't work with multiple nested resources
Screen Shot 2020-02-12 at 11 35 04 AM

When clicking "add another citation" on the issue form, the "Numismatics reference" dropdown doesn't appear.

Also, it was requested to move the reference autocomplete field above the part and number fields

jrgriffiniii:
I am still testing this, but given that this seems to be a higher priority issue, I thought that I should offer this in order to expedite requesting input.

Resolves #3666

### DIFF
--- a/spec/resources/numismatics/issues/numismatics/issues_feature_spec.rb
+++ b/spec/resources/numismatics/issues/numismatics/issues_feature_spec.rb
@@ -196,58 +196,6 @@ RSpec.feature "Numismatics::Issues" do
     end
   end
 
-  context "with an existing reference" do
-    let(:reference1) { FactoryBot.create_for_repository(:numismatic_reference) }
-    let(:reference2) { FactoryBot.create_for_repository(:numismatic_reference, title: "Test Reference2") }
-
-    before do
-      reference1
-      reference2
-    end
-
-    scenario "when users add citations for an Issue", js: true do
-      visit edit_numismatics_issue_path(numismatic_issue)
-
-      # find("#numismatic_citation input.vs__search").send_keys("Test Reference", :enter)
-
-      expect(find_all("#numismatic_citation + .nested-fields input")).to be_empty
-      find("#numismatic_citation + .links .add").click
-      expect(find_all("#numismatic_citation + .nested-fields input").length).to eq(2)
-
-      # find("#numismatic_citation + .nested-fields input.vs__search").send_keys("Test Reference2", :enter)
-      # click("Save")
-
-      # expect(page).to have_css ".attribute.citations", text: "Test Reference"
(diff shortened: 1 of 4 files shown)


jrgriffiniii:
This is working for me locally, I've just been struggling with implementing feature tests (the Capybara/Selenium element selectors are not functioning as expected).

spec/resources/numismatics/issues/numismatics/issues_feature_spec.rb
```diff
@@ -196,6 +196,58 @@
...
+
+    before do
+      reference1
+      reference2
+    end
+
+    scenario "when users add citations for an Issue", js: true do
+      visit edit_numismatics_issue_path(numismatic_issue)
+
+      # find("#numismatic_citation input.vs__search").send_keys("Test Reference", :enter)
```
jrgriffiniii: I could not manage to find any way to trigger this behavior - I spent some time attempting various combinations of CSS selectors, using Nokogiri to isolate the DOM elements, and using jQuery through https://rubydoc.info/github/jnicklas/capybara/master/Capybara/Session#evaluate_script-instance_method

I can remove these statements (or tests entirely), as I am not quite certain how to proceed with testing for this.

spec/resources/numismatics/issues/numismatics/issues_feature_spec.rb
```diff
@@ -196,6 +196,58 @@
...
+
+    before do
+      person1
+      person2
+    end
+
+    scenario "when users add artists for an Issue" do
+      visit edit_numismatics_issue_path(numismatic_issue)
+
+      # find("#numismatic_artist input.vs__search").send_keys("name1", :enter)
```
jrgriffiniii: Please see the comment above.

jrgriffiniii:
This isn't completely ready to be merged, however, I would need to please request a review in response to https://github.com/pulibrary/figgy/pull/3680/files#r379537180

spec/resources/numismatics/issues/numismatics/issues_feature_spec.rb
eliotjordan: Maybe remove the statements. Do you think the tests are useful without them? If not, I might remove them as well.  I have found doing JS testing of these components with Capybara nearly impossible.

jrgriffiniii: I can remove the statements, thank you for reviewing this.

There is one approach I might be able to propose for future PRs, using https://jestjs.io/docs/en/puppeteer, but I will delay this for future work.

PR #3691, opened 2020-02-17
Make complete the default state on numismatic Issues

eliotjordan:
I wrote up a detailed ticket describing why this PR might be desirable, but it didn't save for some reason and I don't feel up to recreating it in full. Long story short, I think that we should set Numismatic Issue states to "complete" by default and not propagate state changes to Coin members.

- Issues are like containers that hold metadata but no files. Users will add coins as they are acquired and cataloged.
- Focusing on getting the numismatics lab users to use workflow functionality for Coin resources seems much more important.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,6 +16,14 @@ Release notes template:
 
 # 2020-02-14
 
+## Changed
+
+ * Sets default state of Numismatics Issues to "complete".
+ * Prevents propagation of Issue state to Coin members.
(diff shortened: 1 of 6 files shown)


eliotjordan:
A second question is whether we should be minting ARKs for Issues. Issues appear to me to be more like Ephemera boxes than resources that need to be directly accessed. I've pushed a commit that removes that functionality.

escowles:
I agree that workflow for Coins seems much more important than for Issues.

PR #3700, opened 2020-02-20
Uses ENV["OBJID"] to associate identifier with existing resource

cwulfman:
identifier is now passed into IngestFolderJob, so that images can be associated with already-created ScannedResources from EADs.


cwulfman:
This PR is blocked on an incomplete test in `spec/jobs/delete_archival_collection_job_spec.rb`. Generation of a dummy resource to test deletion does not function.

PR #3749, opened 2020-03-05
Add a re-indexer service for resources indexed into Orangelight

eliotjordan:


app/services/orangelight_reindexer.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+class OrangelightReindexer
+  # Reindexes resources indexed into Orangelight by sending record updated messages
+  # for all complete resources.
+  # @param optional logger [Logger] instance of logger class
+  def self.reindex_orangelight(logger: Logger.new(STDOUT))
+    new(logger: logger).reindex_orangelight
+  end
+
+  attr_reader :logger, :ogm_repo_path
```
tpendragon: `ogm_repo_path` left over? Can you remove it?

app/services/orangelight_reindexer.rb
```diff
@@ -0,0 +1,48 @@
...
+  private
+
+    def all_orangelight_resources
+      [
+        complete_resources(model: Numismatics::Coin)
+      ]
+    end
+
+    def complete_resources(model:)
+      resources = query_service.find_all_of_model(model: model)
```
tpendragon: This is going to be really slow, right? If I spun up a PR real quick to add model filtering and lazy-loaded cursors to `find_by_property` would you use that instead?

app/services/orangelight_reindexer.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  attr_reader :logger, :ogm_repo_path
+  def initialize(logger:)
+    @logger = logger
+  end
+
+  def reindex_orangelight
+    all_orangelight_resources.each do |resources|
+      resources.each do |resource|
```
tpendragon: Why are there two arrays?

lib/tasks/reindex.rake
```diff
@@ -28,3 +28,10 @@ namespace :geoblacklight do
     GeoResourceReindexer.reindex_geoblacklight
   end
 end
+
+namespace :orangelight do
+  desc "Issue rabbitmq messages that each complete orangelight resource is updated"
+  task reindex: :environment do
+    OrangelightReindexer.reindex_orangelight
```
tpendragon: Maybe you should force `ENV["BULK"] = true` here if it's important, and test for it in the message.

hackmastera: I like this idea

app/services/orangelight_reindexer.rb
tpendragon: https://github.com/pulibrary/figgy/pull/3756 has that feature.

hackmastera: incorporated

app/services/orangelight_reindexer.rb
hackmastera: it was copied from the geo service. I un-nested them.

lib/tasks/reindex.rake
hackmastera: @tpendragon I forced it in the service. Is 1b10d38 what you were thinking?

tpendragon: Looks good to me.

PR #3751, opened 2020-03-06
Generate JP2s with OpenJPEG

### LINKED ISSUE #2989: Replace Kakadu
We are currently using Kakadu's `kdu_compress` to convert TIFFs to JP2k.  Replace `kdu_compress` with another tool, such as OpenJPEG.

Related to #163

tpendragon:
Our Kakadu license is running out and we believe this will work for us.

Closes #2989

Requires https://github.com/pulibrary/princeton_ansible/pull/1359

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
       - checkout
       - run: sudo sh bin/ci_mediainfo_install.sh
       - run: sudo apt-get update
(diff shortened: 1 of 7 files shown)


tpendragon:
@roelmoon took a look at some items and says they looked okay. There's color shift between the originals and the viewer, but it's the same color shift that currently exists (maybe the image server doesn't respect the color profile?).

PR #3765, opened 2020-03-18
Use Pyramidal Tiffs

tpendragon:
This uses Pyramidal Tiffs instead of JP2s. Doing so allows us to use the much faster and open source VIPS library as well as cloud delivery options such as https://github.com/nulib/serverless-iiif. As is this works with Cantaloupe, which in staging is now serving both JP2s and pyramidals without any change.

Currently this PR continues to generate the JP2 so that we can change our minds at any point. If it works out we could revert that decision and stop generating a JP2 (except for Hathi submission.) We could also decide to push the Pyramidals up to AWS and whenever a pyramidal exists to link to the AWS lambda-powered service.

There are trade-offs however.

1. This compresses the individual tiles with 90 quality JPEG compression. Before the tiles finish loading it's noticeable.

2. The pyramidal tiffs are significantly larger than their JP2 counterparts:

    Pyramidal: `-rw-r--r-- 1 pudl pudl  55M Mar 18 19:53 intermediate_file.tif`
    JP2: `-rw-r--r-- 1 pudl pudl  26M Mar 18 19:53 intermediate_file.jp2`

It takes between 10-20 seconds to generate a pyramidal tiff. I'm generating 1000 now so I can hammer Cantaloupe and see how it holds up.

Cantaloupe configured with the JAIProcessor still swaps in ~ 40 seconds when requesting full resolution tiffs.

Performance Testing:
It takes 90 seconds for responses to iiif-staging1 to start failing with Kakadu serving up JP2s, 40 seconds for iiif-staging1 to start failing with Tiffs. Same issue, it’s just too many requests - RAM fills up.
~ 40 seconds for OpenJPEG to crash.

When asking for 1000px wide images instead of full:

It takes 97 seconds to crash with pyramidals (and each image is served up in ~ 3 seconds), with OpenJPEG it takes 6-16 seconds to generate each tile but memory never goes up in any considerable way, Kakadu takes between 3-8 seconds and memory usage slowly rises.


tpendragon:
The AWS Lambda is shipped and is taking ~ 3 seconds to serve literally as many requests as you can possibly give it. I zoom'd with Michael Klein from Northwestern for a bit and he helped me gain some insight to the stack.

Theirs responds faster. They use the following derivative generation rules:

```
    const pyramidTiff = await sharp(inputFile, { limitInputPixels: false })
      .resize({
        width: 15000,
        height: 15000,
        fit: "inside",
        withoutEnlargement: true
      })
      .tiff({
        compression: "jpeg",
        quality: 75,
        tile: true,
        tileHeight: 256,
        tileWidth: 256,
        pyramid: true
      })
      .toBuffer();
```

They limit derivatives to 15k pixels at the largest, jpeg compress to 75 quality, and set tiles to 256x256.

tpendragon:
If I set the JPEG quality to `75` and the tiles to `256x256` the really large images load in a few seconds, but smaller zoomed in tiles take about half a second. The derivatives are much smaller too. Only issue is they take much longer to generate.

app/change_set_persisters/change_set_persister.rb
```diff
@@ -164,8 +164,10 @@ def add
       def run; end
     end
 
-    def with(metadata_adapter:)
-      yield self.class.new(metadata_adapter: metadata_adapter, storage_adapter: storage_adapter, transaction: true, characterize: @characterize, queue: queue, handlers: handlers)
+    def with(metadata_adapter: self.metadata_adapter, storage_adapter: self.storage_adapter)
```
hackmastera, line 167: I'd never noticed this method before. It's an interesting kind-of-a-factory? Could you add a comment explaining its use?

config/initializers/valkyrie.rb
```diff
@@ -136,6 +136,41 @@
...
+      Valkyrie::Storage::Shrine.new(
+        Shrine.storages[:pyramidal_storage],
+        nil,
+        Valkyrie::Storage::Disk::BucketedStorage,
+        identifier_prefix: "pyramidal-derivatives"
+      ),
+      :pyramidal_derivatives
+    )
+  else
+    Valkyrie::StorageAdapter.register(
```
hackmastera, line 159: Maybe add a comment that this is the development use case?

app/services/manifest_builder.rb
```diff
@@ -773,6 +774,18 @@ def base_url(file_set)
         nil
       end
     end
+
+    private
+
+      def pyramidal_url(file_set)
```
hackmastera, line 780: Does this provide the pyramidal_url for objects stored in s3, also? Even though it's the `CantaloupeHelper`?

app/services/manifest_builder.rb
```diff
@@ -739,6 +739,8 @@ def protocol
     def manifest_image_path(resource)
       if (Rails.env.development? && Figgy.config["pyramidals_bucket"].blank?) || Rails.env.test?
         RiiifHelper.new.base_url(resource.id)
+      elsif resource.pyramidal_derivative.present?
+        PyramidalHelper.new.base_url(resource)
```
hackmastera, line 743: Thanks -- it's nice to have this made explicit.

app/derivative_services/vips_derivative_service.rb
```diff
@@ -0,0 +1,167 @@
...
+    )
+    raise "Unable to store pyramidal TIFF for #{filename}!" unless File.exist?(temporary_output.path)
+  end
+
+  def vips_image
+    @vips_image ||=
+      begin
+        image = Vips::Image.new_from_file(filename.to_s)
+        if image.height >= REDUCTION_THRESHOLD || image.width >= REDUCTION_THRESHOLD
+          image.resize(0.5)
```
eliotjordan, line 96: I wonder if it would be better to set the scale factor proportionally based on the pixels above the REDUCTION_THRESHOLD. For example, if there is an image with a height of 31,000 pixels, the resized image will still be above 15,000. Such large images are probably not that common, I would image, so maybe it's not worth the trouble.

PR #3774, opened 2020-03-25
Adds identifier property to ScannedResourceChangeSet

cwulfman:
Required so that identifier can be passed in to IngestFolderJob


cwulfman:
I tweaked the failing test, so everything should pass now.

PR #3778, opened 2020-03-30
Adding support for PNG images

### LINKED ISSUE #3777: Add support for PNG images
We currently support ingesting TIFF and JPEG images, but not PNG. We have ephemera selectors downloading PNGs and would like to ingest them.

escowles:
Fixes #3777


escowles:
@hackmastera Good point — I added similar support for PNG to the VIPS derivative service.

PR #3784, opened 2020-04-06
Don't validate pyramidals post-upload

tpendragon:
Validating post-upload was causing the file to be re-downloaded after it
was done uploading. This was expensive and should be unnecessary with the
underlying library.


app/adapters/shrine/null_verifier.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Shrine
+  # Null S3 verifier. Implemented so that every file uploaded isn't immediately
```
hackmastera, line 4: 👍

PR #3817, opened 2020-04-20
Upload height/width to S3 Metadata with VIPS.

tpendragon:
This allows our lambda image server to work.

### DIFF
--- a/app/derivative_services/vips_derivative_service.rb
+++ b/app/derivative_services/vips_derivative_service.rb
@@ -17,11 +17,12 @@ class VIPSDerivativeService
   end
 
   class IoDecorator < SimpleDelegator
-    attr_reader :original_filename, :content_type, :use
-    def initialize(io, original_filename, content_type, use)
+    attr_reader :original_filename, :content_type, :use, :upload_options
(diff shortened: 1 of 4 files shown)


app/derivative_services/vips_derivative_service.rb
```diff
@@ -17,11 +17,12 @@ def new(id:)
   end
 
   class IoDecorator < SimpleDelegator
-    attr_reader :original_filename, :content_type, :use
-    def initialize(io, original_filename, content_type, use)
+    attr_reader :original_filename, :content_type, :use, :width, :height, :upload_options
```
hackmastera, line 20: are you using `:width` and `:height` readers?

app/models/pending_upload.rb
```diff
@@ -8,6 +8,7 @@ class PendingUpload < Valkyrie::Resource
   attribute :auth_header
   attribute :type
   attribute :provider
+  attribute :upload_arguments, Valkyrie::Types::Hash
```
hackmastera, line 11: Can you comment this new attribute with its intended use?

PR #3844, opened 2020-04-27
 Use first valid coverage value when generating GeoBlacklight document

### LINKED ISSUE #3843: GeoBlacklight document generator should default to the first valid coverage value
Some scanned maps have valid coverages set on the resource (usually from FGDC metadata documents), but have invalid imported coverages. The GeoBlacklight document generator errors on these records because it simply uses the imported value. It should choose the first valid coverage value instead. 

Example record:

https://figgy-staging.princeton.edu/catalog/41f9a2b9-c778-4415-9974-803b4bcdd3e5

https://figgy-staging.princeton.edu/concern/scanned_maps/41f9a2b9-c778-4415-9974-803b4bcdd3e5/geoblacklight

eliotjordan:
Closes #3843

### DIFF
--- a/app/services/geo_discovery/document_builder/spatial_builder.rb
+++ b/app/services/geo_discovery/document_builder/spatial_builder.rb
@@ -19,7 +19,10 @@ module GeoDiscovery
         # Parses coverage field from geo resource and instantiates a coverage object.
         # @return [GeoCoverage] coverage object
         def coverage
-          @coverage ||= GeoCoverage.parse(resource_decorator.coverage.first)
+          @coverage ||= begin
+                          valid_coverages = resource_decorator.coverage.map { |c| GeoCoverage.parse(c) }.compact
+                          valid_coverages.first
+                        end
         end
 
         # Returns the coverage in solr format. For example:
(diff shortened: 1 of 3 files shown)


app/services/geo_discovery/document_builder/spatial_builder.rb
```diff
@@ -19,7 +19,10 @@ def build(document)
         # Parses coverage field from geo resource and instantiates a coverage object.
         # @return [GeoCoverage] coverage object
         def coverage
-          @coverage ||= GeoCoverage.parse(resource_decorator.coverage.first)
+          @coverage ||= begin
+                          valid_coverages = resource_decorator.coverage.map { |c| GeoCoverage.parse(c) }.compact
```
hackmastera, line 23: Do we want to push this logic to the decorator?

eliotjordan: In this case, the decorator is special decorator that merges direct and imported metadata. I'd rather not clutter that with one-off business logic.

https://github.com/pulibrary/figgy/blob/master/app/decorators/geoblacklight_metadata_decorator.rb

PR #3853, opened 2020-04-29
Add link to catalog and finding aids to jsonld

hackmastera:


app/services/identifier_service.rb
```diff
@@ -21,7 +21,7 @@ def self.get_ark_result(ark:)
 
   def self.url_for(resource)
     return Rails.application.routes.url_helpers.solr_document_url(resource, host: Figgy.default_url_options[:host]) if resource.try(:source_metadata_identifier).blank?
-    return "https://catalog.princeton.edu/catalog/#{resource.source_metadata_identifier.first}#view" if PulMetadataServices::Client.bibdata?(resource.source_metadata_identifier.first)
```
tpendragon, line 24: This will change where all of our ARKs point, which seems like an unintended side effect.

hackmastera: Will this mess up existing ARKs?

hackmastera: I don't really understand why they point there.

tpendragon: It will mess up any ARKs that are updated, yeah. They point here so that when you click the ARK it jumps to the viewer if possible.

hackmastera: Okay I pushed a better idea.

Issue #3839, opened 2020-04-24
Some Finding Aids fields are not showing up in Figgy and one field is mapped to the wrong place in Figgy


kelea99:
This was discovered during a meeting with Gabriel Swift, regarding collection review of [WC064 in Figgy](https://figgy.princeton.edu/catalog?utf8=✓&q=WC064) (Western Americana)versus the [Finding Aid](https://findingaids.princeton.edu/collections/WC064/c423) before completion. @cwulfman and I checked another collection - [Sylvia Beach 0108 in Figgy](https://figgy.princeton.edu/?f%5Bmember_of_collection_titles_ssim%5D%5B%5D=Sylvia+Beach+Papers+C0108&q=Sylvia+beach) and in [Finding Aids](https://findingaids.princeton.edu/collections/C0108/c000002357) and the findings were the same. This is a probable issue with pulling in the FA metadata in general versus collection-specific.

1. The **Creator** field in figgy is pulling in the metadata from the **Collection Creator** field in the finding aid. There are two separate entries in the FA: **Creator** and **Collection Creator**. It is important to have the Creator information for the Western Americana collection, as it identifies the photographer, which is the primary method of discovery for this collection by patrons, according to Gabriel.
2. **Description** field from FA (this appears to have been the same metadata as the "notes" field in PUDL) is nowhere in Figgy item. **Description** field is important for the Western Americana collection, as it gives additional information for the piece and, in some cases, who holds the copyright.
3. **Dimensions** field from FA is nowhere in Figgy item.

kelea99:
@escowles , this will probably affect the grounds and buildings work as well. I'd view this as a priority. thoughts?

tpendragon:
Before I move this to backlog can we get some specific references to items that should have these metadata fields? Both a link to their finding aid page, as well as the item ingested in Figgy, so we have a good handle on the success state.

tpendragon:
@kelea99 Can you find examples of the above?

kelea99:
sorry, I just saw this, @tpendragon. one sec

kelea99:
@tpendragon getting more examples now, but there are links to a couple in the original comment. is this what you are asking for? can also make screenshots.

kelea99:
Just a question: are all fields from the FA supposed to be represented in Figgy? if so, there are others that should be added to the "needed in figgy" column. 

**Berthier Collection** 

[...]

PR #3877, opened 2020-05-07
Implements component_creators

cwulfman:
For the creator of any pulfa record, the previous version always returned the collection_creators (as expressed in the `pulfa:context` element added to the EAD at pulfa ingest). Gabrielle Swift (#3839) asked that the creator of individual components be drawn from the component's `ead:origination` element, if it exists.

### DIFF
--- a/spec/services/pul_metadata_services/pulfa_record_spec.rb
+++ b/spec/services/pul_metadata_services/pulfa_record_spec.rb
@@ -8,20 +8,41 @@ describe PulMetadataServices::PulfaRecord do
   let(:source) { file_fixture("pulfa/#{pulfa_id}.xml").read }
 
   describe "#attributes" do
-    it "works" do
-      expected = {
-        title: ["19th Century Catalog and Correspondence, Pre-Vinton, 1811-"],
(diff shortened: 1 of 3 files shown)


spec/services/pul_metadata_services/pulfa_record_spec.rb
```diff
@@ -8,20 +8,41 @@
...
+      end
+    end
+
+    context "with a component that has origination" do
+      let(:pulfa_id) { "WC064_c1630" }
+      it "works" do
+        expected = {
+          title: ["Picnic west of Syracuse about 1885"],
+          created: ["1885-01-01T00:00:00Z/1885-12-31T23:59:59Z"],
+          creator: ["Savage, C. R. (Charles Roscoe),\n                1832-1909"],
```
tpendragon, line 35: Should we find some way to parse this to not have all that white space and the newline? Maybe a different xpath query? Or however traject processes these strings?

cwulfman: There's a lot of extra whitespace in these EADs; I was assuming this data was normalized somewhere further up the chain. Is there a general whitespace normalizer Figgy implements/uses, other than `strip`?

Issue #3874, opened 2020-05-05
Index more numismatics fields into Orangelight


eliotjordan:
Blocking  pulibrary/orangelight#1569

In order to generate values for the numismatics search form, we need to index more facet values:

- Donor (person component of accession)
- Department (? looks like we didn't migrate this field from the old database)
- Divide minting location into city, state, region

eliotjordan:
Closed by #3881

PR #3881, opened 2020-05-08
Update numismatics indexing into orangelight

eliotjordan:
Adds city, state, region, and donor.

refs #3874


app/resources/numismatics/coins/orangelight_coin_builder.rb
```diff
@@ -60,7 +60,8 @@ def document_coin_hash
         size_s: decorator.size,
         technique_s: decorator.technique,
         weight_s: decorator.weight,
-        holdings_1display: holdings_hash
+        holdings_1display: holdings_hash,
+        accession_info_s: decorator.accession_person
```
hackmastera, line 64: Should we call it `donor` here? We sort of lose the fact that it's a person in this step.

eliotjordan: 👍 That's a good idea.

eliotjordan: Updated.

PR #3884, opened 2020-05-08
File Manager Improvements

tpendragon:
* Orders files alphabetically
* No longer displays hidden files.

This should fix an issue where files are ingested in random order.


tpendragon:
@hackmastera Bah! Fixed, thanks for the ping.

PR #3900, opened 2020-05-15
Convert color space to SRGB through VIPS

tpendragon:
Fixes color muting issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,9 +19,10 @@ jobs:
       - checkout
       - run: sudo sh bin/ci_mediainfo_install.sh
       - run: sudo apt-get update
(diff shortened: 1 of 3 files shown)


tpendragon:
Do you have a test recommendation? I don't know much about color spaces. Presumably there's some tool that will analyze the embedded space of an image and tell you what it is.

Now CI VIPS matches the version installed on the servers. Before it was older on CI.

hackmastera:
Is this helpful? It says

```You can test for the presence of an embedded profile with vips_image_get_typeof() with VIPS_META_ICC_NAME as an argument. This will return GType 0 if there is no profile.```

https://libvips.github.io/libvips/API/current/libvips-colour.html#vips-icc-import

tpendragon:
@hackmastera Yeah, we do that to see if there's a profile, but I can't really tell what the profile is. Maybe I can check to see if the profile's changed, but I'm hesitant to check what the profile actually is in case it changes in a future version or something.

PR #3906, opened 2020-05-18
Attach numismatic monograms as members

### LINKED ISSUE #3449: Improve the process for creating and attaching monograms

eliotjordan:
- Monograms are now attached as members on the Numismatic Issue show page.

Closes #3449

monogram

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -460,10 +460,15 @@ class ManifestBuilder
   end
 
   class Numismatics::IssueNode < CollectionNode
-    # Only include coins which have file sets;
+    # Only include Coin resources which have file sets;
     # otherwise there is no image for the viewer.
     def members
-      @members ||= super.select { |coin| coin.member_ids.present? }
+      @members ||=
+        begin
+          decorate.coins_with_filesets.select do |member|
+            !current_ability || current_ability.can?(:read, member)
+          end
+        end
     end
   end
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,14 @@ Release notes template:
 ## Removed
 
 -->
+# 2020-05-18
+
+## Changed
+
+* Numismatic Monograms are now attached to Numismatic Issue resources as members. Instead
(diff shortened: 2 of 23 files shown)


app/services/manifest_builder.rb
```diff
@@ -460,10 +460,10 @@ def geo_image?(member)
   end
 
   class Numismatics::IssueNode < CollectionNode
-    # Only include coins which have file sets;
+    # Only include Coin resources which have file sets;
     # otherwise there is no image for the viewer.
     def members
-      @members ||= super.select { |coin| coin.member_ids.present? }
+      @members ||= super.select { |r| r.is_a? Numismatics::Coin }.select { |coin| coin.member_ids.present? }
```
hackmastera, line 466: I think this would be clearer if we pulled them directly from a decorator method and reimplemented the bit of permissions code that's being reused by `super`.

eliotjordan: Pushed logic to the Issue decorator.

eliotjordan:
@hackmastera Changed "Attach Monogram" to "Attach Child"

As to the larger question, I don't think that attaching monograms as members will be a problem. Ilya said that it is rare for a monogram to be shared among Issues; case in point, monogram images are labeled by issue number. If it becomes an issue we can modify the UI so that a existing monogram can be attached. Like what we do for geo resources:

Screenshot 2020-05-18 16 41 41

I don't think that it's worth spending time on it until it becomes a user request.

hackmastera:
Maybe I'm misremembering and it's only a file set that is not allowed to have more than one parent. Perhaps @tpendragon could arbitrate here. 

Otherwise 👍 on the edits. There's a failing test.

eliotjordan:
Even if having two parents is frowned upon, I don't think that's a show stopper here. Neither Alan or Ilya was bothered by uploading an image twice in the rare event of a duplicated monogram. The other option is to go back to where we were before and iterate on a convoluted interface-- how does one choose a monogram from a list of thousands, mostly without useful titles? -- to support an edge case.

tpendragon:
There's a bunch of these in there, can we close this @eliotjordan ?

PR #3952, opened 2020-06-02
Fix CI Build.

tpendragon:
Webpacker failed to compile with 12.18.1 for some reason, but that's not
what's on our servers.


tpendragon:
Closing this. Staging just got updated to 12.17, so we have to fix this build error.

Issue #3975, opened 2020-06-15
Add / populate optional fields in oai_dc


hackmastera:
All these are already defined, but not populated:
- [x] language
* ~~coverage~~ not desired
- [x] subject
- [x] source
- [x] (optional) contributor? Do any of these have contributors?
- [x] (optional) description? Do any of ours have descriptions?

Consult gallica documentation for requirements within each field.

@kelea99 which of these do we care about publishing? And if source is one of those, what should it be? It's supposed to correlate with data they already have but I'm not sure what this means; the note says "For partners having transmitted documents to the BnF for diffusion in Gallica (via a BnF digitisation contract or file integration operation), the source reference present in the OAI-PMH repository must be the same as the one already present in Gallica notices, in order to allow cross-searches 'Per library' in harvested and integrated collections."

kelea99:
@hackmastera Im reviewing the entire collection we submitted. 
Example of France in the Americas item (MARC) with Description and Contributor: https://figgy.princeton.edu/catalog/525767d2-75d4-41eb-89cf-11f973ec1701

kelea99:
**Source decision:**  Institution [Princeton University Library], identifier [bibid / component id]. @hackmastera will concentrate the two into one.  Discussed options with both Jennifer Baxmeyer and @escowles .

kelea99:
Question from @hackmastera confirm / deny whether we want language, coverage, subject:
(Note that in figgy coverage tends to mean geographic, but here it’s a date range):

- language: yes.  comply with requested ISO 639-2b standard 

- coverage: no. we provide the date(s) as part of mandatory fields, which should be sufficient.

- subject: it would be nice?But... Question: we pull subjects from the MARC records, but it does not look like we pull them for FA materials into figgy. If we provide subjects, does this mean we would have to provide them for FA content too? it looks like it works differently there.

hackmastera:
@kelea99 I don't think we'd have to provide subjects for the FA content. The field appears fully optional.

PR #4008, opened 2020-06-24
Add language to oai_dc; use ISO-639-2

hackmastera:
refs #3975


hackmastera:
@tpendragon this had to be rebased, and also there was a test failing; review again?

PR #4034, opened 2020-06-29
Extract viewer to controller.

### LINKED ISSUE #4025: Convert `/viewer` to a controller/view/template.
Right now it's a static HTML page, but we're about to add a bunch of logic and JS to support controlled digital lending.

The controller should have an `index` action, its own template, and one view. Extract JS to a file.

tpendragon:
Should make it easier for us to refactor the javascript.

Closes #4025.


app/assets/javascripts/viewer.js
```diff
@@ -0,0 +1,68 @@
+function loadUV () {
+  let urlDataProvider = new UV.URLDataProvider(false)
```
hackmastera, line 2: It looks like the scoping here is not right. See the deploy failure: https://gist.github.com/tpendragon/bcaa3d58547a04e6c78a28b9b907650a

tpendragon: @hackmastera It was just the `let` vs a `var`. Should be fixed now - just tested a deploy.

Issue #4033, opened 2020-06-29
Develop an abstraction to check which items for a given SourceMetadataIdentifier are eligible to be charged for CDL.


tpendragon:
Something like a
```ruby
class ChargedItemsQuery
  def initialize(source_metadata_identifier:)
  end

  def item_ids
    ['1234', '5678']
  end
end
```

This could be a valkyrie CustomQuery, but I don't know if that's necessary, and wouldn't be very separated from the rest of the stack.

The query shouldn't return the item IDs which aren't currently charged (that's for the CDL::ChargeManager to figure out), just those which have been placed on hold for the CDL program.

This will query bibdata.

hackmastera:
Per discussion in checkin, this will likely be a service with the query as a method on that service.

PR #4040, opened 2020-06-30
Add CDL::ChargeManager

### LINKED ISSUE #4026: Create a CDL::ChargeManager
A CDL::ChargeManager is a persisted object representing the charges for items attached to a digital object.

See https://docs.google.com/drawings/d/1GdntEd_fiS4rsFbILK63woF9862ZnO8KgV5zE4p-VVU/edit

Open Questions:

1. Should it be attached to a figgy resource or a source metadata identifier? If an identifier has two Figgy resources, do they share the same charge manager?
2. If bibdata's suppressing items, how will we know which ones are eligible for CDL? Maybe we can't, and those ones shouldn't be charged?

tpendragon:
The `CDL::ChargeManager` is a business logic object which creates and
manages `CDL::ResourceChargeList` and its `CDL::ChargeItem`s.

The implementation currently stubs the EligibleItemService, but after
 #4033 it will pull from bibdata.

Closes #4026

### DIFF
--- a/app/resources/cdl/charge_manager.rb
+++ b/app/resources/cdl/charge_manager.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+# Controlled Digital Lending
+module CDL
+  class UnavailableForCharge < StandardError; end
+  class ChargeManager
+    attr_reader :resource_id, :eligible_item_service, :change_set_persister
+    # TODO: default eligible_item_service from #4033
+    def initialize(resource_id:, eligible_item_service:, change_set_persister:)
+      @resource_id = resource_id
+      @eligible_item_service = eligible_item_service
+      @change_set_persister = change_set_persister
+    end
+
+    def available_for_charge?
+      return false unless item_ids.present?
+      resource_charge_list.charged_items.count < item_ids.count
+    end
+
+    def create_charge(netid:)
+      raise CDL::UnavailableForCharge unless available_for_charge?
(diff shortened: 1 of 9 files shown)


app/resources/cdl/charge_manager.rb
```diff
@@ -0,0 +1,55 @@
...
+# Controlled Digital Lending
+module CDL
+  class UnavailableForCharge < StandardError; end
+  class ChargeManager
+    attr_reader :resource_id, :eligible_item_service, :change_set_persister
+    # TODO: default eligible_item_service from #4033
+    def initialize(resource_id:, eligible_item_service:, change_set_persister:)
+      @resource_id = resource_id
+      @eligible_item_service = eligible_item_service
+      @change_set_persister = change_set_persister
```
christinach, line 12: I will update the name on my Pr to be EligibleItemService

christinach: Can we call it EligibleCdlService ? or just CdlService ?

christinach: Never mind since CDL is there it would be redundant to include the Cdl in the class name. I'm ok with EligibleItemService.

hackmastera: @christinach the name of the class doesn't matter, and does not need to change -- I think yours is more descriptive. It's just the method name and parameter signature we need to agree on. From the perspective of this class, it will be an "eligible item service" -- that's just the variable where it will hold the reference to the class you're writing.

hackmastera: Oh you already changed it -- it looks good that way too. Might as well let it be!!

PR #4041, opened 2020-06-30
CDL Manifest Permissions

### LINKED ISSUE #4027: Users should be able to view the manifest of an item which they have a current non-expired CDL::ChargedItem for.
See https://docs.google.com/drawings/d/1GdntEd_fiS4rsFbILK63woF9862ZnO8KgV5zE4p-VVU/edit

This logic is likely similar to how access tokens work.

### LINKED ISSUE #4028: Users should not see the download dialog even if they have a current CDL::ChargedItem instance.
The following lines control this logic, we can likely just adjust the Ability permissions:

https://github.com/pulibrary/figgy/blob/274166463dd1f7dfae31f35d464e557679dd175f/app/controllers/application_controller.rb#L89-L93

tpendragon:
Closes #4027, Closes #4028

### DIFF
--- a/app/resources/vector_resources/vector_resource_decorator.rb
+++ b/app/resources/vector_resources/vector_resource_decorator.rb
@@ -104,4 +104,13 @@ class VectorResourceDecorator < Valkyrie::ResourceDecorator
     return ["#{super.first} (#{portion_note.first})"] unless portion_note.blank?
     super
   end
+
+  # Overridden because we don't care if the object isn't visible - downloadable
+  # takes precedence for vector resources.
+  def downloadable?
+    return false unless respond_to?(:downloadable)
+    return visible? && public_readable_state? if downloadable.nil?
+
+    downloadable&.include?("public")
+  end
 end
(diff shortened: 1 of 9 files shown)


app/resources/vector_resources/vector_resource_decorator.rb
```diff
@@ -104,4 +104,13 @@ def title
     return ["#{super.first} (#{portion_note.first})"] unless portion_note.blank?
     super
   end
+
+  # Overridden because we don't care if the object isn't visible - downloadable
+  # takes precedence for vector resources.
+  def downloadable?
+    return false unless respond_to?(:downloadable)
+    return visible? && public_readable_state? if downloadable.nil?
+
```
christinach, line 113: @tpendragon can you explain a bit more this part?

tpendragon: I made it so that if an item's private it's not downloadable. However, for vector resources there's a test that says even if it's private, if someone marked it downloadable you should be able to download it.

That line says if nobody's marked it downloadable, then it's downloadable if it's both visible and in a public readable state (which is either "complete" or "flagged" atm)

christinach: Thank you.

PR #4042, opened 2020-07-01
Adds CDL service,

### LINKED ISSUE #4033: Develop an abstraction to check which items for a given SourceMetadataIdentifier are eligible to be charged for CDL.

christinach:
Adds CDL service,
Retrieves items with "on_cdl"= "Y" - from a given bibid (on_cdl="Y" is an assumption for the cdl charged items)
resolves #4033

### DIFF
--- a/app/services/cdl/eligible_item_service.rb
+++ b/app/services/cdl/eligible_item_service.rb
@@ -2,12 +2,14 @@
 
 module Cdl
   class EligibleItemService
+    # EligibleItemService will return from Voyager items eligible for Controlled Digital Lending (Cdl)
+    # on_cdl == "Y"
     class << self
       def item_ids(source_metadata_identifier:)
         item_ids = []
         conn = Faraday.new(url: "https://bibdata.princeton.edu/")
         response = conn.get("#{source_metadata_identifier}/items")
-        items = JSON.parse(response.body).values.map { |l| l[0]["items"] }
+        items = JSON.parse(response.body).values.map { |l| l[0]["items"] }.compact
         items.each do |i|
           if i[0]["on_cdl"] == "Y"
             item_ids << i[0]["id"]
--- a/app/services/cdl/eligible_item_service.rb
+++ b/app/services/cdl/eligible_item_service.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Cdl
-  class BibdataService
+  class EligibleItemService
     class << self
       def item_ids(source_metadata_identifier:)
         item_ids = []
--- a/app/services/cdl/eligible_item_service.rb
+++ b/app/services/cdl/eligible_item_service.rb
@@ -6,16 +6,10 @@ module Cdl
     # on_cdl == "Y"
     class << self
       def item_ids(source_metadata_identifier:)
-        item_ids = []
         conn = Faraday.new(url: "https://bibdata.princeton.edu/")
         response = conn.get("#{source_metadata_identifier}/items")
         items = JSON.parse(response.body).values.map { |l| l[0]["items"] }.compact
-        items.each do |i|
-          if i[0]["on_cdl"] == "Y"
-            item_ids << i[0]["id"]
-          end
-        end
-        item_ids
+        items.select { |i| i[0]["on_cdl"] == "Y" }.map { |i| i[0]["id"] }
       end
     end
   end
--- a/spec/services/cdl/bibdata_service_spec.rb
+++ b/spec/services/cdl/bibdata_service_spec.rb
@@ -96,7 +96,7 @@ RSpec.describe Cdl::BibdataService do
                  headers: { "Content-Type" => "application/json" })
   end
 
-  describe ".cdl_items" do
+  describe ".item_ids" do
     let(:id) { "7214786" }
 
     it "is a cdl charged item" do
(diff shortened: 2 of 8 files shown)


spec/services/cdl/bibdata_service_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+             {
+               :f => [
+                 {
+                   :holding_id => 1_581_046,
+                   :call_number => "NA203 .G5 1967",
+                   :items => [
+                     {
+                       :id => 1_666_778,
+                       :on_cdl => "N",
+                       :on_reserve => "N",
```
christinach, line 18: Assuming that the item will have an on_cdl attribute similar to the on_reserve.

spec/services/cdl/eligible_item_service_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+                               "Not Charged"
+                             ]
+                           }
+                         ]
+                       }
+                     ] }.to_json, headers: {})
+      end
+      let(:bib_id) { "7214786" }
+
+      it "will not error if the on_cdl is null" do
```
hackmastera, line 135: instead of "will not error" can we say "will return empty array"

app/services/cdl/eligible_item_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Cdl
+  class EligibleItemService
```
hackmastera, line 4: I think a class-level comment that says what this class is for would be nice here.

spec/services/cdl/eligible_item_service_spec.rb
```diff
@@ -0,0 +1,159 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe Cdl::EligibleItemService do
+  before do
+    stub_request(:get, "https://bibdata.princeton.edu/#{bib_id}/items")
+      .to_return(status: 200,
+                 body:
```
hackmastera, line 8: Can we move these bodies to fixture files so the test is easier to read?

spec/services/cdl/eligible_item_service_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+        stub_request(:get, "https://bibdata.princeton.edu/#{bib_id}/items")
+          .to_return(status: 200,
+                     body: { uesrf: [
+                       {
+                         holding_id: 1_581_050,
+                         call_number: "NA203 .G5 1967",
+                         items: [
+                           {
+                             id: 1_666_782,
+                             on_cdl: "null",
```
hackmastera, line 111: Should we test that `on_cdl` is just missing? Or is that essentially the same?

spec/services/cdl/eligible_item_service_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+    end
+
+    context "querying a suppressed bib" do
+      before do
+        stub_request(:get, "https://bibdata.princeton.edu/#{bib_id}/items")
+          .to_return(status: 404,
+                     body: {}.to_json, headers: {})
+      end
+      let(:bib_id) { "11174664" }
+      it "will not error" do
```
hackmastera, line 147: I think we should use "will return empty array" instead of saying what it will not do.

spec/services/cdl/eligible_item_service_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+        expect(described_class.item_ids(source_metadata_identifier: bib_id)).to eq []
+      end
+    end
+
+    context "a bib_id with items in more than one locations" do
+      let(:bib_id) { "7214786" }
+      it "returns only the cdl charged items" do
+        expect(described_class.item_ids(source_metadata_identifier: bib_id)).to eq [1_666_779, 1_666_782]
```
hackmastera, line 155: 👍

spec/services/cdl/eligible_item_service_spec.rb
christinach: I wouldn't expect the item not to have an `on_cdl`. My assumption is that it will be in the `Item` table in voyager. But it doesn't hurt to test this as well.

christinach:
@hackmastera can you check again?

spec/services/cdl/eligible_item_service_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+require "rails_helper"
+
+RSpec.describe Cdl::EligibleItemService do
+  describe ".item_ids" do
+    context "on_cdl is null" do
+      before do
+        stub_request(:get, "https://bibdata.princeton.edu/#{bib_id}/items")
+          .to_return(status: 200,
+                     body: file_fixture("bibdata/#{bib_id}.json").read, headers: { "Content-Type" => "application/json" })
+      end
```
christinach, line 11: I tried moving this stub_request out of the context and below the describe to make the spec thinner but then the stub_request with the 404 which is in the other context would not pass.

christinach:
Thank you @hackmastera . I updated the method. The specific header was not creating any problems but I agree that for consistency it would be good to be there.

christinach:
ofcourse I didn't save after updating the header. so I'm pushing again

PR #4044, opened 2020-07-01
Don't use JS to handle viewer auth flow.

### LINKED ISSUE #4043: Refactor Viewer control logic to use views rather than JS

tpendragon:
This should make it easier for us to handle things like CDL auth flows without a bunch of complex javascript branching or stashing data in manifest bodies.

Closes #4043

### DIFF
--- a/app/views/layouts/viewer_layout.html.erb
+++ b/app/views/layouts/viewer_layout.html.erb
@@ -3,13 +3,14 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
     <link rel="stylesheet" type="text/css" href="uv/uv.css">
     <%= stylesheet_link_tag 'viewer' %>
-    <script type="text/javascript" src="uv/lib/offline.js"></script>
-    <script type="text/javascript" src="uv/lib/offline.js"></script>
-    <script type="text/javascript" src="uv/helpers.js"></script>
+    <%= yield :head %>
+    <script type="text/javascript" src="/uv/lib/offline.js"></script>
+    <script type="text/javascript" src="/uv/lib/offline.js"></script>
+    <script type="text/javascript" src="/uv/helpers.js"></script>
   </head>
   <body>
     <%= yield %>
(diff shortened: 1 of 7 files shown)


app/views/layouts/viewer_layout.html.erb
```diff
@@ -3,13 +3,14 @@
...
+    <%= yield :head %>
+    <script type="text/javascript" src="/uv/lib/offline.js"></script>
+    <script type="text/javascript" src="/uv/lib/offline.js"></script>
+    <script type="text/javascript" src="/uv/helpers.js"></script>
   </head>
   <body>
     <%= yield %>
     <%= javascript_include_tag 'viewer' %>
-    <script type="text/javascript" src="uv/uv.js"></script>
+    <script type="text/javascript" src="/uv/uv.js"></script>
```
hackmastera, line 14: Was this previously broken? What bug was observed?

spec/controllers/viewer_controller_spec.rb
```diff
@@ -9,7 +9,29 @@
...
+    context "when the user is logged in" do
+      it "redirects back to the viewer" do
+        sign_in FactoryBot.create(:user)
+        resource = FactoryBot.create_for_repository(:complete_campus_only_scanned_resource)
+
+        get :auth, params: { id: resource.id.to_s }
+
+        expect(response).to redirect_to viewer_index_path(anchor: "?manifest=http://www.example.com/concern/scanned_resources/#{resource.id}/manifest")
+      end
     end
```
hackmastera, line 35: Test if the user is not logged in but the resource is private / non-cdl?

app/views/layouts/viewer_layout.html.erb
tpendragon: Yeah - when it was on auth you'd get 404s for these, since it was trying to load '/auth/uv/uv.js'

spec/controllers/viewer_controller_spec.rb
tpendragon: Sure, added.

PR #4046, opened 2020-07-02
Clean up expired charges when ChargeManager is initialized

### LINKED ISSUE #4039: CDL::ChargeManager needs to clean up ChargedItems
Ways to end a charge:

* patron uses button to return the digital item
* charge expires

Options:

1. A cleanup cron that checks for expired ChargedItems.
1. Every time the charge manager is initialized, it runs a cleanup method. It doesn't have to save it can just delete old ChargedItems from the in-memory array.

We like option 2. Note that a future notification cron will kick this off as well in this case.

hackmastera:
Seeking feedback on refactors or adjustments

fixes #4039

### DIFF
--- a/app/resources/cdl/charge_manager.rb
+++ b/app/resources/cdl/charge_manager.rb
@@ -10,6 +10,11 @@ module CDL
       @resource_id = resource_id
       @eligible_item_service = eligible_item_service
       @change_set_persister = change_set_persister
+      clear_expired_charges
+    end
+
+    def clear_expired_charges
+      resource_charge_list.charged_items = resource_charge_list.charged_items.reject(&:expired?)
     end
 
     def available_for_charge?
--- a/spec/resources/cdl/charge_manager_spec.rb
+++ b/spec/resources/cdl/charge_manager_spec.rb
@@ -21,6 +21,29 @@ describe CDL::ChargeManager do
 
   let(:change_set_persister) { ScannedResourcesController.change_set_persister }
 
+  describe "#initialize" do
+    it "clears expired charges from in-memory array" do
+      stub_bibdata(bib_id: "123456")
+      resource = FactoryBot.create_for_repository(:scanned_resource, source_metadata_identifier: "123456")
+      eligible_item_service = EligibleItemService.new(item_ids: ["1234"])
+
+      Timecop.freeze(Time.current - 1.hour) do
+        charged_items = [
+          FactoryBot.build(:charged_item, expiration_time: Time.current + 1.hour)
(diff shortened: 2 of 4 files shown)


app/resources/cdl/charge_manager.rb
```diff
@@ -10,6 +10,22 @@ def initialize(resource_id:, eligible_item_service:, change_set_persister:)
       @resource_id = resource_id
       @eligible_item_service = eligible_item_service
       @change_set_persister = change_set_persister
+      clean_expired_charges
+    end
+
+    def clean_expired_charges
+      resource_charge_list.charged_items = resource_charge_list.charged_items.map do |item|
+        if expired?(item)
+          change_set_persister.delete(change_set: ChargedItemChangeSet.new(item))
```
tpendragon, line 19: ChargedItems aren't persisted, they're just embedded, so I don't think you need to do this.

app/resources/cdl/charge_manager.rb
```diff
@@ -10,6 +10,22 @@ def initialize(resource_id:, eligible_item_service:, change_set_persister:)
       @resource_id = resource_id
       @eligible_item_service = eligible_item_service
       @change_set_persister = change_set_persister
+      clean_expired_charges
+    end
+
+    def clean_expired_charges
+      resource_charge_list.charged_items = resource_charge_list.charged_items.map do |item|
```
tpendragon, line 17: I think you can just do `resource_charge_list.charged_items = resource_charge_list.charged_items.select(&:expired?)`, and then add the `expired?` method to the `ChargedItem` class.

spec/resources/cdl/charge_manager_spec.rb
```diff
@@ -21,6 +21,32 @@ def item_ids(source_metadata_identifier:)
...
+    end
+
+    it "cleans up expired charges" do
+      stub_bibdata(bib_id: "123456")
+      resource = FactoryBot.create_for_repository(:scanned_resource, source_metadata_identifier: "123456")
+      eligible_item_service = EligibleItemService.new(item_ids: ["1234"])
+
+      Timecop.freeze(Time.current - 1.hour) do
+        charged_items = [
+          FactoryBot.create_for_repository(:charged_item, item_id: "1234", netid: "zelda", expiration_time: Time.current + 1.hour)
```
tpendragon, line 36: You can just do `FactoryBot.build`, since these are just embedded in the resource.

hackmastera:
Thanks @tpendragon. This is ready for re-review.

PR #4047, opened 2020-07-02
Ensure that PDFs staged for OCR match ILLiad naming scheme

### LINKED ISSUE #4045: Filter PDFs to be OCRed
Need to filter PDFs in the the filewatcher so that only files that match the ILLiad naming pattern are processed. Easiest thing to do is test that the file name has only numerals.

eliotjordan:
Closes #4045


eliotjordan:
@tpendragon Yeah, testing is difficult (but probably not impossible) because of the way the filewatcher runs. I'll create a ticket to look into this.

PR #4052, opened 2020-07-02
Rescue ActiveJob::DeserializationError

### LINKED ISSUE #4037: ActiveStorage::AnalyzeJob raises exception when BrowseEverything or OcrRequest blob is purged
[Figgy/production] ActiveJob::DeserializationError: Error while trying to deserialize arguments: Couldn't find ActiveStorage::Blob with 'id'=129 

## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/64053021)

Components that utilize ActiveStorage (BrowseEverything and OcrRequests) will delete the uploaded blob before ActiveStorage::AnalyzeJob has had time to run. 

Options:

1. Suppress analyze jobs altogether. 
1. Find a way to rescue ActiveJob::DeserializationError on AnalyzeJob so an exception is not raised and the job will complete without complaint. Maybe override the job itself?

eliotjordan:
Closes #4037

### DIFF
--- a/config/initializers/analyze_job.rb
+++ b/config/initializers/analyze_job.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Monkey patch AnalyzeJob so that an ActiveJob::DeserializationError is rescued and
+# does not cause the job to be moved to the retry queue. This is used mainly for
+# OcrRequest blobs that are purged before AnalyzeJob is run.
+ActiveStorage::AnalyzeJob.class_eval do
+  discard_on ActiveJob::DeserializationError
+end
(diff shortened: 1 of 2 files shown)


config/initializers/analyze_job.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Monkey patch AnalyzeJob so that an ActiveJob::DeserializationError is rescued and
+# does not cause the job to be moved to the retry queue. This is used mainly for
+# OcrRequest blobs that are purged before AnalyzeJob is run.
+ActiveStorage::AnalyzeJob.class_eval do
+  discard_on ActiveJob::DeserializationError
```
jrgriffiniii, line 7: Please see https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-discard_on for reference.

PR #4057, opened 2020-07-06
Add CDL Charge Status API

### LINKED ISSUE #4056: CDL Status Endpoint
We need an endpoint to check the status of a charge for the current user so we can expire the viewer. A simple JSON API would be the following:

GET `/cdl/<resourceid>/status`
```json
{
  "charged": true/false,
  "available": false,
  "expires_at": "<utctimestamp>" # Only if it's currently charged by the user logged in.
}
```

tpendragon:
Closes #4056

### DIFF
--- a/app/resources/cdl/status.rb
+++ b/app/resources/cdl/status.rb
@@ -31,7 +31,7 @@ module CDL
     end
 
     def charged_item
-      charge_manager.charged_items.find do |charged_item|
+      @charged_item ||= charge_manager.charged_items.find do |charged_item|
         charged_item.netid == user.uid
       end
     end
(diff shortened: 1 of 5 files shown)


app/resources/cdl/status.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def expired_hash
+      return {} unless charged_item.present?
+      {
+        expires_at: charged_item.expiration_time.to_i
+      }
+    end
+
+    def charged_item
+      charge_manager.charged_items.find do |charged_item|
```
eliotjordan, line 34: suggestion: I wonder if this should be memoized.
It's potentially called three times for every `as_json` call.

hackmastera:
I realize this is merged already but is there no case where both `charged` and `available` would be false?

PR #4065, opened 2020-07-08
Add Expiration Timer to UV for CDL

### LINKED ISSUE #4032: When a user's CDL checkout expires the page should disable the viewer.
We can likely control this via kicking off a JS timer with the remaining time when they get to the page, and deleting the viewer/refreshing the page when it expires.

tpendragon:
This PR also includes a fix to `Ability` which wasn't blocking expired charges from seeing the item.

Closes #4032

Screen Shot 2020-07-07 at 4 36 47 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,9 +1,13 @@
 module.exports = {
-    "extends": [
-      "standard",
-      'plugin:vue/recommended'
-    ],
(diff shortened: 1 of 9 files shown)


tpendragon:
Screen Shot 2020-07-08 at 2 00 59 PM

PR #4097, opened 2020-07-14
Adds ResourceChargeList#hold_queue

### LINKED ISSUE #4088: Add `CDL::ResourceChargeList#hold_queue`
## Implementation Notes

The hold queue should be an array of `CDL::Hold`s

### CDL::Hold 

`CDL::Hold` should have the following properties:

`netid` - NetID of user with hold.
`expiration` - Expiration of the hold. Is null until notified.
`notification_time` - Time when the user was notified.

It should respond to `#active?`, which returns true if `expiration` is set, signaling that the user has a certain amount of time (12 hours currently) to convert the hold to a charge.

### CDL::ChargeManager

`CDL::ChargeManager#available_for_charge?` should be updated to return false if `hold_count + charge_count` >= `item_ids.count`

It should also, similar to how it does for charges, automatically remove holds which have gone past their expiration time.

All of the above should be done before this ticket can be closed.

tpendragon:
Closes #4088

* Adds a CDL::Hold to store a hold.
* Adds a `hold_queue` to `CDL::ResourceChargeList`
* Adds `netid:` to `available_for_charge` so it can take into account existing active holds.
* Takes the `hold_queue` into account for `available_for_charge?`
* Clear out `hold_queue` when charging an item.

### DIFF
--- a/spec/resources/cdl/charge_manager_spec.rb
+++ b/spec/resources/cdl/charge_manager_spec.rb
@@ -102,7 +102,7 @@ describe CDL::ChargeManager do
         charge_manager = described_class.new(resource_id: resource.id, eligible_item_service: eligible_item_service, change_set_persister: change_set_persister)
 
         expect(charge_manager.eligible?).to eq false
-        expect(charge_manager.available_for_charge?).to eq false
+        expect(charge_manager.available_for_charge?(netid: "miku")).to eq false
       end
     end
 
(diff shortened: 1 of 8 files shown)


spec/resources/cdl/charge_manager_spec.rb
```diff
@@ -60,6 +60,20 @@ def item_ids(source_metadata_identifier:)
           reloaded_charges = Valkyrie.config.metadata_adapter.query_service.find_by(id: resource_charge_list.id)
           expect(reloaded_charges.charged_items).to be_present
         end
+        it "removes any existing holds for that netid" do
```
christinach, line 63: These are existing holds for the same item_id, correct? If so can you update the spec title to: removes any existing holds for the same item_id for that netid

tpendragon: No, holds aren't for a specific item_id, just the bib as a whole.

christinach:
@tpendragon it looks great. I only added a comment to update the spec for the specific case.

spec/resources/cdl/charge_manager_spec.rb
christinach: oh ok then!

PR #4103, opened 2020-07-15
Add ChargeManager#create_hold

### LINKED ISSUE #4089: Add `ChargeManager#create_hold(netid:)`
## Implementation Notes

This should add a `CDL::Hold` with no expiration time or notification time to the appropriate ResourceChargeList's `hold_queue`.

tpendragon:
Closes #4089

### DIFF
--- a/spec/resources/cdl/charge_manager_spec.rb
+++ b/spec/resources/cdl/charge_manager_spec.rb
@@ -45,6 +45,86 @@ describe CDL::ChargeManager do
     end
   end
 
+  describe "#create_hold" do
+    context "it is not available for charge" do
+      context "there is a ResourceChargeList and an existing unexpired hold for that user" do
+        it "doesn't do anything" do
+          charged_items = [
+            CDL::ChargedItem.new(item_id: "1234", netid: "skye", expiration_time: Time.current + 3.hours)
+          ]
+          holds = [
+            CDL::Hold.new(netid: "zelda")
+          ]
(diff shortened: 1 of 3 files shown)


spec/resources/cdl/charge_manager_spec.rb
```diff
@@ -45,6 +45,86 @@ def item_ids(source_metadata_identifier:)
...
+          charge_manager = described_class.new(resource_id: resource.id, eligible_item_service: eligible_item_service, change_set_persister: change_set_persister)
+
+          charge_manager.create_hold(netid: "zelda")
+          reloaded_charges = Valkyrie.config.metadata_adapter.query_service.find_by(id: resource_charge_list.id)
+          expect(reloaded_charges.hold_queue.length).to eq 1
+        end
+      end
+      context "there is a ResourceChargeList and an existing expired hold for that user" do
+        it "creates a new hold" do
+          charged_items = [
```
christinach, line 72: Do we keep in `Hold` an expired hold?

christinach: Maybe we’ve said that and I don’t remember. I thought when a hold expires - for a user - we remove it but now I see that a user can create a hold on an expired item.

tpendragon: Sorry - yeah. We keep the hold here because the cron job will remove it, so that it only happens in one place and the notification can be sent that it's expired.

christinach: thanks!

PR #4104, opened 2020-07-16
G and b

cwulfman:
Adds a pair of Rake tasks to ingest Grounds and Buildings metadata


app/services/grounds_and_buildings_service.rb
```diff
@@ -0,0 +1,38 @@
...
+  def add_members_to_mvw(mvw)
+    change_set = ChangeSet.for(mvw)
+    change_set.validate(member_ids: children(mvw.source_metadata_identifier.first))
+    change_set_persister.save(change_set: change_set)
+  end
+end
+
+private
+
+def figgy_id(url)
```
tpendragon, line 36: This should be inside the class definition.

app/services/grounds_and_buildings_service.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+require "csv"
+class GroundsAndBuildingsService
```
tpendragon, line 3: Can you add a comment about what this service does at a high level?

cwulfman:
Throwing this away; replaced with simpler PR.

PR #4122, opened 2020-07-22
PDF Derivative Service

### LINKED ISSUE #4067: PDFs should be split into page images
There are two cases where PDFs don't meet our needs for end-user display:
1. For CDL, PDFs are downloadable and shouldn't be
2. For very large PDF files, the viewer performance is much worse than page images.

So we should split PDFs into page images and use those images to drive the viewer. This will provide a better experience and be compatible with our CDL requirements.

## Technical Requirements

The PDF should be converted to Tiffs
The original PDF should be preserved.
The PDF shouldn't show up in the viewer.

## Possible Implementations

1. Don't embed PDFs at all in manifests anymore, and just ignore PDFs. (Don't do this if there's a time lag between this and #4119)
2. Mark the PDF as a preservation master, and each of the tiffs as an intermediate file.
3. Store the PDF on the resource or in some other thing not collected as a member.

hackmastera:
Closes #4067


spec/services/manifest_builder_spec.rb
```diff
@@ -956,5 +978,16 @@ def logical_structure(file_set_id)
       expect(media_sequence["elements"][0]["@type"]).to eq "foaf:Document"
       expect(media_sequence["elements"][0]["label"]).to eq scanned_resource.title.first
     end
+    # it "builds a UV compatible PDF manifest" do
```
tpendragon, line 981: Very small, can you delete all this commented code?

hackmastera:
@tpendragon deleted the commented code; thanks for catching that.

PR #4123, opened 2020-07-22
Activate holds based on a cron job

tpendragon:
This runs a cron which activates all possible holds every 10 minutes, sending the appropriate emails.

There's a possible alternate implementation which would use scheduled jobs, but it'd be significantly more complex (although likely require less processing power than this solution.)

Closes #4123


tpendragon:
@hackmastera Sure, added.

PR #4128, opened 2020-07-23
Track events in google analytics for the CDL Hold Queue

### LINKED ISSUE #4099: Log CDL hold events
Log the following events on holds:

* Total number of holds per item
  * Total number of holds that successfully led to a digital checkout
  * Total number of holds that expired

Accomplish this by doing the following:

When a user makes a hold, send an analytics event with an &ev which is equal to the current number of holds on that item.

When a user charges an item, if they have a hold then name the event "hold-charge" (this way we know that a hold was converted to a charge)

When a hold is expired and a notification is sent, send an analytics event with the name "hold-expired"

### LINKED ISSUE #4100: Log CDL queueing events
Log requests to join the queue for already checked-out items, per item.

tpendragon:
Closes #4100, closes #4099

### DIFF
--- a/spec/resources/cdl/event_logging_spec.rb
+++ b/spec/resources/cdl/event_logging_spec.rb
@@ -68,4 +68,21 @@ RSpec.describe CDL::EventLogging do
       .with(body: "cid=1&ea=hold&ec=CDL-staff&el=12345&ev=1&t=event&tid=UA-15870237-29&ua=Figgy&v=1")).to have_been_made
     end
   end
+
+  describe ".google_hold_charged_event" do
+    let(:netid) { "abc123" }
+    let(:bibid) { "12345" }
+    before do
+      stub_request(:get, "https://bibdata.princeton.edu/patron/#{netid}")
+        .to_return(status: 200,
+                   body: file_fixture("bibdata/#{netid}.json").read, headers: { "Content-Type" => "application/json" })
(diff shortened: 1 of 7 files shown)


spec/resources/cdl/event_logging_spec.rb
```diff
@@ -51,4 +51,55 @@
...
+    let(:netid) { "abc123" }
+    let(:bibid) { "12345" }
+    before do
+      stub_request(:get, "https://bibdata.princeton.edu/patron/#{netid}")
+        .to_return(status: 200,
+                   body: file_fixture("bibdata/#{netid}.json").read, headers: { "Content-Type" => "application/json" })
+      stub_request(:post, "https://www.google-analytics.com/collect")
+      allow(SecureRandom).to receive(:uuid).and_return("1")
+    end
+    it "sends a hold-charged event to google analytics" do
```
hackmastera, line 99: should say "hold-expired"

tpendragon:
@hackmastera Can you take another look?

hackmastera:
@tpendragon looks good. Do you want it squashed or rebased on merge?

tpendragon:
Up to you, I'm not attached to the commits.

hackmastera:
@tpendragon sorry, I stepped away. Squash seems good. If I had a strong opinion I would have just done it one way or the other! thanks!

PR #4135, opened 2020-07-30
Rake tasks to ingest and update Grounds and Buildings images

cwulfman:
replaces old pull request

### DIFF
--- a/lib/tasks/grounds_and_buildings.rake
+++ b/lib/tasks/grounds_and_buildings.rake
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+require "csv"
+
+namespace :sip do
+  desc "Update finding-aid components with members"
+  task update_components: :environment do
+    collection_id = ENV["COLL"]
+    table_path = ENV["TABLE"]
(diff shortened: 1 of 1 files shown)


cwulfman:
@hackmastera that was the approach I took first, and it's what should happen ultimately, but I decided that, at this point, the code is too grounds-and-buildings specific to codify as a service.  I would like to go ahead and get Grounds and Buildings processed, and then generalize the approach with the next collection like this (Historic Postcards).  @kelea99 , @escowles, and I have been discussing this for a couple of weeks now.

I'm thinking now this isn't something to include in the Figgy codebase at all; it's really a private, one-off script.  What do you think?

hackmastera:
I think it's fine to have specific things in the codebase, and I'd rather have them there so that if something goes weird with the data we can see what was run.

Especially if you hope to generalize it in the future, putting it in the codebase up front is good.

We can always delete things later if we decide they shouldn't be there, and we have done this.

If there's a time concern w/r/t/ writing tests to get the materials in I won't block this PR on that requirement.

Issue #4133, opened 2020-07-28
CDL Automated Ingest


tpendragon:
As we need to handle bulk ingest of hundreds of PDFs to enable as-fast-as-possible scanning for CDL, we'd like a low-touch method of ingesting them. We accept that several of the objects that will be ingested this way and marked complete will likely have problems, and suffer from a lack of mediation.

## Implementation Details

1. Every night a cron job should run which ingests all PDFs from a directory. They should be ingested as a `CDL::ResourceChangeSet`, which will be used as a change set on top of a ScannedResource. The following defaults should be set:
    * Source Metadata Identifier is the name of the PDF.
    * Visibility: Private
    * Collection: https://figgy.princeton.edu/catalog/fad1929d-01ca-4888-aa32-f59e51bca7c9
    * Rights Statement: In Copyright
    * Download: None
    * ocr_language nil

2. Every hour a cron job runs which looks for every `CDL::ResourceChangeSet` which is pending, checks to see if all of its derivatives have been generated, and makes sure `ManifestBuilder` is able to generate a manifest from it. If it can, mark that resource complete.

Three Parts:

- [x] Add `CDL::ResourceChangeSet`
- [x] Add automatic ingester
- [ ] Add automatic completer.

kelynch:
Regarding open question #2, this is a good name, and the share has been created.  The figgy service account utilized in the current ocr_pdf workflow has been granted read/write access to the share.

escowles:
Some answers:

* 1: I've created a [CDL collection](https://figgy.princeton.edu/catalog/fad1929d-01ca-4888-aa32-f59e51bca7c9) that items should be added to.
* 3: Let's leave OCR language blank for now. I suspect we don't need to set this unless we want to do OCR to help creating a text alternative for accessibility reasons.
* 4: Yes, it's OK to email when an item is marked complete

tpendragon:
The automatic completer turned out to be more complicated than I expected. Priorities have shifted a bit, so I'm putting this back into Ready and moving to a high priority ticket.

PR #4157, opened 2020-08-04
Add Automatic CDL Ingester

tpendragon:
Work towards #4133 

Ansible PR: https://github.com/pulibrary/princeton_ansible/pull/1663

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,7 @@ group :development do
   gem "foreman"
   gem "listen"
   gem "ruby-prof"
+  gem "solargraph"
   gem "spring"
   gem "spring-commands-rspec"
   gem "web-console"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -163,6 +163,7 @@ GEM
     babel-transpiler (0.7.0)
       babel-source (>= 4.0, < 6)
       execjs (~> 2.0)
+    backport (1.1.2)
     bagit (0.4.2)
       docopt (~> 0.5.0)
       validatable (~> 1.6)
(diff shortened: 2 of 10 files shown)


hackmastera:
Reviewing:

in spec/resources/cdl/automatic_ingester_spec.rb:12, shouldn't the code be doing this? It doesn't seem like it should be part of test setup. Line 18 is testing that this happened.

One general question. A file goes into the `cdl_in_path` and is moved to `cdl_in_path/ingesting` and then ingested via move from there. If the file is corrupted during these moves it seems like we lose the file entirely and it has to be re-scanned to be reingested -- is that correct? Is there some other guard against this?

tpendragon:
Line 12 is an extra file to make sure it doesn't reingest files that are currently ingesting by a job which is either running or in the retry queue. I'll add a comment. 

We've never really had a problem with file system copies that I can think of. I can adjust it so that it doesn't remove the file until after the ingest is done if that makes you more comfortable?

hackmastera:
In that case on line 12, can you use a different filename.

We have definitely had problems that show up during derivative generation and aren't really caught until much later. I'm not sure but I think sometimes the studio has found those files and they looked fine so they just ingested them again. That's the scenario I'm thinking of.

tpendragon:
I just realized I forgot to get this to add it to a collection anyways. After I do that maybe we can chat about this on zoom?

hackmastera:
Do you think this should get a release note?

Regardless, :thumbsup: approved

tpendragon:
I think the release note should be for when the whole feature's done, personally.

PR #4182, opened 2020-08-12
Collection Restricted Viewers

### LINKED ISSUE #4181: Collections should have a `restricted_viewers` property, allowing those with matching NetIDs to view complete non-private items.
Use Case: I'm running a class and have a collection of reading-room only materials I want only members of the class to have access to. There are too many items to individually track, or there may be items that are constantly being added to that collection. I can provide a list of the student NetIDs who are taking the class.

## Technical Implementation

Collections need a `restricted_viewers` property.

Update `Ability` to do the following:

* If a resource is NOT PRIVATE and is a member of a collection which have a subset of users that can access it, I should be able to discover it.
* If a resource is NOT PRIVATE and I'm logged in as a user on that collection's restricted viewers, I should be able to READ it.
If a resource is NOT PRIVATE and I'm logged in as a user on that collection's restricted viewers, I should not be able to download it.

Update viewer to show the following login text if able to DISCOVER, not READ, and the collections it's a part of have a list of restricted viewers:

`This material is restricted to authorized users. If you have access to this material for a class or some other reason please Login."

## Side Effects

Items of collections that have a list of restricted viewers will sync to DPUL.

tpendragon:
Closes #4181

### DIFF
--- a/app/views/viewer/auth.html.erb
+++ b/app/views/viewer/auth.html.erb
@@ -6,7 +6,8 @@
   <% if @charge_manager.eligible? %>
     <%= link_to 'Princeton Users: Log In to Digitally Check Out', main_app.new_user_session_path, class: 'btn btn-primary', id: "login" %>
   <% elsif restricted_viewership? %>
-    <%= link_to 'Authorized Users: Log In to View Restricted Material', main_app.new_user_session_path, class: 'btn btn-primary', id: "login" %>
+    <h1>Access to this material is limited to specific classes. Use your Princeton credentials to login.</h1>
+    <%= link_to 'Login', main_app.new_user_session_path, class: 'btn btn-primary', id: "login" %>
   <% else %>
     <%= link_to 'Princeton Users: Log in to View', main_app.new_user_session_path, class: 'btn btn-primary', id: "login" %>
   <% end %>
--- a/app/views/viewer/auth.html.erb
+++ b/app/views/viewer/auth.html.erb
@@ -5,6 +5,8 @@
 <div id="loginContainer">
   <% if @charge_manager.eligible? %>
     <%= link_to 'Princeton Users: Log In to Digitally Check Out', main_app.new_user_session_path, class: 'btn btn-primary', id: "login" %>
+  <% elsif restricted_viewership? %>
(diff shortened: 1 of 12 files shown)


spec/models/ability_spec.rb
```diff
@@ -139,6 +140,21 @@
   let(:non_contributor_ephemera_folder) { FactoryBot.create_for_repository(:ephemera_folder) }
   let(:non_contributor_ephemera_box) { FactoryBot.create_for_repository(:ephemera_box, member_ids: contributor_ephemera_folder_in_box.id) }
   let(:non_contributor_ephemera_folder_in_box) { FactoryBot.create_for_repository(:ephemera_folder) }
+  let(:restricted_viewer_collection) do
+    FactoryBot.create_for_repository(:collection, restricted_viewers: [current_user&.uid || "rando"])
```
hackmastera, line 144: Why this `or`?

app/views/viewer/auth.html.erb
```diff
@@ -5,6 +5,8 @@
 <div id="loginContainer">
   <% if @charge_manager.eligible? %>
     <%= link_to 'Princeton Users: Log In to Digitally Check Out', main_app.new_user_session_path, class: 'btn btn-primary', id: "login" %>
+  <% elsif restricted_viewership? %>
+    <%= link_to 'Authorized Users: Log In to View Restricted Material', main_app.new_user_session_path, class: 'btn btn-primary', id: "login" %>
```
escowles, line 9: Special Collections asked for the text to be "Access to this material is limited to specific classes. Use your Princeton credentials to login."

spec/models/ability_spec.rb
tpendragon: There's no current_user for the anonymous tests, and it's important for the testing that `restricted_viewers` not be empty.

hackmastera: I see. Could you add this as a comment; it doesn't look like it was a pre-existing convention in this spec.

tpendragon: The CDL items do it, but yeah I'm happy to add a comment.

tpendragon:
Screen Shot 2020-08-13 at 9 46 31 AM

PR #4200, opened 2020-08-18
Add return action to CDL controller

### LINKED ISSUE #4083: Add CDL::CDLController#return
It should accept the Figgy ID and delete the appropriate charge from the ResourceChargeList for that user.

It should do this via calling `ChargeManager#return(netid:)`, which should be implemented as part of this.

hackmastera:
closes #4083

### DIFF
--- a/spec/resources/cdl/charge_manager_spec.rb
+++ b/spec/resources/cdl/charge_manager_spec.rb
@@ -421,4 +421,47 @@ describe CDL::ChargeManager do
       end
     end
   end
+
+  describe "#return" do
+    context "it is not charged to the user" do
+      it "raises CDL::NotCharged" do
+        eligible_item_service = EligibleItemService.new(item_ids: ["1234"])
+        stub_bibdata(bib_id: "123456")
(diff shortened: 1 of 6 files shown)


spec/resources/cdl/charge_manager_spec.rb
```diff
@@ -421,4 +421,57 @@ def item_ids(source_metadata_identifier:)
...
+  describe "#return" do
+    context "it is not charged to the user" do
+      it "raises CDL::NotCharged" do
+        eligible_item_service = EligibleItemService.new(item_ids: ["1234"])
+        stub_bibdata(bib_id: "123456")
+        resource = FactoryBot.create_for_repository(:scanned_resource, source_metadata_identifier: "123456")
+
+        User.create!(uid: "zelda", email: "zelda@princeton.edu")
+
+        # resource_charge_list = FactoryBot.create_for_repository(
```
tpendragon, line 434: Can you remove these comments?

hackmastera: dang it! yes.

hackmastera: removed

PR #4201, opened 2020-08-18
Automatic completion

### LINKED ISSUE #4133: CDL Automated Ingest

tpendragon:
Closes #4133

### DIFF
--- a/app/resources/cdl/automatic_approver.rb
+++ b/app/resources/cdl/automatic_approver.rb
@@ -68,6 +68,7 @@ module CDL
 
       def pages_match?
         first_member.mime_type.include?("application/pdf")
+        # The first member is the PDF itself, so do member_count - 1.
         pdf_page_count == resource.member_ids.size - 1
       end
 
--- a/app/resources/cdl/automatic_approver.rb
+++ b/app/resources/cdl/automatic_approver.rb
@@ -20,12 +20,19 @@ module CDL
     end
 
     def run
+      resources = []
       pending_cdl_resources.each do |pending_resource|
         next unless ready_for_completion?(pending_resource)
         change_set = ChangeSet.for(pending_resource)
         change_set.validate(state: "complete")
-        change_set_persister.save(change_set: change_set)
+        resource = change_set_persister.save(change_set: change_set)
+        resources += [resource]
       end
+      notify(resources: resources)
+    end
+
+    def notify(resources:)
+      CDL::CompleteMailer.with(resource_ids: resources.map(&:id).map(&:to_s)).resources_completed.deliver_later
     end
 
     def pending_cdl_resources
--- a/app/resources/cdl/automatic_approver.rb
+++ b/app/resources/cdl/automatic_approver.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+module CDL
+  class AutomaticApprover
+    def self.run
+      change_set_persister.buffer_into_index do |buffered_change_set_persister|
+        new(change_set_persister: buffered_change_set_persister).run
+      end
+    end
+
+    def self.change_set_persister
+      ScannedResourcesController.change_set_persister
+    end
+
+    delegate :metadata_adapter, to: :change_set_persister
+    delegate :query_service, to: :metadata_adapter
+    attr_reader :change_set_persister
+    def initialize(change_set_persister:)
+      @change_set_persister = change_set_persister
+    end
+
+    def run
+      pending_cdl_resources.each do |pending_resource|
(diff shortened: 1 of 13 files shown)


app/resources/cdl/automatic_approver.rb
```diff
@@ -0,0 +1,94 @@
...
+      def first_member
+        @first_member ||= query_service.find_by(id: resource.member_ids.first)
+      end
+
+      def pdf_file_path
+        @pdf_file_path ||= Valkyrie::StorageAdapter.find_by(id: first_member.primary_file.file_identifiers.first).disk_path
+      end
+
+      def pdf_page_count
+        @pdf_page_count ||= Vips::Image.pdfload(pdf_file_path.to_s, access: :sequential, memory: true).get_value("pdf-n_pages")
```
hackmastera, line 90: I wonder if we should consider saving this value on the file set during derivatives generation, or even characterization.

app/resources/cdl/automatic_approver.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      # All child file sets are processed.
+      def all_processed?
+        query_service.custom_queries.find_deep_children_with_property(resource: resource, model: FileSet, property: :processing_status, value: "in process", count: true).zero?
+      end
+
+      def pages_match?
+        first_member.mime_type.include?("application/pdf")
+        pdf_page_count == resource.member_ids.size - 1
```
hackmastera, line 71: Why are they off by one? is page count zero-indexed? Maybe it would be more clear to add 1 to the value in `pdf_page_count` instead of subtracting here.

app/resources/cdl/automatic_approver.rb
tpendragon: During characterization makes sense to me, but it would be a big thing to add right now.

app/resources/cdl/automatic_approver.rb
tpendragon: I don't think it would be - there's one less member size, not one less PDF page size. It's because the first member is the PDF itself.

hackmastera: Aha, of course, yes a comment will be helpful thank you.

app/resources/cdl/automatic_approver.rb
hackmastera: Yes I agree I don't think we should do it right now. Maybe an issue with the 'optimization' label.

spec/resources/cdl/automatic_completer_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+require "rails_helper"
+
+RSpec.describe CDL::AutomaticCompleter do
+  let(:file) { fixture_file_upload("files/sample.pdf", "application/pdf") }
+  let(:query_service) { Valkyrie.config.metadata_adapter.query_service }
+  let(:change_set_persister) { ScannedResourcesController.change_set_persister }
+  describe ".run" do
+    context "when there's in process CDL items" do
+      it "does not approve them" do
```
hackmastera, line 11: Can you also change these test names to say "complete" instead of "approve"?

app/resources/cdl/automatic_approver.rb
tpendragon: https://github.com/pulibrary/figgy/issues/4203

spec/resources/cdl/automatic_completer_spec.rb
tpendragon: Sure, done.

PR #4216, opened 2020-08-24
PDF Derivative Service should extract/move files in batches.

### LINKED ISSUE #4215: Ingest of huge (> 2000 page PDFs) use up all the worker disk space.
Possible fixes:

* Make it dump the files to the isilon somewhere.
* Attach/extract the files in batches
* Make the PDF quality worse

tpendragon:
The lazy array makes it extract groups of 200 files at a time, then adds
them to the parent (in a transaction), which moves the files off the tmp
location. This should take care of the issue where we run out of space,
but if the extraction fails we may end up with orphan files in
production.

Closes #4215

### DIFF
--- a/app/derivative_services/pdf_derivative_service.rb
+++ b/app/derivative_services/pdf_derivative_service.rb
@@ -36,10 +36,12 @@ class PDFDerivativeService
 
   def add_file_sets(files)
     resource = parent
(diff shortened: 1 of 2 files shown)


tpendragon:
ADR for unlinked files: https://github.com/pulibrary/figgy/pull/4220

PR #4217, opened 2020-08-25
Don't process CDL Timer code if status check fails.

### LINKED ISSUE #4214: "Return this Item" showing on non-CDL items.
Seems to happen if you're logged into Figgy.

See https://catalog.princeton.edu/catalog/2516638

tpendragon:
This prevents attempting to parse a timer that doesn't exist as well as
displaying the return button on error.

Closes #4214

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,12 @@ Release notes template:
 ## Removed
 
 -->
+# 2020-08-25
+
+## Fixed
+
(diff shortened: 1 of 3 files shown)


tpendragon:
Yeah, it is, but I think in general if something went wrong we wouldn't want a button to appear.

hackmastera:
👍 thanks. I'm went ahead and added to the description of that issue.

PR #4220, opened 2020-08-25
Add ADR about unlinked files.

### LINKED ISSUE #4219: ADR for unlinked files
We should document the fact that there may unlinked files now, and what implications there are for those.

tpendragon:
Closes #4219

### DIFF
--- a/architecture-decisions/0009-unlinked-files.md
+++ b/architecture-decisions/0009-unlinked-files.md
@@ -1,0 +1,33 @@
+# 8. Unlinked Files
+
+Date: 2020-08-25 / Archaeological
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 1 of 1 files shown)


escowles:
This is somewhat related to https://github.com/pulibrary/figgy/issues/549 — at least in the broader issue of there being a disconnect between different parts of persistence.

hackmastera:
@escowles We weren't sure that we wanted to do something like "find and delete unlinked files" right now. The transaction disk storage adapter is the issue that would probably make the most sense, but we were hesitant to make it since it would likely go straight to the icebox. That's partially why we wanted to have the ADR.

escowles:
@hackmastera That sounds like a reasonable assessment to me, so I'm going to merge this ADR to document what's happening now.

PR #4226, opened 2020-08-26
Disable share/download for CDL enabled items.

### LINKED ISSUE #4134: Don't show Share or IIIF widgets in CDL viewer
These will not work correctly with CDL content, so we should not show them.

tpendragon:
Closes #4134

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -85,16 +85,31 @@ class ApplicationController < ActionController::Base
   # Retrieve the viewer configuration for a given resource
   def viewer_config
     resource = find_resource(resource_id_param)
-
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -85,18 +85,33 @@ def guest_user
...
-
     respond_to do |format|
-      format.json { render json: config }
+      format.json { render json: viewer_config_value(resource) }
     end
   end
 
+  def viewer_config_value(resource)
+    return default_uv_config if full_viewer_capability?(resource)
+    downloads_disabled_uv_config.tap do |config|
```
hackmastera, line 95: nice use of tap!

PR #4232, opened 2020-08-29
Migrate postcards

cwulfman:
Implements an experimental class for ingesting the historical postcards collection (pudl0009) and similar legacy collections with item-level MODS data as ephemera.


cwulfman:
Backs out classes for ingesting postcards as ephemera; adds rake task to ingest a directory of Stroop-style METS/MODS instead.

cwulfman:
abandoning this approach.

PR #4256, opened 2020-09-11
adds rake task to import derrida files

cwulfman:
This is just a short-term task to get a few items up for stakeholder to see; not intended to be a permanent solution.

### DIFF
--- a/lib/tasks/bulk.rake
+++ b/lib/tasks/bulk.rake
@@ -102,6 +102,80 @@ namespace :bulk do
     end
   end
 
+  # Derrida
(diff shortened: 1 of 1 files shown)


escowles (commented):
The only changes I see from `bulk:ingest` are adding title and note fields — can we just add those fields to `bulk:ingest` instead?

cwulfman:
I didn't want to interfere with existing code; that would require clearance from Trey or Anna, right?

tpendragon:
As long as it's not a required property it seems fine to me.

PR #4259, opened 2020-09-15
Add Caching to Figgy Manifests.

tpendragon:
This will provide < 200ms responses for the second generation of commonly accessed manifests. Because of how OL works every manifest is requested twice, so this should provide significant performance gains (~ 4 seconds for > 800 page books.)

Requires https://github.com/pulibrary/princeton_ansible/pull/1731


Gemfile
```diff
@@ -8,6 +8,7 @@ gem "blacklight_range_limit", "~> 6.0"
 gem "bootstrap-sass", "~> 3.4"
 gem "bundler", "~> 2.0"
 gem "bunny"
+gem "dalli"
```
hackmastera, line 11: How is this used? I don't see the setup as documented in https://github.com/petergoldstein/dalli#usage-with-rails-3x-and-4x

app/resources/scanned_resources/scanned_resources_controller.rb
```diff
@@ -41,6 +41,12 @@ def manifest
     redirect_to manifest_scanned_resource_path(id: @resource.id.to_s)
   end
 
+  def cached_manifest(resource, auth_token_param)
+    Rails.cache.fetch("#{ManifestKey.for(resource)}/#{auth_token_param}") do
```
hackmastera, line 45: Could we add a test that illustrates the role of the auth token here? Showing what happens if the same resources is requested with / without. I think it's worth calling attention to beyond just having it in the method.

hackmastera (changes requested):

Gemfile
tpendragon: I think it was Rails 4 or 5 where Rails just started using it if it's available, no config necessary: https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-memcachestore

app/resources/scanned_resources/scanned_resources_controller.rb
tpendragon: Added a spec.

hackmastera (approved):

Issue #4254, opened 2020-09-10
Add a date facet to Figgy


kelynch:
The VIrtual Reading Room (VRR) [functional requirement](https://docs.google.com/spreadsheets/d/1z8Z_k42inQtAHarRrOfjMGzSwnYoCpIOr0J99RL6kqk/edit#gid=0) R006 that "components be searchable by component name, collection name, date, etc."

This metadata should currently be in the index, however to meet this requirement, it's been requested that we add a date facet to Figgy's search interface. 

## Implementation Details

* Use the "created" field for this facet.
* If possible, implement this facet with a slider (such as the [blacklight_range_limit plugin](https://github.com/projectblacklight/blacklight_range_limit)).
* This facet will just pull in the value of the start date, where there is otherwise a date range.

hackmastera:
date_created is a string field which according to our tests expects to handle values such as "circa 1865" and "1941-1985".

In order to facet these will need to be normalized into year-only dates. In order to use blacklight-range-limit they will need to be sortable, integer values.

Orangelight uses blacklight-range-limit by pulling the start date from the 008 ([logic here](https://github.com/pulibrary/marc_liberation/blob/493949186cbb8cf5f6bb888aced888cfca149051/marc_to_solr/lib/princeton_marc.rb#L59-L66)) And in Figgy that is the same logic used to populated "date created" (logic [here](https://github.com/pulibrary/figgy/blob/3c3283c47c3154398b55ddb83e7f025060686fef/app/services/pul_metadata_services/bib_record.rb#L19) and [here](https://github.com/pulibrary/figgy/blob/3c3283c47c3154398b55ddb83e7f025060686fef/app/services/pul_metadata_services/bib_record.rb#L86-L88)).

For finding aid items, figgy pulls the display date into the `date_created` field but we pull a normalized iso8601 date into the `created` field. It would be preferable to use this normalized value as the source for the facet.

Implementation proposal:

1. Create a new Valkyrie field on ScannedResource called `pub_date_start_sort` (this is what it's called in orangelilght) to use as the facet field. Then:
   1. Either pull the value during metadata synchronization.
   1. Or at index time check the source metadata identifier. if it's a bib number copy the `date_created` field into the new facet field. If it's a component ID extract the year from the `created` field.

tpendragon:
I have a super fuzzy memory of bibdata providing two date fields, one normalized. Is that not the case?

tpendragon:
[...]

PR #4260, opened 2020-09-16
Add index field pub_date_start_itsi

hackmastera:
advances #4254

### DIFF
--- a/app/indexers/facet_indexer.rb
+++ b/app/indexers/facet_indexer.rb
@@ -56,7 +56,7 @@ class FacetIndexer
     date =
       begin
         Time.zone.parse(date) unless date.is_a?(Time)
-      rescue TypeError
+      rescue TypeError, ArgumentError
         nil
       end
     return unless date
--- a/app/indexers/facet_indexer.rb
+++ b/app/indexers/facet_indexer.rb
@@ -10,7 +10,8 @@ class FacetIndexer
       {
         display_subject_ssim: resource.primary_imported_metadata.subject,
         display_language_ssim: imported_language,
-        has_structure_bsi: structure?
+        has_structure_bsi: structure?,
+        pub_date_start_itsi: pub_date_start
       }
     else
       {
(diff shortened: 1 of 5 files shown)


app/indexers/facet_indexer.rb
```diff
@@ -48,6 +49,14 @@ def structure?
     false
   end
 
+  def pub_date_start
+    date = resource.primary_imported_metadata.created
+    return unless date
```
tpendragon, line 54: Should this be `date.present?`? I assume created is a Set and could be an empty array.

app/indexers/facet_indexer.rb
```diff
@@ -48,6 +49,14 @@ def structure?
...
+    date = date.first
+    date = Time.zone.parse(date) unless date.respond_to?(:year)
```
tpendragon, line 56: Should we capture parsing errors here or do you think we should just wait for it to happen? Could end up failing to save objects.

spec/indexers/facet_indexer_spec.rb
```diff
@@ -11,8 +11,20 @@
 
         expect(output[:display_subject_ssim]).to eq scanned_resource.imported_metadata.first.subject
         expect(output[:display_language_ssim]).to eq ["English"]
+        expect(output[:pub_date_start_itsi]).to eq 1982
+      end
+
+      it "parses the first year from a date range" do
```
tpendragon, line 17: Do we need a similar test for a bibid date?

tpendragon (changes requested):

app/indexers/facet_indexer.rb
hackmastera: I added a test where the source doesn't have the created date and it goes okay. but `present?` works too and is more flexible so 👍

hackmastera:
@tpendragon I added a bunch more error handling. should be ready for another look.

tpendragon (approved):

PR #4262, opened 2020-09-17
Test FacetIndexer from Solr directly

hackmastera:
The tests for this class actually hit the class twice, once when the
record is created (and the metadata is pulled), and then again when
`described_class.to_solr` is called. The first time dates come through
as Time objects. The second time they come through as strings. There was
a bug in the code giving a nil value when the date was a Time object,
which the tests didn't catch because they were only checking values via
`described_class`. Adding a test that looks at the index directly gave a
failing test from which to resolve the bug.

### DIFF
--- a/app/indexers/facet_indexer.rb
+++ b/app/indexers/facet_indexer.rb
@@ -52,15 +52,16 @@ class FacetIndexer
   def pub_date_start
     date = resource.primary_imported_metadata.created
     return unless date.present?
-    date = date.first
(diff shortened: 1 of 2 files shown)


tpendragon (approved):

tpendragon:
@hackmastera Looks good, but there's a failing test.

tpendragon:
Oh the failing spec is mine. I'll deal with it.

PR #4266, opened 2020-09-21
Don't reindex PreservationObjects.

tpendragon:
There's millions, most of them point to FileSets, and we don't really
use their Solr records.


hackmastera (changes requested):
This looks good except the rubocop issue.

hackmastera (approved):

Issue #4185, opened 2020-08-13
I would like bulk add a new collection to all items within the Spainish 15M Ephemera Collection


kelea99:
New Collection Name to be added : [Latin American and Iberian Studies Collections](https://figgy.princeton.edu/catalog/e9709523-ec23-4eb0-9dd3-002746b91d13)

Collection that needs the new tag. **Important** : please retain the original collection tag as well. this will be used as a sub-collection facet in DPUL):  https://figgy.princeton.edu/catalog/70433c80-ca9e-4ade-9687-8c98f72875bc

I have also updated [#2862](https://github.com/pulibrary/figgy/issues/2862#issuecomment-673590757) with this use case.

tpendragon:
@kelea99 Do you want all folders, whether they're in a box or not, to be added to this collection?

kelea99:
@tpendragon , I'm about to have a look, but can you give an example of what you mean, to ensure I'm understanding correctly?

tpendragon:
@kelea99 Just confirming that, for example, https://figgy.princeton.edu/catalog/parent/bddbafd7-e08d-4eab-82d9-b585bd31661b/993573ea-cbb8-4538-94f4-e89cf62cf6eb should have the collection added and not just the boxless folders.

kelea99:
@tpendragon, correct. Boxed folders and boxless folders are both included in the Spain 15M dpul exhibit. The complete collection is needing that new collection tag, in addition to the Spain 15 M one. thanks!!

kelea99:
reopening this. I cant see the second collection and it isnt mapping into DPUL. can we try this again?

tpendragon:
@kelea99 Is this done now? Looking at a few they're in the collection.

kelea99:
Yes. this is complete. closing.

Issue #2862, opened 2019-04-10
Add "additional collections" to Bulk Edit form, to add all items to another collection.


kelea99:
is this possible? Example: adding another collection (ex: Edith Wharton) to items within an already assigned collection: https://figgy.princeton.edu/catalog?utf8=✓&q=Edith+Wharton+Manuscripts+C0118

escowles:
@kelea99 Yes, we plan to add a few more things to the bulk edit screen:
1. collection membership
2. rights statement
3. preservation policy

See #2586 and #2842

So far, we have assumed that all the items being bulk edited would be in a single collection, and we only show the bulk edit button when the search is limited to a collection (to limit the potential for accidental system-wide changes).  That is true for the Edith Wharton items (where they are all in the Manuscripts collection).  But I'd appreciate feedback on other scenarios where that might not be the case.

kelea99:
Absolutely, re: feedback. this one actually came from Squirrel in RBSC, and I thought it was a good idea as well (Edith Wharton's stuff is currently in 2 collections: MSS and one other, hence the want to create a new collection for them).

kelea99:
Another use case for this for all the Latin American and Iberian Studies Collections. the feature should have the functionality to add only those items that are not already in the collection to it.

Issue #2586, opened 2019-02-15
Expand bulk complete functionality


escowles:
- [ ] Add additional collections
- [x] Set rights statement and note
- [x] OCR Language

See #697 and #2550

cwulfman:
In re. additional collections see #2862, #4185

PR #4270, opened 2020-09-24
2586 bulk edit

cwulfman:
addresses #2586, #2862, and #4185 

- Extends bulk_edit_service to enable all items of a collection to be added to another collection.
- Provides a rake task to support this extension.

The GUI form is not addressed in this PR.

### DIFF
--- a/app/services/bulk_edit_service.rb
+++ b/app/services/bulk_edit_service.rb
@@ -20,8 +20,8 @@ class BulkEditService
     incorporated_attributes = {}
     proposed_attributes.each_key do |key|
       incorporated_attributes[key] = case key
-                                     when :member_of_collection_ids
-                                       existing_attributes[key] << proposed_attributes[key]
+                                     when :append_collection_ids
+                                       existing_attributes[:member_of_collection_ids] << proposed_attributes[key]
                                      else
                                        proposed_attributes[key]
                                      end
--- a/app/services/bulk_edit_service.rb
+++ b/app/services/bulk_edit_service.rb
@@ -8,7 +8,7 @@ class BulkEditService
     c.decorate.members.each do |member|
       logger.info "Updating attributes for #{member}"
       change_set = ChangeSet.for(member)
-      if change_set.validate(incorporate_proposed_attributes(member.attributes, attributes))
+      if change_set.validate(append_or_replace_attributes(member.attributes, attributes))
         change_set_persister.save(change_set: change_set)
       else
         logger.warn "  Failed validation: #{change_set.errors}"
@@ -16,7 +16,7 @@ class BulkEditService
     end
   end
 
-  def self.incorporate_proposed_attributes(existing_attributes, proposed_attributes)
+  def self.append_or_replace_attributes(existing_attributes, proposed_attributes)
     incorporated_attributes = {}
     proposed_attributes.each_key do |key|
       incorporated_attributes[key] = case key
--- a/app/services/bulk_edit_service.rb
+++ b/app/services/bulk_edit_service.rb
@@ -8,11 +8,24 @@ class BulkEditService
     c.decorate.members.each do |member|
       logger.info "Updating attributes for #{member}"
       change_set = ChangeSet.for(member)
-      if change_set.validate(attributes)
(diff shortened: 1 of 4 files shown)


app/services/bulk_edit_service.rb
```diff
@@ -8,11 +8,24 @@ def self.perform(collection_id:, attributes:, metadata_adapter: Valkyrie.config.
...
-      if change_set.validate(attributes)
+      if change_set.validate(incorporate_proposed_attributes(member.attributes, attributes))
         change_set_persister.save(change_set: change_set)
       else
         logger.warn "  Failed validation: #{change_set.errors}"
       end
     end
   end
+
+  def self.incorporate_proposed_attributes(existing_attributes, proposed_attributes)
```
hackmastera, line 19: I might call it something else like maybe "append_or_replace_attributes"?

hackmastera (commented):
It looks pretty good to me overall; see inline comment.

I wonder if @tpendragon would mind taking a look.

app/services/bulk_edit_service.rb
tpendragon: :+1: to the rename. It's more work, but I'd probably propose adding an `append_to_collection_id` virtual property similar to how we do `append_id`, so you can just set that property and we don't need special logic. I don't want to grind this work to a halt, though.

cwulfman: Done

cwulfman (commented):
I surveyed append_id; it seems to center on a subclass of ChangeSetPersister, AppendToParent (append_to_parent.rb), which is registered as an after-save handler in change_set_persister.   Am I getting warm?   Would the idea be to create an AppendToCollection handler that would be registered as a before_save handler in ChangeSetPersister?

tpendragon:
@cwulfman I wonder if instead of overriding the meaning of `member_of_collection_ids` as a simpler precaution in this class we can just change it so that it's like

```ruby
      incorporated_attributes[key] = case key
                                     when :append_collection_ids
                                       existing_attributes[:member_of_collection_ids] << proposed_attributes[key]
                                     else
                                       proposed_attributes[key]
                                     end
```

That way there's a difference in value for the rake task and you don't have to go digging around change set persister to get this in? I'm concerned if we just merge it this way we'll never look at it again until it bites us.

lib/tasks/bulk.rake
```diff
@@ -314,4 +314,15 @@ namespace :bulk do
...
   end
+
+  desc "Adds all members of a Collection to an additional collection"
+  task update_attrs: :environment do
+    coll = ENV["COLL"]
+    append_coll = ENV["APPEND_COLL"]
+
+    abort "usage: rake bulk:update_attrs COLL=[collection id] APPEND_COLL=[new collection id]" unless coll && append_coll
+    logger = Logger.new(STDOUT)
+    attrs = { member_of_collection_ids: Valkyrie::ID.new(append_coll) }
```
tpendragon, line 325: This should be :append_collection_ids now.

tpendragon (changes requested):

cwulfman:
The result of careless query-replace.  Fixed now

On Sep 24, 2020, at 20:12, Trey Pendragon <notifications@github.com<mailto:notifications@github.com>> wrote:



@tpendragon requested changes on this pull request.

________________________________

In lib/tasks/bulk.rake<https://github.com/pulibrary/figgy/pull/4270#discussion_r494677096>:

> @@ -314,4 +314,15 @@ namespace :bulk do
       @logger.error "Error: #{e.message}"
     end
   end
+
+  desc "Adds all members of a Collection to an additional collection"
+  task update_attrs: :environment do
+    coll = ENV["COLL"]
+    append_coll = ENV["APPEND_COLL"]
+
+    abort "usage: rake bulk:update_attrs COLL=[collection id] APPEND_COLL=[new collection id]" unless coll && append_coll
+    logger = Logger.new(STDOUT)
+    attrs = { member_of_collection_ids: Valkyrie::ID.new(append_coll) }


This should be :append_collection_ids now.

tpendragon (approved):

PR #4277, opened 2020-09-29
Run CDL a lot more often.

tpendragon:
This will ingest any CDL file that hasn't been modified in the last
hour. Should decrease maximum time between copy and ingest to two hours.

An unfortunate side effect of this is they'll get emails a lot more often, probably, and throughout the day.


kelynch (approved):
This looks good.  I don't think the extra notifications should be a problem either.

PR #4279, opened 2020-10-01
Adds from_csv task

cwulfman:
I'm opening this PR to start a discussion about how to do this properly.  The rake task from_csv is a placeholder; it should be replaced with a service/job pair. But my primary question has to do with how to handle the mapping of csv headers to resource attributes: as implemented here it is a casual hack.  Some sort of mapping table is one possibility.  Another (or an extension of the mapping-table idea) is to create an Ingestible class that can take something like a CSV row (or a hash, or whatever) and create a valid set of parameters for ingestion: class of resource, path to resource (validated to exist), required attributes, additional attributes. Many of these bulk rake tasks could then be refactored to work with Ingestibles.

Or perhaps I'm missing something entirely; perhaps this is already implemented in some way; perhaps you have better ideas?

### DIFF
--- a/lib/tasks/bulk.rake
+++ b/lib/tasks/bulk.rake
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 require "csv"
 
-
 namespace :bulk do
   desc "Migrates directory of METS files"
   task ingest_mets: :environment do
@@ -335,7 +334,7 @@ namespace :bulk do
     coll = ENV["COLL"]
     csvfile = ENV["CSV"]
 
-    abort "usage: COLL=collection_id BASEDIR=directory CSV=csvfile rake bulk:from_csv" unless coll && basedir 
+    abort "usage: COLL=collection_id BASEDIR=directory CSV=csvfile rake bulk:from_csv" unless coll && basedir
     abort "no such file #{csvfile}" unless File.file?(csvfile)
 
     user = User.find_by_user_key(ENV["USER"]) if ENV["USER"]
@@ -366,8 +365,7 @@ namespace :bulk do
         member_of_collection_ids: [coll],
         file_filters: filters,
         attributes: attrs
-        )
-      
+      )
     end
   end
 end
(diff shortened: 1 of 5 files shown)


lib/tasks/bulk.rake
```diff
@@ -325,4 +327,45 @@ namespace :bulk do
...
+      csv = CSV.read(csvfile, headers: true)
+    rescue => e
+      @logger.error "Error: #{e.message}"
+      @logger.error e.backtrace
+    end
+
+    logger.info "processing #{csv.length} rows"
+    csv.each do |row|
+      logger.info "processing #{row}"
+      attrs = {}
```
escowles, line 355: This could be simplified to:
```
attrs = row.map { |k, v| [k.to_sym, v] }.to_h
```

lib/tasks/bulk.rake
```diff
@@ -325,4 +327,45 @@ namespace :bulk do
...
+  task from_csv: :environment do
+    basedir = ENV["BASEDIR"]
+    coll = ENV["COLL"]
+    csvfile = ENV["CSV"]
+
+    abort "usage: COLL=collection_id BASEDIR=directory CSV=csvfile rake bulk:from_csv" unless coll && basedir
+    abort "no such file #{csvfile}" unless File.file?(csvfile)
+
+    user = User.find_by_user_key(ENV["USER"]) if ENV["USER"]
+    user = User.all.select(&:admin?).first unless user
```
escowles, line 341: There's a rubocop error about this not being used — and I'm looking throughout the file, and I don't see the `user` variable being used  in most of the places it's defined. Maybe we need to be adding `depositor: user` to most of the properties to make sure that gets in?

lib/tasks/bulk.rake
cwulfman: As an FP afficionado I should have seen that!  Done.

lib/tasks/bulk.rake
cwulfman: I copy/pasted the base of this task from somewhere else and this came along.  Removing it for now.

escowles (dismissed):

PR #4305, opened 2020-10-30
fixes #4304 removes unnecessary reference to LAE Areas vocab

### LINKED ISSUE #4304: ingesting ephemera: hard-coded vocab label breaks production system
https://github.com/pulibrary/figgy/blob/a3d741e794f25cfe3a0b6d58520005cadd994633/app/services/ingest_ephemera_csv.rb#L122

In the production environment, the vocabulary label is "LAE Geographic Areas".  This seems like a hack; there should be a way to find out what vocabularies are being used by a project and referring to them.

cwulfman:


app/services/ingest_ephemera_csv.rb
```diff
@@ -15,19 +14,16 @@ def initialize(project_id, mdata_file, imgdir, change_set_persister, logger)
...
     mdata_table.collect do |row|
+      change_set = BoxlessEphemeraFolderChangeSet.new(EphemeraFolder.new)
       folder_data = FolderData.new(base_path: imgdir, change_set_persister: change_set_persister, **row.to_h)
       change_set.validate(folder_data.attributes)
       change_set.validate(files: folder_data.files)
-      change_set.validate(append_id: project_id)
-      change_set_persister.save(change_set: change_set)
+      folder_data.member_of_collection_ids.each do |pid|
+        change_set.validate(append_id: pid.id)
+        change_set_persister.save(change_set: change_set)
```
escowles, line 23: I'm not sure I understand what's going on here — can't you do:
```
change_set.validate(member_of_collection_ids: folder_data.member_of_collection_ids)
```
Or would that not save to all the collections correctly?

tpendragon: I assume this is adding folders to projects, which is a member relationship and not a collection relationship.

app/services/ingest_ephemera_csv.rb
```diff
@@ -104,22 +103,44 @@ def language
...
-      vocab_service.find_subject_by(category: sub["category"], topic: sub["topic"]).id
+      subject = vocab_service.find_subject_by(category: sub["category"], topic: sub["topic"])
+      subject&.id
     end
   end
 
   def geo_subject
     return unless fields[:geo_subject].present?
-    Array(vocab_service.find_term(label: Array(fields[:geo_subject]).first, vocab: "LAE Areas"))
+    Array(vocab_service.find_term(label: Array(fields[:geo_subject]).first))
```
tpendragon, line 127: I think what'll happen here is this'll look right in the UI and everything, but in the edit form unless the project's geographic_subject is associated with whatever vocabulary this happens to find first, it won't be selectable or show up from the dropdown.

tpendragon (changes requested):
One question about LAE areas, but besides that can you rebase this? There's a ton of commits here that seem to just catch up to main.

app/services/ingest_ephemera_csv.rb
escowles: Hmm, I think we need to support both — projects are needed for the forms to work, and collections to have multiple collections for DPUL exhibits.

app/services/ingest_ephemera_csv.rb
cwulfman: This seemed like a hack to me, to hard-code the vocabulary for a particular project here, though I've come to understand that "LAE Areas" is actually a general-purpose vocabulary. 

I'll restore it.

cwulfman (commented):
changed it

app/services/ingest_ephemera_csv.rb
cwulfman: How do you suggest I support both?

escowles: I think the project membership and collection membership need to be handled separately:

For project:
```
change_set.validate(append_id: project_id)
```
For collections:
```
change_set.validate(member_of_collection_ids: folder_data.member_of_collection_ids)
```

cwulfman: Various rebases, etc. have confounded the code.  There shouldn't be a project_id parameter, because the projects are included in the spreadsheet/csv file, in conformance with the forms

PR #4311, opened 2020-11-09
Fix display of fixity checks from deleted resources.

### LINKED ISSUE #3814: Fixity Dashboard should render if there are events pointing to deleted resources
In staging there are events for resources that have since been deleted. This is okay - we keep events around because you can bring the object back from the dead if it was preserved, but the table should still render. We can just add some null checks to the table.

## Backtrace

    line 13 of [PROJECT_ROOT]/app/views/fixity_dashboard/_cloud_fixity_table.html.erb: block in _app_views_fixity_dashboard__cloud_fixity_table_html_erb__3972264412316033458_69912600229460
    line 11 of [PROJECT_ROOT]/app/views/fixity_dashboard/_cloud_fixity_table.html.erb: each
    line 11 of [PROJECT_ROOT]/app/views/fixity_dashboard/_cloud_fixity_table.html.erb: _app_views_fixity_dashboard__cloud_fixity_table_html_erb__3972264412316033458_69912600229460

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/62967771)

tpendragon:
Makes fixity dashboard display again.

Closes #3814

### DIFF
--- a/spec/features/fixity_dashboard_spec.rb
+++ b/spec/features/fixity_dashboard_spec.rb
@@ -32,6 +32,22 @@ RSpec.feature "Fixity dashboard" do
     end
   end
 
+  context "the fixity check for a resource by a cloud service provider is for a deleted resource" do
+    it "displays that the event was a success and that the preservation object was deleted" do
+      cloud_fixity_event
+      file_set
+      preservation_object
+      metadata_adapter.persister.delete(resource: file_set)
+      metadata_adapter.persister.delete(resource: preservation_object)
+      metadata_adapter.persister.delete(resource: scanned_resource)
+
+      visit fixity_dashboard_path
+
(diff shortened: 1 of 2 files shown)


spec/features/fixity_dashboard_spec.rb
```diff
@@ -32,6 +32,22 @@
     end
   end
 
+  context "the fixity check for a resource by a cloud service provider is for a deleted resource" do
+    it "displays that the event was a failure" do
```
hackartisan, line 36: How does this test check that it's in the failure table?

It seems to check for the text "SUCCESS" which is confusing given the title of the test.

hackartisan (changes requested):

spec/features/fixity_dashboard_spec.rb
tpendragon: @hackartisan Thanks for the catch! Fixed.

hackartisan (approved):
👍 🎉

PR #4313, opened 2020-11-10
Implements ingest of resources from a csv file

cwulfman:


app/services/ingest_ephemera_csv.rb
```diff
@@ -15,19 +14,16 @@ def initialize(project_id, mdata_file, imgdir, change_set_persister, logger)
...
+      change_set = BoxlessEphemeraFolderChangeSet.new(EphemeraFolder.new)
       folder_data = FolderData.new(base_path: imgdir, change_set_persister: change_set_persister, **row.to_h)
       change_set.validate(folder_data.attributes)
       change_set.validate(files: folder_data.files)
-      change_set.validate(append_id: project_id)
-      change_set_persister.save(change_set: change_set)
+      folder_data.member_of_collection_ids.each do |pid|
+        change_set.validate(append_id: pid.id)
+        change_set_persister.save(change_set: change_set)
+      end
```
escowles, line 24: I think this still needs to be changed:
* Even if the mapping is coming from the CSV file, the folder ID needs to be added to the project's `member_ids` property, and then the project needs to be saved.
* For collections, just assigning the IDs to the `member_of_collection_ids` property should be enough — there's no need to add them individually and save after each one.

cwulfman: fixed

escowles (dismissed):

escowles (approved):

escowles (approved):

PR #4330, opened 2020-12-01
Add local VPN IPs to Figgy's campus IPs

tpendragon:
The Global Protect client IPs look like these local ranges to our
servers.


hackartisan (approved):

hackartisan (changes requested):
Shouldn't `128.112.0.0/16` cover those last 5 entries?

Can you add another comment where the new IPs came from? The current comment doesn't cover the new values and I think it will be helpful to know how we got this list.

tpendragon:
@hackartisan I updated the comment. I agree `128.112.0.0/16` covers the last few entries, but I left them in because they're copy/pasted from the KB article. Seemed better than expecting whoever looks at this is going to explode the IP ranges.

hackartisan (approved):

PR #4350, opened 2020-12-11
implements add_ephemera_to_collection service

cwulfman:


app/services/add_ephemera_to_collection.rb
```diff
@@ -0,0 +1,39 @@
...
+    col_ids = folder.member_of_collection_ids + [collection_id]
+    change_set = ChangeSet.for(folder)
+    change_set_persister.save(change_set: change_set) if change_set.validate(member_of_collection_ids: col_ids)
+  end
+
+  def perform_now
+    project = query_service.find_by(id: project_id)
+    collection = query_service.find_by(id: collection_id)
+    logger.info("Adding members of Ephemera Project #{project.decorate.title} to Collection #{collection.decorate.title}")
+    project.decorate.members.each do |member|
```
escowles, line 30: I think you should use the `EphemeraProjectDecorator.folders` and `EphemeraProjectDecorator.boxes` methods  instead of doing the `case` logic.

app/services/add_ephemera_to_collection.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def process_folder(folder)
+    logger.info("Processing a folder: #{folder.decorate.title.first}")
+    col_ids = folder.member_of_collection_ids + [collection_id]
+    change_set = ChangeSet.for(folder)
+    change_set_persister.save(change_set: change_set) if change_set.validate(member_of_collection_ids: col_ids)
+  end
+
+  def perform_now
```
escowles, line 26: Can you use a different method name to avoid confusion with ActiveJob perform/perform_now? Having a pretty generic method like `execute` or whatever is fine, if there's nothing more semantically appropriate.

escowles (commented):
This looks pretty good to me. Just a couple of small comments.

app/services/add_ephemera_to_collection.rb
cwulfman: Yes -- the case logic had a strong code smell.

app/services/add_ephemera_to_collection.rb
cwulfman: Good idea

escowles (approved):

PR #4366, opened 2021-01-07
Restoring location to the IIIF manifest metadata hash

escowles:
Location was removed from the IIIF manifest metadata hash in a broader cleanup guided by feedback from the Cicognara project (#2358). But there was no substantive discussion of that field in particular, and that project no longer uses the metadata hash for its displays.

The Grounds and Buildings collection has item-level location information for thousands of photographs, and displaying this field will allow researchers to request viewing the original photographs.


tpendragon:
@escowles There's a small test failure, just need to increase the metadata length by 1 in a test it looks like.

tpendragon (approved):

PR #4377, opened 2021-01-11
Staff can delete their own FileSets.

### LINKED ISSUE #1367: Staff users can't delete FileSets they have created

### LINKED ISSUE #4379: `ability_spec.rb` tests its functionality without persisted resources, requiring code that never runs in production.
https://github.com/pulibrary/figgy/blob/dd172ae6404d12dee256c11112be286884bc2e4b/app/resources/ephemera_folders/ephemera_folder_decorator.rb#L59-L63

captures an ArgumentError solely because of

https://github.com/pulibrary/figgy/blob/dd172ae6404d12dee256c11112be286884bc2e4b/spec/models/ability_spec.rb#L624 and similar calls using `FactoryBot.create` instead of `FactoryBot.create_for_repository`, the former of which doesn't actually persist the resource.

tpendragon:
This also fixes up `ability.rb` so it uses the correct
`.create_for_repository`.

Closes #1367
Closes #4379

### DIFF
--- a/app/resources/ephemera_folders/ephemera_folder_decorator.rb
+++ b/app/resources/ephemera_folders/ephemera_folder_decorator.rb
@@ -58,8 +58,6 @@ class EphemeraFolderDecorator < Valkyrie::ResourceDecorator
   # TODO: Rename to decorated_ephemera_box
   def ephemera_box
     wayfinder.decorated_ephemera_box
-  rescue ArgumentError # Necessary because ability_spec.rb doesn't always save objects. TODO: Fix.
-    nil
   end
 
   def ephemera_project
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -25,7 +25,7 @@ class Ability
     can [:destroy], Template
     can [:destroy], OcrRequest
     can [:destroy], FileSet do |obj|
-      obj.depositor == [current_user.uid]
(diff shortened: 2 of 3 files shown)


hackartisan (commented):
👍🏻 but not sure what is going on with CI

tpendragon:
Coverage should be fixed.

app/resources/ephemera_folders/ephemera_folder_decorator.rb
```diff
@@ -58,8 +58,6 @@ def collections
   # TODO: Rename to decorated_ephemera_box
   def ephemera_box
     wayfinder.decorated_ephemera_box
-  rescue ArgumentError # Necessary because ability_spec.rb doesn't always save objects. TODO: Fix.
```
hackartisan, line 61: I don't really understand from this comment. Maybe you could write it in an issue?

hackartisan (changes requested):

app/resources/ephemera_folders/ephemera_folder_decorator.rb
tpendragon: Sure. See #4379.

hackartisan: Thank you!

hackartisan (approved):

PR #4380, opened 2021-01-12
Speed up IngestFolderLocator

tpendragon:
Find.find passed to an Enumerator#find finds a directory in staging in
about 4 seconds, where the previous method took 23 seconds.


hackartisan (approved):

hackartisan (approved):
cool!

PR #4391, opened 2021-01-26
Improves ephemera SIP ingestion

cwulfman:
Numerous fixes and additions to provide default values for fields in ephemera ingest


escowles (approved):

app/services/ingest_ephemera_csv.rb
```diff
@@ -80,7 +80,7 @@ def attributes
   # rubocop:enable Metrics/MethodLength
 
   def files
-    @files || Dir.glob("#{image_path}/*.{TIF,TIFF,tif,tiff,jpg,jpeg,png}").sort.map do |file|
+    @files || Dir.glob("#{image_path}/*.{tif,tiff,jpg,jpeg,png}").sort.map do |file|
```
escowles, line 83: Looks like this deletion is causing the test failure. Maybe downcasing `image_path` to avoid having to have case-sensitive matching would be a better approach?

cwulfman: Agreed; used File::FNM_CASEFOLD.

escowles (approved):

PR #4396, opened 2021-01-29
Pass parent object for CDL readability check for MVW child works

### LINKED ISSUE #4357: MVW CDL items are not viewable to some who have checked them out in the online catalog
We (Colleen Burlingham, @kelynch , and @kelea99 ) were not able to replicate this error that Colleen and @kelea99 witnessed with Jennifer Hunter this morning, who was not able to access any reels in the viewer for Trentonian, CDL, in the online catalog.  What we did:
Colleen:
1. we destroyed colleen's role in figgy.
2. Colleen checked to ensure her cache/history was clear in Firefox
3. Colleen navigated to the [Trentonian in the online catalog](https://catalog.princeton.edu/catalog/9670839)
4. She clicked on "Princeton Users: Log In to Digitally Check Out" 
5. This time she was directed to the same screen that Jennifer Hunter saw. This is different from those who have figgy roles, which takes one directly to the item/viewer.
6. The viewer appeared and we could see all reels available in the trentonian, along with the checkout button and timer in the upper left

We tried this same thing with Edge browser. Colleen was able to access the item again.

Jen Hunter, who does not have any figgy role assigned to her, could not access the content. she used Chrome:
1. Jen Navigated to the [Trentonian in the online catalog](https://catalog.princeton.edu/catalog/9670839)
2. She clicked on "Princeton Users: Log In to Digitally Check Out" 
3. she was directed to the same screen that Colleen saw when we removed her figgy role. This is different from those who have figgy roles, which takes one directly to the item/viewer.
4.  A black/blankviewer appeared, with nothing inside it, Jen could see the checkout button and timer in the upper left, which was counting down.

kelynch:
Closes #4357

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -161,7 +161,7 @@ class Ability
   end
 
   def cdl_readable?(obj)
-    resource_charge_list = Wayfinder.for(single_or_multivolume?(obj)).try(:resource_charge_list)
+    resource_charge_list = Wayfinder.for(parent_or_self?(obj)).try(:resource_charge_list)
     return false unless resource_charge_list
     resource_charge_list.charged_items.reject(&:expired?).map(&:netid).include?(current_user.uid)
   end
@@ -256,7 +256,7 @@ class Ability
     obj.decorate.downloadable? || authorized_by_token?(obj) || (!current_user.nil? && (current_user.staff? || current_user.admin?))
   end
 
-  def single_or_multivolume?(obj)
+  def parent_or_self?(obj)
     Wayfinder.for(obj).try(:parent) || obj
   end
 
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -161,6 +161,7 @@ class Ability
   end
 
   def cdl_readable?(obj)
+    obj = obj.cached_parent_id.present? ? find_by(id: obj.cached_parent_id.id) : obj if obj.respond_to?(:cached_parent_id)
     resource_charge_list = Wayfinder.for(obj).try(:resource_charge_list)
     return false unless resource_charge_list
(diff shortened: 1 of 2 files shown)


tpendragon:
We'll need a test, and we should use `Wayfinder.for(obj).parent`

app/models/ability.rb
```diff
@@ -161,6 +161,7 @@ def valkyrie_test_manifest(obj)
   end
 
   def cdl_readable?(obj)
+    obj = Wayfinder.for(obj).parent.present? ? Wayfinder.for(obj).parent : obj if Wayfinder.for(obj).respond_to?(:parent)
```
escowles, line 164: This is a small thing, but I found the ternary confusing here, then realized it was assinging `obj = obj` if the parent wasn't there, which seems unnecessary. Could this check be
`if Wayfinder.for(obj).respond_to?(:parent) && Wayfinder.for(obj).parent.present?` instead?

tpendragon: Or just `Wayfinder.for(obj).try(:parent)`

kelynch: Sure, this makes sense.  I'll make the adjustment.

app/models/ability.rb
```diff
@@ -256,6 +256,10 @@ def downloadable?(obj)
     obj.decorate.downloadable? || authorized_by_token?(obj) || (!current_user.nil? && (current_user.staff? || current_user.admin?))
   end
 
+  def single_or_multivolume?(obj)
```
tpendragon, line 259: Maybe call this `parent_or_self`

tpendragon (changes requested):
Small thing, then looks good!

app/models/ability.rb
kelynch: Cool, thanks!

tpendragon (approved):

Issue #1656, opened 2018-08-13
Collection manifests are loading really slow


tpendragon:
https://app.datadoghq.com/apm/resource/figgy/rack.request/9d35212e3b635381?start=1533590121621&end=1534194921621&env=none&paused=false&traceID=3968344314415845361&spanID=3560391818623389533&graphType=span_list

It looks like it's running a query for every object in a collection. Maybe to get thumbnails?

tpendragon:
This is preventing creating the dpul collection for Cairo Geniza, making this a priority

tpendragon:
Example trace: https://app.datadoghq.com/apm/trace/7020982465634036058?spanID=5101830313444406528&env=none&sort=time&colorBy=service&spanViewType=metadata&graphType=flamegraph&shouldShowLegend=true

PR #4420, opened 2021-03-01
Don't build child thumbnails for collection manifests.

tpendragon:
We need MVWs to have them for sync to DPUL, but we don't need it for
collection manifests and it adds a ton of extra queries.

Work towards #1656

### DIFF
--- a/app/services/manifest_builder/collection_manifest_service_locator.rb
+++ b/app/services/manifest_builder/collection_manifest_service_locator.rb
@@ -4,8 +4,9 @@ class ManifestBuilder
   # performance.
   class CollectionManifestServiceLocator < ManifestBuilder::ManifestServiceLocator
     class << self
-      ##
       # Builder for a manifest which is a sub-item in a collection.
+      # @note This is the same as in ManifestServiceLocator, but there's no
+      #   thumbnail builder, to improve render time.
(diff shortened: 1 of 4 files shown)


app/services/manifest_builder/collection_manifest_service_locator.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+class ManifestBuilder
+  # Collections have fewer builders (notably the thumbnail builder) to improve
+  # performance.
+  class CollectionManifestServiceLocator < ManifestBuilder::ManifestServiceLocator
+    class << self
+      ##
+      # Builder for a manifest which is a sub-item in a collection.
```
eliotjordan, line 8: Might be useful to explain in a comment that the `thumbnail_builder` is not added / thumbnail urls aren't generated in collection manifests. It's a subtle difference in the code that's easy to miss.

eliotjordan (changes requested):

app/services/manifest_builder/collection_manifest_service_locator.rb
tpendragon: Added a comment. Take a look?

PR #4424, opened 2021-03-03
Improves CSV ingestion and adds some validation

cwulfman:
Needed stat to ingest south asian ephemera on deadline.


app/jobs/validate_ephemera_csv_job.rb
```diff
@@ -0,0 +1,16 @@
...
+class ValidateEphemeraCSVJob < ApplicationJob
+  def perform(project_id, csvfile, basedir)
+    logger.info "Validating csv file #{csvfile}"
+    change_set_persister = ChangeSetPersister.new(
+      metadata_adapter: Valkyrie::MetadataAdapter.find(:indexing_persister),
+      storage_adapter: Valkyrie::StorageAdapter.find(:disk_via_copy)
+    )
+    change_set_persister.queue = queue_name
+    output = nil
+    change_set_persister.buffer_into_index do |buffered_changeset_persister|
```
escowles, line 11: I'm not sure why you're using the `ChangeSetPersister` here — I thought the validation would check before the items were loaded, and I'm not seeing any records being saved in `IngestEphemeraCSV#validate`. Could this be simplified to just calling `validate` and logging the result?

escowles: And if so, then maybe the job isn't required and the Rake task could simply do that?

cwulfman: This was a quick-and-dirty copy and revise. Mea Culpa.

escowles (approved):

PR #4428, opened 2021-03-05
Expand fields included in ephemera metadata export

### LINKED ISSUE #4431: As a figgy ephemera project owner, I would like to download a CSV of all folders in my project, including those that are boxless
I'm looking at the SAE ephemera project and notice that those items that are not in boxes are not showing up when I try to download an ephemera report.  I would like the report to include all folders in the project, in boxes and those that have no box assigned.

escowles:
* Adds support for all of the metadata fields in the ephemera forms
* Include boxless folders in the reports

Fixes #4431

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -137,6 +137,7 @@ Metrics/LineLength:
     - 'spec/helpers/application_helper_spec.rb'
     - 'spec/resources/collections/collections_controller_spec.rb'
     - 'spec/models/oai/figgy/valkyrie_provider_model_spec.rb'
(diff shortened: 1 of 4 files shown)


kelea99:
See also  https://github.com/pulibrary/figgy/issues/4431

cwulfman:
I think you could include the folders this way:
``` ruby
@resources = @ephemera_project.boxes.map(&:folders).flatten + @ephemera_project.folders
```

tpendragon (approved):

Issue #4440, opened 2021-03-16
Ephemera CSV Export doesn't include all fields


escowles:
The recent expansion of fields includes most fields, but omits a few:
* Box#
* sort_title
* ocr_language	
* date_range_start
* date_range_end
* date_range_approximate
* member_of_collection

PR #4445, opened 2021-03-23
adds additional fields to report

cwulfman:
addresses #4440


spec/controllers/reports_controller_spec.rb
```diff
@@ -48,7 +34,11 @@
     end
     it "allows downloading a CSV file" do
       get :ephemera_data, params: { project_id: project.id }, format: "csv"
-      expect(response.body).to eq(data)
+      # expect(response.body).to eq(data)
```
escowles, line 37: I think this is a good change, but it would be good to cover roughly the same amount of detail above to make sure that the whole set of fields is indeed being exported. And removing the commented-out line above would be good too.

escowles (approved):

PR #4451, opened 2021-03-26
Look for the cdl field that bibdata-on-alma will return

hackartisan:
I'm not sure why the item id was previous an int, or whether it will be a
problem for it to be a string here.

fixes pulibrary/bibdata#903

### DIFF
--- a/app/resources/cdl/eligible_item_service.rb
+++ b/app/resources/cdl/eligible_item_service.rb
@@ -4,6 +4,7 @@ module CDL
   class EligibleItemService
     # EligibleItemService will return from Voyager items eligible for Controlled Digital Lending (Cdl)
     # patron_group_charged == "CDL"
+    # with alma the field will be "cdl": true
     class << self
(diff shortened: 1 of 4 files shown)


hackartisan:
Okay I traced this through and I think it makes more sense for it to be a string.

The `CDL::EligibleItemService` is used by a `CDL::ChargeManager` to manage `CDL::ChargedItem`s in `CDL::ResourceChargeList`s. The `item_id` is ultimately stored as a string in the charged item: https://github.com/pulibrary/figgy/blob/0755b93ef23156b493e67281d97b163a925696ba/app/resources/cdl/charged_item.rb#L5 

The number of available items is computed such that I would expect both to be strings: https://github.com/pulibrary/figgy/blob/0755b93ef23156b493e67281d97b163a925696ba/app/resources/cdl/charge_manager.rb#L122

And in fact the item ids used in the charge manager specs are all strings.

tpendragon (approved):

PR #4462, opened 2021-04-06
Add support for Alma ID/Voyager ID Equivalence

### LINKED ISSUE #4406: Migrate bib ids to alma mms ids
When bibdata is migrated to fully run over Alma the IDs in figgy will no longer be valid.

We should add in some logic such that if the record is not found it will convert the id to its alma equivalent and try again. If it is found with the alma id, the record should be updated to hold that ID instead.

Do we also want a task to migrate all voyager IDs to alma IDs?

2 use cases:
- figgy asks orangelight for new data
- orangelight asks figgy for a viewer

This ticket is closed when:
- catalog-alma-qa has viewers for records that don't have an 856. Example record needed -- a CDL item?
- figgy can pull data from catalog-alma-qa

tpendragon:
* Entering a Voyager ID can pull metadata from Alma-powered by bibdata.
* Searching in GraphQL for an Alma ID will return Voyager ID items.

Closes #4406 

Closes https://github.com/pulibrary/bibdata/issues/1083


app/graphql/types/query_type.rb
```diff
@@ -48,12 +48,12 @@ def resources_by_figgy_ids(ids:)
...
 
   def resources_by_bibid(bib_id:)
-    resources = query_service.custom_queries.find_by_property(property: :source_metadata_identifier, value: bib_id).select { |resource| ability.can? :discover, resource }.to_a
+    resources = query_service.custom_queries.find_by_source_metadata_identifier(source_metadata_identifier: bib_id).select { |resource| ability.can? :discover, resource }.to_a
     resources.select { |r| type_defined?(r) }
   end
 
   def resources_by_bibids(bib_ids:)
-    resources = query_service.custom_queries.find_many_by_property(property: :source_metadata_identifier, values: bib_ids)
+    resources = query_service.custom_queries.find_by_source_metadata_identifiers(source_metadata_identifiers: bib_ids)
```
hackartisan, line 56: Why change the naming convention away from `find_many_by...` here? Because of length?

app/models/remote_record.rb
```diff
@@ -16,6 +16,10 @@ def self.bibdata?(source_metadata_identifier)
     PulMetadataServices::Client.bibdata?(source_metadata_identifier)
   end
 
+  def self.alma?(source_metadata_identifier)
+    bibdata?(source_metadata_identifier) && source_metadata_identifier.start_with?("99") && source_metadata_identifier.end_with?("6421")
```
hackartisan, line 20: unfortunately there are voyager ids, e.g. `996421` but I'm not sure there's anything to be done about it.

app/models/remote_record.rb
```diff
@@ -88,7 +92,14 @@ def attributes
   private
 
     def jsonld_request
-      @jsonld_request ||= Faraday.get("https://bibdata.princeton.edu/bibliographic/#{source_metadata_identifier}/jsonld")
+      @jsonld_request ||=
+        begin
+          request = Faraday.get("https://bibdata.princeton.edu/bibliographic/#{source_metadata_identifier}/jsonld")
+          if !self.class.alma?(source_metadata_identifier) && !request.success? && request.status.to_s == "404"
```
hackartisan, line 98: Should we try in any 404 situation? given the chance of voyager IDs that fit the alma pattern?

app/queries/find_by_source_metadata_identifier.rb
```diff
@@ -0,0 +1,34 @@
...
+    if result.blank? && RemoteRecord.alma?(source_metadata_identifier)
+      non_alma_id = source_metadata_identifier.match(/^99([\d]*)3506421/)&.[](1)
+      return result if non_alma_id.nil?
+      return query_service.custom_queries.find_by_property(property: :source_metadata_identifier, value: non_alma_id)
+    end
+    result
+  end
+
+  def find_by_source_metadata_identifiers(source_metadata_identifiers:)
+    alma_ids = source_metadata_identifiers.select do |bib_id|
```
hackartisan, line 26: Same question as previous. Should we just run both versions for every ID in case it's a voyager ID that fits the alma pattern?

hackartisan (changes requested):
this looks great I am a little thinking about edge cases let me know what you think.

app/queries/find_by_source_metadata_identifier.rb
tpendragon: Ah, what a shame. I can do that I guess, but it'll increase the number of queries by a fair number and I have some concern about the number of boolean queries. We can do it by ear though, and it'd only effect performance of thumbnail resolution for the catalog search page I think.

app/graphql/types/query_type.rb
tpendragon: I did this because `find_by_source_metadata_identifier` also returns an array of results. I can add some typedocs.

hackartisan: 👍🏻

hackartisan (approved):

PR #4464, opened 2021-04-07
Filtering resources included in OAI endpoint based on rights statement

### LINKED ISSUE #4463: Some items are available via OAI that probably shouldn't be
I recently reviewed the Figgy OAI endpoint for DPLA to harvest, and noticed there are some items that should probably not be harvested. The biggest category of these items is items we've acquired with an MOU (e.g., the JTS Geniza items, or the Derrida Seminars papers) where we didn't arrange to republish them.

escowles:
Fixes #4463 

This would suppress items from the OAI endpoint based on rights statement. For now, the list of rights statements to suppress is just the No Copyright/Contractual Restrictions which seems like a reasonably good fit for this scenario.

### DIFF
--- a/app/models/oai/figgy/valkyrie_provider_model.rb
+++ b/app/models/oai/figgy/valkyrie_provider_model.rb
@@ -79,7 +79,8 @@ module OAI::Figgy
         marc_only: options.marc?,
         from: options.from,
         until_time: options.until,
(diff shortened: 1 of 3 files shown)


tpendragon (approved):

tpendragon:
This is well executed, thanks!

PR #4473, opened 2021-04-15
Adding '/pdf' action to redirect to best PDF version of a resource

### LINKED ISSUE #4472: Provide an easy way to link to Figgy PDFs by ARK
Currently, to link to either the PDF that Figgy generates, or the original PDF ingested into Figgy, requires going to copy the URL (and navigating the File Manager for the original PDF).

Related to #4469 , it would be good to be able to link to the PDF using the ARK, e.g., by appending "/pdf" to the end of the ARK url: `http://arks.princeton.edu/ark:/88436/xyz12345x/ark`. This would let us link to PDFs using ARKs.

escowles:
Fixes #4472

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -201,11 +201,12 @@ class CatalogController < ApplicationController
 
   def pdf
     _, @document = fetch(params[:solr_document_id])
-    authorize! :show, resource
 
-    pdf_path = helpers.pdf_path(resource)
-    if pdf_path
-      redirect_to "#{request.base_url}#{pdf_path}"
+    source_pdf = Wayfinder.for(resource).source_pdf
+    if source_pdf && can?(:download, source_pdf)
+      redirect_to "#{request.base_url}#{helpers.fileset_download_path(source_pdf)}"
+    elsif helpers.pdf_allowed?(resource) && can?(:pdf, resource)
+      redirect_to "#{request.base_url}#{polymorphic_path([:pdf, resource])}"
     else
       redirect_to solr_document_url(resource), notice: "No PDF available for this item"
     end
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -32,7 +32,7 @@ class CatalogController < ApplicationController
   end
 
   # enforce hydra access controls
-  before_action :set_id, only: :iiif_search
+  before_action :set_id, only: [:iiif_search, :pdf]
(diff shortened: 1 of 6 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -198,4 +198,16 @@ def lookup_manifest
       render json: { message: "No manifest found for #{ark}" }, status: 404
     end
   end
+
+  def pdf
+    _, @document = fetch(params[:solr_document_id])
+    authorize! :show, resource
```
tpendragon, line 204: I'm not sure if this is the right authorization to have. For example, to view the Figgy-generated PDF you only need `:pdf` permission, which isn't true for anonymous users if there's no pdf color-scheme set. For the full res PDF, it's more about the ability to `:download," which is super important for CDL. This is kind of a bummer, because it means you have to vary permission after you figure out which PDF to show them - and if they're not allowed to get the original PDF, can they get the Figgy generated PDF?

app/helpers/application_helper.rb
```diff
@@ -180,6 +180,33 @@ def universal_viewer_path(resource)
     "/viewer#?manifest=#{manifest_url(resource)}"
   end
 
+  def pdf_path(resource)
+    original_pdf_path(resource) || figgy_pdf_path(resource)
+  end
+
+  def original_pdf_path(resource)
+    first_member = resource.decorate.try(:members)&.first
```
tpendragon, line 188: This is a bummer, this'd be really expensive. We'll want to keep our eyes on this and maybe implement a custom query to check members for a mime type. Maybe add a `#source_pdf` to wayfinder?

tpendragon (changes requested):
I'm mostly concerned about the permissions bit.

app/helpers/application_helper.rb
escowles: Yes, I was thinking about how to avoid that. I went ahead and added `#first_member` and `#source_pdf` to the base wayfinder because they could be used by more than half of the wayfinders.

app/controllers/catalog_controller.rb
escowles: 👍🏻  I loaded the source_pdf and used `can?` to check permissions more specifically, so it can fall back on the generated pdf they are not allowed to download the original.

tpendragon (approved):

tpendragon:
@escowles Ha, I don't know why I didn't think of just loading the first member ID. Good work!

Issue #4482, opened 2021-04-26
Address security vulnerabilities


bess:
As of 26 April 2021, Figgy has 13 security issues that dependabot can't handle automatically. 

https://github.com/pulibrary/figgy/security/dependabot

bess:
The remaining issues here are related to the cloud-fixity and I don't want to upgrade anything before it is in CI.

eliotjordan:
Security vulnerabilities addressed as much as possible at this point. What's left:

1. Created [ticket](https://github.com/pulibrary/figgy/issues/6131) for updating Google Pub/Sub and Storage packages
2. Dependencies from Lux - will be replaced at some point
3. OmniAuth issue - tried patch in Pulmap and it's not great.

PR #4490, opened 2021-04-26
Upgrade shine and things that depend on it

bess:
Addresses GHSA-5jjv-x4fq-qjwp
See https://github.com/advisories/GHSA-5jjv-x4fq-qjwp for more info.

Connected to #4482


tpendragon:
We're not sure why this failed, but we'll come back and try again in the future.

PR #4517, opened 2021-05-13
Synchronize with ASpace on Complete

tpendragon:
- [x] Synchronize ZIP Downloads.
- [x] Update existing links (should we do this? We do it now..)
- [x] Delete old links? (This is hard, and we don't do it now - maybe don't.)
- [x] Extract to job and run in background.
- [x] Remove PulfaExporter.
- [x] Fix tests.

Refs https://github.com/pulibrary/princeton_ansible/pull/2324


hackartisan (commented):
Cool!

app/services/dao_updater.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def update!
+    archival_object = aspace_client.find_archival_object_by_component_id(component_id: change_set.source_metadata_identifier)
+
+    # Create digital object.
+    digital_object = create_digital_object(archival_object)
+    # Assign digital object to Archival Object.
+    link_digital_object(archival_object: archival_object, digital_object: digital_object)
+  end
+
```
eliotjordan, line 18: suggestion: Make the methods following `update!` private.

eliotjordan (changes requested):

eliotjordan (approved):
I meant to make that previous review an "approve" with a comment. Overall, this looks great.

PR #4535, opened 2021-06-07
Encode utf-8 in ephemera terms for uri generation

### LINKED ISSUE #4534: [Figgy/production] ActionView::Template::Error: URI must be ascii only "sanJos\u00E9,CostaRica"
## Backtrace

    line 24 of [PROJECT_ROOT]/app/resources/ephemera_terms/ephemera_term_decorator.rb: internal_url
    line 28 of [PROJECT_ROOT]/app/resources/ephemera_terms/ephemera_term_decorator.rb: uri
    line 21 of [PROJECT_ROOT]/app/views/ephemera_terms/index.html.erb: block in _app_views_ephemera_terms_index_html_erb__1312353001481529227_69996287787620

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/79746309)

hackartisan:
fixes #4534

### DIFF
--- a/app/resources/ephemera_terms/ephemera_term_decorator.rb
+++ b/app/resources/ephemera_terms/ephemera_term_decorator.rb
@@ -21,7 +21,7 @@ class EphemeraTermDecorator < Valkyrie::ResourceDecorator
   def internal_url
     return Array.wrap(model.uri).first if vocabulary.blank?
     vocabulary_uri = vocabulary.uri.to_s.end_with?("/") ? vocabulary.uri.to_s : vocabulary.uri.to_s + "/"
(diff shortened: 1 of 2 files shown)


hackartisan:
@tpendragon is this what we want it to do?

tpendragon (changes requested):
Can we do this in `internal_url` instead of `camelized_label` and just always URI.escape so we don't have to check if it's ASCII only?

hackartisan:
Sure. doesn't seem to be much risk of an already-escaped value being in there.

hackartisan:
@tpendragon done

tpendragon (approved):

PR #4569, opened 2021-06-30
Add service to migrate Voyager bibids to Alma

### LINKED ISSUE #4549: Rake task to change voyager IDs to Alma IDs
We'll want to change them at some point rather than just checking for both, and it would be good to have this available in case we  need to run it urgently.

eliotjordan:
Service uses Postgres update queries run in batches. I figured updating millions of records one at a time via ChangeSet Persister would be inefficient. When the updates are complete, Figgy solr is re-indexed to update source meta identifier fields.

Closes #4549 
Closes https://github.com/pulibrary/orangelight/issues/2520


app/services/bibid_updater.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def update
+    logger.info "Updating #{total_count} records"
+    progress_bar
+    (1..iterations).each do
+      run_query(update_query)
+      progress_bar.progress += 1
+    end
+
+    Reindexer.reindex_all
```
tpendragon, line 29: Suggestion: I'd only do reindex_works. FileSets don't have source_metadata_identifiers. In fact, I wonder if there's a way to get the ones that have changed?

app/services/bibid_updater.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def total_count_query
+      <<-SQL
+        SELECT COUNT(*)
+        #{voyager_resources_query}
+      SQL
+    end
+
+    def update_query
```
tpendragon, line 53: issue: A concern I have about this is it won't update the metadata in preservation storage or re-fetch metadata from the alma records to make sure it's okay.

app/services/bibid_updater.rb
```diff
@@ -0,0 +1,78 @@
...
+      SQL
+    end
+
+    def voyager_resources_query
+      <<-SQL
+        -- Find all resources that have bibids
+        FROM orm_resources a,
+        jsonb_array_elements_text(a.metadata->'source_metadata_identifier') AS b(identifier)
+        -- Skip records that have an Alma identifier
+        WHERE NOT identifier ~ '99.*3506421'
```
tpendragon, line 75: issue: source_metadata_identifier can have EAD IDs. This'd make a bunch of 99C0652_c0113506421

tpendragon (changes requested):
There are a few individual comments in the PR review, but in general I have some concerns about this strategy. Except for reindexing it's definitely a lot faster, but it bypasses all the logic that happens when updating resources. No messages would be fired, no syncing would happen, no preservation would occur.

Also, the fact that finding aids materials use the same field is an issue (I really wish we had stored a source property somewhere.)

It'd be a lot slower, but I wonder if doing something like:

```ruby
resources = query_service.custom_queries.find_by_property(property: :source_metadata_identifier, value: [], lazy: true).select{|resource| RemoteRecord.bibdata?(resource.source_metadata_identifier.first)}
```

and then going through the CSP in a background job or tmux'd process would be okay. 

I ran the query and selection code on Prod and it takes about 5 minutes to iterate over all of them and make them Valkyrie resources (there's 71,482). I imagine it'd take a really long time to do all the CSP stuff after all, especially preservation. Maybe doing the query and bypassing the change set persister would be okay? It would index the appropriate ones too, and we'd just have to be okay with no events being fired and preservation metadata being out of date...

app/services/bibid_updater.rb
eliotjordan: Yep. Forgot about EAD IDs. :(

eliotjordan:
@tpendragon You make good points here. I started down this road, then thought it might not be the best, but decided to give it a whirl anyway. Forgot about the syncing and preservation angle. Was really hoping we could do this in a nice clean SQL query :) However, the CSP does a lot for us and I think we'll just have to live with the long execution time. I'll rework this.

tpendragon:
@eliotjordan It's good work - I think the fact that we can't do it in a nice SQL query (and that we want to) points to some issues or potential improvements we could make in the future. Maybe we can chat about this tomorrow? Updating preservation for all those items is definitely going to take a -long- time.

eliotjordan:
@tpendragon Implemented a simple solution that uses the CSP. It is going to take a long time to complete running in a tmux session, but it can be restarted easily.

app/services/bibid_updater.rb
```diff
@@ -0,0 +1,55 @@
...
+    def resources
+      @resources ||= query_service.custom_queries.find_by_property(property: :source_metadata_identifier, value: [], lazy: true).select do |resource|
+        id = resource.source_metadata_identifier.first
+        next if id =~ /99.*3506421/
+        RemoteRecord.bibdata?(id)
+      end
+    end
+
+    def total_count
+      @total_count ||= resources.count
```
tpendragon, line 49: Suggestion: I think calling this is going to load all 12k resources into memory before the progress bar even starts. Do we have to provide a total for the progress bar to work? I'm tempted to even just give it a roughly-correct number.

tpendragon (changes requested):
Just a small comment. I guess if it only takes a few minutes it might be fine to load all these into memory, what do you think?

app/services/bibid_updater.rb
eliotjordan: Good idea. I'll put in a place holder number.

eliotjordan:
@tpendragon Fixed.

tpendragon (dismissed):

app/services/bibid_updater.rb
```diff
@@ -0,0 +1,54 @@
...
+      change_set_persister.save(change_set: change_set)
+      progress_bar.progress += 1
+    end
+  end
+
+  private
+
+    def change_set_persister
+      @change_set_perisister ||= ChangeSetPersister.new(metadata_adapter: metadata_adapter,
+                                                        storage_adapter: Valkyrie.config.storage_adapter)
```
christinach, line 23: @eliotjordan should this be @change_set_persister?

christinach (changes requested):

app/services/bibid_updater.rb
eliotjordan: yep.

christinach (approved):

tpendragon (approved):

Issue #4174, opened 2020-08-10
Analyze Orphaned FileSets to figure out why they're orphaned.


tpendragon:

tpendragon:
It seems like several of these are because of https://github.com/samvera/valkyrie/issues/837

eliotjordan:
I'm coming across a large number of malformed orphaned file sets in a honey badge exception:

https://app.honeybadger.io/projects/53391/faults/63228597/01F39THEQWK7YYNQB433WC388R#notice-summary

Too many to clean up manually. Perhaps we should create a task to periodically delete them?

Issue #2188, opened 2018-11-16
Derivatives are being lost on bulk ingest


tpendragon:
Reported by @roelmoon 

https://app.datadoghq.com/logs?cols=%5B%22core_host%22%2C%22core_service%22%5D&event&from_ts=1541791469524&index=main&live=true&query=40b6f4df-4c14-4d7d-a7b3-4a2f937a7749&stream_sort=asc&to_ts=1542396269524 has a log of what's going on. A summary is below:

So what's going on: CreateDerivativesJob is running derivatives. The PROCESS of running derivatives seems to be running a CheckFixityJob. The CreateDerivativesJob keeps working, holding onto a row lock for the FileSet until it's done. Meanwhile, CheckFixityJob runs. It pulls in the file, runs fixity, and fires an update - which WAITS (nicely! Thanks postgres!) for the lock in CreateDerivativesJob to be released. That finally happens ("finally", it's like, a second time-span) and that update goes through, but that update doesn't have information about the derivative, because it pulled it before CreateDerivativesJob committed its transaction. Tl;dr, optimistic locking retries would have fixed this.

So we can fix this by turning on optimistic locking or stopping CreateDerivativesJob from running CheckFixityJob twice.

tpendragon:
Turns out there are two problems:

`DefaultDerivativeService` runs in two transactions and #2190

tpendragon:
We should be able to close this after #2192 and #2191 are merged.

PR #4592, opened 2021-07-28
Test Valkyrie Upgrade

tpendragon:
Work towards #4174 

This is testing a branch of Valkyrie. Merge and deploy when we have the energy to watch for potential performance degradation or surprise issues.


app/services/preserver/blind_importer.rb
```diff
@@ -33,6 +33,7 @@ def import!
     end
     # Get rid of non-preserved members.
     source_change_set.try(:member_ids=, member_ids)
+    source_change_set.resource.new_record = true
```
hackartisan, line 36: ?

app/services/preserver/importer.rb
```diff
@@ -85,6 +85,7 @@ def build_file_set(file_identifier)
           build_optimistic_lock_token(lock_json)
         end
         file_set.optimistic_lock_token = optimistic_lock_token
+        file_set.new_record = true
```
hackartisan, line 88: Is this is a new record why do we have to explicitly set this?

spec/change_set_persisters/change_set_persister_spec.rb
```diff
@@ -644,8 +644,8 @@
         expect(members.first).to be_kind_of FileSet
 
         expect(members.first.date_of_digitization).not_to be_empty
-        expect(members.first.date_of_digitization.first).to be_a Time
-        expect(members.first.date_of_digitization.first).to eq DateTime.iso8601("2009-03-30T19:49:13.000Z").to_time.utc
+        expect(members.first.date_of_digitization.first).to be_a DateTime
+        expect(members.first.date_of_digitization.first).to eq DateTime.iso8601("2009-03-30T19:49:13.000Z").utc
```
hackartisan, line 648: We won't need any data migration or anything?

spec/resources/numismatics/coins/numismatics/coins_controller_spec.rb
```diff
@@ -75,7 +75,7 @@
     end
     let(:resource) { FactoryBot.create_for_repository(:coin) }
     it "retrieves an array of facet values to for use in populating select boxes" do
-      metadata_adapter = Valkyrie::MetadataAdapter.find(:index_solr)
+      metadata_adapter = Valkyrie::MetadataAdapter.find(:indexing_persister)
```
hackartisan, line 78: does this mean we haven't been persisting changes to coins?

spec/derivative_services/default_derivative_service_spec.rb
```diff
@@ -60,8 +59,6 @@
     expect(derivative).to be_present
     derivative_file = Valkyrie::StorageAdapter.find_by(id: derivative.file_identifiers.first)
     expect(derivative_file.read).not_to be_blank
-    # Ensure only one transaction is used
-    expect(adapter.metadata_adapter.connection).to have_received(:transaction).exactly(1).times
```
hackartisan, line 64: Why did we care about the number of transactions before? And why don't we care anymore?

spec/queries/find_invalid_thumbnail_resources_spec.rb
```diff
@@ -13,7 +13,7 @@
   end
   let(:file) { fixture_file_upload("files/color-landscape.tif", "image/tiff") }
   let(:resource2) { FactoryBot.create_for_repository(:scanned_resource, files: [file]) }
-  let(:change_set_persister) { ChangeSetPersister.new(metadata_adapter: metadata_adapter, storage_adapter: Valkyrie.config.storage_adapter) }
+  let(:change_set_persister) { ScannedResourcesController.change_set_persister }
```
hackartisan, line 16: `ChangeSetPersister.default`?

spec/queries/find_missing_thumbnail_resources_spec.rb
```diff
@@ -13,7 +13,7 @@
   end
   let(:file) { fixture_file_upload("files/color-landscape.tif", "image/tiff") }
   let(:resource2) { FactoryBot.create_for_repository(:scanned_resource, files: [file]) }
-  let(:change_set_persister) { ChangeSetPersister.new(metadata_adapter: metadata_adapter, storage_adapter: Valkyrie.config.storage_adapter) }
+  let(:change_set_persister) { ScannedResourcesController.change_set_persister }
```
hackartisan, line 16: `ChangeSetPersister.default`?

hackartisan (changes requested):
I looked at the upstream changes and feel generally good about this. But I have some questions inline; if it's easier we could talk through them.

tpendragon:
@hackartisan I'm happy to talk through them! I'll leave some responses inline until we can get together.

app/services/preserver/blind_importer.rb
tpendragon: https://github.com/samvera/valkyrie/pull/867/commits/d81d2ae2562a86a5767ed945ecfa59ffe3ebe918#diff-a006f149ad43c693a7064a57d903dacc5dcc4a4609d21ed2f54340c9fbd3829eL34-L35

The following is a little fuzzy:

Valkyrie used to have a hack in it to pretend records were new so it wouldn't perform ID or optimistic locking checks. Now there's external_resource: true in CompositePersister so it wasn't necessary. Unfortunately we were relying on that for importers, which create new records by duplicating the previous object.

It might be we can use external_record: true here, but I don't think ChangeSetPersister provides access to that quite yet.

spec/change_set_persisters/change_set_persister_spec.rb
tpendragon: We shouldn't. This was just a parsing bug in Valkyrie that's gotten fixed, so it'll parse out as a DateTime now and render back the same way.

spec/resources/numismatics/coins/numismatics/coins_controller_spec.rb
tpendragon: The actual controller persists to both places, this test just tried to be more efficient by only persisting to Solr (which we can't query now.)

spec/derivative_services/default_derivative_service_spec.rb
tpendragon: Thank you. The research here was fruitful - pushing an update that fixes this test to check for the buffer rather than the transaction, and links to https://github.com/pulibrary/figgy/issues/2188 (which is why this feature was added.)

This started to fail because failable transactions are now used to check if an invalid ID was given to postgres. 

https://github.com/samvera-labs/valkyrie-sequel/pull/41

app/services/preserver/blind_importer.rb
tpendragon: I think I can get rid of this, running tests locally.

hackartisan (changes requested):
These changes look good, thanks!

there are a few more places where change set persisters are getting pulled out of controllers. They're all numismatics-y, but I still wonder if it sets a better example in the coded to use `ChangeSetPersister.default`.

tpendragon:
@hackartisan I think I got 'em.

hackartisan (approved):
👍🏻 so is the plan to merge this or deploy from this branch as a trial?

tpendragon:
I think we should do a "squash and merge" so we have commit we can revert, and then deploy from main.

PR #4619, opened 2021-08-10
3897 select2

### LINKED ISSUE #3897: Stop using hidden fields in select2 form fields
It's deprecated. These can be removed and their `name` element moved to the select field directly.

See https://select2.org/upgrading/migrating-from-35#no-more-hidden-input-tags; https://github.com/pulibrary/figgy/blob/9cf3ed854fe1f9804e23c3ac407ac32f69003640/app/views/numismatics/issues/edit_fields/_shape.html.erb

Tests look for these hidden fields directly; should maybe be adjusted to check the selected value.

XanderGardner:
closes #3897. Removed deprecated select2 elements and rewrote tests to stop searching for hidden elements.

Also note: in issues_feature_spec.rb, I got an error from adding `binding.pry` and calling `page.select`:
```
pry(#<RSpec::ExampleGroups::NumismaticsIssues::FormEditing::WhenIssuesHaveBeenSaved>)> page.select("edit_numismatics_issue")
Errno::EMFILE: Failed to open TCP connection to 127.0.0.1:9517 (Too many open files - socket(2) for "127.0.0.1" port 9517)
```
this can be fixed by adding in the spec_helper:
`WebMock.disable_net_connect!(net_http_connect_on_start: true).`
although it causes tests to fails.


hackartisan:
Hm, the tests are failing now in CI, seems likely due to the WebMock change. Shall we try removing that commit?

tpendragon (changes requested):
This looks good, but when I create an Issue in figgy and select an "object type" or any other select value it doesn't save.

tpendragon (approved):

PR #4628, opened 2021-08-13
BibID Updater Fixes

tpendragon:
I found that the cursor used in the find_by_property query forces a
transaction, meaning any changes weren't reflected in the edit forms of
the UI, and resulted in Solr being in a different state than postgres.

There's plenty of memory to load it in, and it only takes a minute or
so.

### DIFF
--- a/app/services/bibid_updater.rb
+++ b/app/services/bibid_updater.rb
@@ -21,8 +21,7 @@ class BibidUpdater
   private
 
     def change_set_persister
-      @change_set_persister ||= ChangeSetPersister.new(metadata_adapter: Valkyrie.config.metadata_adapter,
-                                                       storage_adapter: Valkyrie.config.storage_adapter)
+      @change_set_persister ||= ScannedResourcesController.change_set_persister
     end
 
     def progress_bar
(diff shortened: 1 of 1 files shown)


app/services/bibid_updater.rb
```diff
@@ -21,29 +21,32 @@ def update
   private
 
     def change_set_persister
-      @change_set_persister ||= ChangeSetPersister.new(metadata_adapter: Valkyrie.config.metadata_adapter,
-                                                       storage_adapter: Valkyrie.config.storage_adapter)
+      @change_set_persister ||= ScannedResourcesController.change_set_persister
```
hackartisan, line 24: You like this better? I think it's fine in a temporary service like this but I don't know that I'd want it sprinkled about.

hackartisan (approved):

app/services/bibid_updater.rb
tpendragon: The previous one was actually wrong. If we started doing this for real I think `ChangeSetPersister.default` would be good.

hackartisan: 👍🏻 something like that would be good.

tpendragon: Created https://github.com/pulibrary/figgy/issues/4629

PR #4633, opened 2021-08-16
Add default method to ChangeSetPersister

### LINKED ISSUE #4629: Add class method `ChangeSetPersister.default` and use it.
Extract

https://github.com/pulibrary/figgy/blob/07edc703bf18acb845c9de7b99aa0751729efc33/app/resources/scanned_resources/scanned_resources_controller.rb#L4-L7

into a class method on ChangeSetPersister. Then search for any cases that are exactly like above and call `ChangeSetPersister.default` instead.

thanyabegum:
Closes #4629

### DIFF
--- a/app/change_set_persisters/change_set_persister.rb
+++ b/app/change_set_persisters/change_set_persister.rb
@@ -9,6 +9,13 @@ class ChangeSetPersister
               handlers: registered_handlers)
   end
 
+  def self.default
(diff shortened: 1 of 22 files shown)


tpendragon (approved):

hackartisan:
Wow I didn't realize how often the change set persister was pulled from that controller. This is so much better!!

PR #4635, opened 2021-08-17
Remove LetterChangeSet and tests

### LINKED ISSUE #3546: delete letter change set
We ended up not using it for the Eliot / Hale collection.

thanyabegum:
Closes #3546


app/change_sets/change_set.rb
```diff
@@ -135,19 +129,22 @@ def fix_multivalued_keys
...
-        collection.delete_at(index)
-        return skip!
-      else
-        item
-      end
-    elsif delete_fragment?(fragment)
+
+    if delete_fragment?(fragment)
+      # If the object should be deleted, then delete it if it exists
+      collection.delete_at(index) if item
```
tpendragon, line 141: Suggestion: Since you have the tests set up now, what is `index` when the item doesn't exist? Is it nil? If so, maybe we can key off that instead of finding the item at all?

thanyabegum: When the item doesn't exist, ``index`` isn't nil but the index where we should add the item. In a lot of our test cases, that index is zero.

tpendragon (approved):

PR #4638, opened 2021-08-19
Avoid updating Alma native records.

### LINKED ISSUE #4637: Figgy records which were in only Alma got "converted" to Alma IDs
Example:
99125378001906421 became 99991253780019064213506421

This is because the migration tool looked for "3506421" which is only the case for imported records. We'll have to go back and manually fix this.

tpendragon:
Closes #4637

### DIFF
--- a/app/services/bibid_updater.rb
+++ b/app/services/bibid_updater.rb
@@ -40,6 +40,7 @@ class BibidUpdater
           query_service.custom_queries.find_by_property(property: :source_metadata_identifier, value: [], lazy: true).select do |resource|
             id = resource.source_metadata_identifier.first
             next if id =~ /99.*3506421/
+            next if transform_id(id).length > 18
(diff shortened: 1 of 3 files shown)


app/services/bibid_updater.rb
```diff
@@ -40,6 +40,7 @@ def resources
           query_service.custom_queries.find_by_property(property: :source_metadata_identifier, value: [], lazy: true).select do |resource|
             id = resource.source_metadata_identifier.first
             next if id =~ /99.*3506421/
+            next if transform_id(id).length > 18
             RemoteRecord.bibdata?(id)
```
christinach, line 44: why is this > 18?

tpendragon: The biggest ID that started with "9999" was 16 characters.

christinach (approved):

Issue #4639, opened 2021-08-23
As Ephemera Project owners, we would like a CSV download of all LAE Subjects controlled vocabulary, so that we can identify as a group core and area specific subjects in the control.


kelea99:
The cleanup for all  anomalies in the ephemera controlled vocab for LAE subjects is complete as of friday afternoon . Could you provide me a CSV dump of all the LAE Subjects? People are meeting on Wednesday and ideally I’d like to get it to folks asap so they can work on what is the core subject vocab and what is specific to their area (LAE/SAE/Slavic/etc).

hackartisan:
This is interesting. In the past we've synchronized a local fixture from production data. But it doesn't look like we kept the synchronization code: https://github.com/pulibrary/figgy/pull/3333

hackartisan:
Got an error running the task: 

```
$ OUTPUT=tmp/lae_subjects.csv bundle exec rake report:lae_subjects
rake aborted!
JSON::ParserError: 809: unexpected token at 'http://viaf.org/viaf/133607895'
/opt/figgy/shared/bundle/ruby/2.6.0/gems/json-2.5.1/lib/json/common.rb:216:in `parse'
/opt/figgy/shared/bundle/ruby/2.6.0/gems/json-2.5.1/lib/json/common.rb:216:in `parse'
/opt/figgy/releases/20210823205334/lib/tasks/report.rake:22:in `block (4 levels) in <main>'
/opt/figgy/releases/20210823205334/lib/tasks/report.rake:18:in `each'
/opt/figgy/releases/20210823205334/lib/tasks/report.rake:18:in `block (3 levels) in <main>'
/opt/figgy/releases/20210823205334/lib/tasks/report.rake:16:in `block (2 levels) in <main>'
/opt/figgy/shared/bundle/ruby/2.6.0/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
Tasks: TOP => report:lae_subjects
(See full trace by running task with --trace)
```

tpendragon:
I think I have this nailed down, turns out the ephemera vocab ingester's been broken ever since we made single valued fields a thing.

kelea99:
Reopening this. Requesting a new report exactly like this for stakeholder use and review! Ellen Ambrosone Request.

kelea99:
re-closing ticket. created  #6086

PR #4645, opened 2021-08-23
Fix ephemera terms report and vocab ingest service.

tpendragon:
Refs #4639


hackartisan (dismissed):
👍🏻 did you also update any of the data?

bess (approved):

PR #4654, opened 2021-09-07
Update directions for a fresh install.

tpendragon:
I just set up Figgy on a fresh MacOS box with asdf, so I thought
scripting the commands together would make it easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,9 @@ A digital repository application in use at Princeton University Library for stor
 
 ## Language Dependencies
 
-Figgy provides a `.tool-versions` file -- consule this file for required languages and their current versions. `asdf` users should ensure all plugins listed there are installed:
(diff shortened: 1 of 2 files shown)


hackartisan (commented):
This is cool!

I think it would be good to adjust the dependencies sections to make it clear what steps can be replaced by running this script. I can take a pass at it and add a commit for consideration.

hackartisan:
@tpendragon what do you think of my changes?

Note that I removed ubuntu instructions since they were incomplete and I don't think we have anyone using them. Also I figured we have them in the ansible codebase.

tpendragon:
I like it!

tpendragon:
I approve of your commit, but I can't approve this PR @hackartisan if you could.

hackartisan (approved):

PR #4660, opened 2021-09-13
Update Structure.sql to use lando generated info.

### LINKED ISSUE #4661: Structure.sql changing on every rake db:migrate
I think this is because a different version of psql checked in that file. Now that we use lando it shouldn't change as much, I hope.

tpendragon:
Closes #4661


hackartisan (approved):
yes thank you

PR #4674, opened 2021-09-17
Delete MediaResource

thanyabegum:


spec/graphql/types/query_type_spec.rb
```diff
@@ -81,19 +81,6 @@
         expect(type.resources_by_bibid(bib_id: "7214786")).to eq []
       end
     end
-
-    context "when the resource does not have a defined graphql type" do
```
hackartisan, line 85: We couldn't find any resources without defined graphql type anymore.

tpendragon (approved):
This is amazing - good work @thanyabegum !

PR #4677, opened 2021-09-20
Fixes TwoUp labeling in Order Manager

### LINKED ISSUE #4476: Automatic pagination for two up images not funtioning in figgy order manager
I’m making a new ticket for this, but I’m positive we already have one. Anna Meerson also mentioned the need for automated  two up pagination for order manager (for the microfilm)

sdellis:
fixes #4476


sdellis:
Note: The page-label-generator ES6 module never imported into the OrderManager correctly via npm. It is referenced from the `/app/javascript/utils` directory for now. This fix simply updates that code to the latest version, removes the package from package.json (avoiding confusion), and deletes an unnecessary file (lablr.js)

hackartisan (approved):
Works great

PR #4693, opened 2021-10-05
Ensure a resource only has one parent when added to a new parent via the show page form

### LINKED ISSUE #4666: "Attach an existing resource by ID" on record page does not remove the attached resource from its previous parent.
This results in a resource being in two parents, which should never be the case.

@eliotjordan Do you have any memory of this being on purpose? I know this feature was originally for MapSets.

hackartisan:
closes #4666

### DIFF
--- a/app/javascript/test/assets/member_resources_table.spec.js
+++ b/app/javascript/test/assets/member_resources_table.spec.js
@@ -1,0 +1,48 @@
+import MemberResourcesTable from '../../../assets/javascripts/relationships/member_resources_table'
+const jQ = jest.requireActual("jquery");
+global.$ = jQ;
+
+describe('MemberResourcesTable', () => {
+
+  describe('bindAddButton', () => {
+    it('instantiates', () => {
+      const parentId = "7709d87f-0a21-4f8d-b17f-78d5caf4803a"
+      const newMemberId = "317fa019-3c12-4423-bfe8-b913111559ed"
+      const initialHTML = `
+        <div class="panel-body member-resources">
+          <table id="members-scanned-resources"
+          data-update-url="http://localhost:3000/concern/scanned_resources/${parentId}"
+          data-query-url="/catalog/${parentId}"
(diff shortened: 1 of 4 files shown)


app/javascript/test/assets/member_resources_table.spec.js
```diff
@@ -0,0 +1,45 @@
...
+          <input type="text" name="scanned_resource[member_ids]"
+          id="scanned_resource_member_ids" class="related_resource_ids">
+          <button name="button" type="submit" class="btn-add-row">Attach</button>
+        </div>`
+      document.body.innerHTML = initialHTML
+
+      const form = null
+      const element = document.getElementsByClassName('member-resources')[0];
+      const table = new MemberResourcesTable(element, form)
+      const ajax_spy = jest.spyOn(table, 'callAjax').mockImplementation(() => true)
```
tpendragon, line 28: Suggestion: We should avoid mocking the object under test. I wonder if we can mock `$.ajax` instead? I know Jest mocks are a major pain.

tpendragon (changes requested):
Couple of bugs (which I think are bugs in the change set persister thing that handles append_id)

* If you add a child resource, then hit detach, that member gets added a second time (I think this is the detach button being broken in combo with append_id not deleting all members or checking if it's already a member of that parent)
* If a parent has the same child twice, and you add that child to a different parent, only one child instance gets removed from the first parent.

I tested this by:

1. Create Test Parent 1 & Test Parent 2
2. Create Child
3. Add Child to Test Parent 1
4. Click "Detach" (duplicate made)
5. Add Child to Test Parent 2 (One instance removed from Test Parent 1, successfully added to Test Parent 2)

hackartisan:
@tpendragon we've fixed those things. Take another look?

The detach button was definitely broken. Created https://github.com/pulibrary/figgy/issues/4701 and https://github.com/pulibrary/figgy/issues/4702 as follow-up issues.

tpendragon (dismissed):

PR #4699, opened 2021-10-06
Use omniauth-rails-csrf_protection gem to resolve CVE-2015-9284

### LINKED ISSUE #3132: CVE-2015-9284
How are we feeling about CVE-2015-9284? I've been watching the omni-auth repo comments, but no fix is forthcoming.

There is a gem that mitigates this attack:
https://github.com/cookpad/omniauth-rails_csrf_protection

Implementation:
https://github.com/brave-intl/publishers/pull/1942

Is this worth trying, or should we wait for an official fix?

XanderGardner:
closes #3132


XanderGardner:
@eliotjordan might want to look into this. The omniauth-rails-csrf_protection gem works for `link_to` methods, but not for `redirect_to` as it can only function as a get method. I've found only two options for implementing this gem: one is two allow `redirect_to` to operate without the protection from the gem, and the other is to add an additional page to the website (effectively replace each `redirect_to` with a `link_to).

PR #4733, opened 2021-10-22
Add navPlace to ScannedMaps

### LINKED ISSUE #4732: Add navPlace extension to ScannedMap manifests
https://iiif.io/api/extension/navplace/

1. Generate Presentation v3 manifests and add new route for them
1. Generate navPlace extensions in v3 manifest

eliotjordan:
Closes #4732

### DIFF
--- a/app/services/manifest_builder_v3.rb
+++ b/app/services/manifest_builder_v3.rb
@@ -154,8 +154,8 @@ class ManifestBuilderV3
       @members ||= begin
         manifestable_members.map do |member|
           decorator = member.decorate
-          if decorator.respond_to?(:scanned_map_members) && decorator.scanned_map_members.empty?
-            member.decorate.geo_members.first
+          if decorator.respond_to?(:decorated_scanned_maps) && decorator.decorated_scanned_maps.empty?
+            decorator.geo_members.first
           else
             member
           end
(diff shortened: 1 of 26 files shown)


app/services/manifest_builder_v3.rb
```diff
@@ -0,0 +1,393 @@
...
+    def source_metadata_identifier
+      resource.try(:source_metadata_identifier)
+    end
+
+    ##
+    # @param [Resource] resource the Resource being modeled as the root
+    def initialize(resource, auth_token = nil, current_ability = nil)
+      @resource = resource
+      @auth_token = auth_token
+      @current_ability = current_ability
```
tpendragon, line 39: Question: This seems to be the case now too, but even if I don't have permission to a child resource it'll get added to the manifest?

config/routes.rb
```diff
@@ -74,6 +74,7 @@
   end
 
   get "/downloads/:resource_id/file/:id", to: "downloads#show", as: :download
+  get "/manifests/:id/v3", to: "manifests#v3", as: :manifest_v3, defaults: { format: :json }
```
tpendragon, line 77: Comment: We'll have to add this route to nginx_extra_config in princeton_ansible for CORS.

tpendragon (approved):
Just one question.

app/services/manifest_builder_v3.rb
eliotjordan: Yes, it's the case now. I can't imagine, though, a case where someone would have permission on the parent and not the child.

config/routes.rb
eliotjordan: https://github.com/pulibrary/princeton_ansible/issues/2551

Issue #4728, opened 2021-10-21
find_number/issue_number/coin_number/accession_number should be Valkyrie::Types::Coercible::Integer


tpendragon:
Right now they're persisting as strings because they're coming out of a form.

See #4726 for background info.

This could include a migration to convert those string values to integers.

hackartisan:
The merged PR had to be reverted; I pushed the 4728-coin-accession branch again so we can work more on this.

Error observed on staging after deploy prevented the migration from running. A replication of the error: 
```
irb(main):013:0> r = query_service.find_all_of_model(model: Numismatics::Coin).last
Traceback (most recent call last):
        6: from (irb):13
        5: from (irb):13:in `rescue in irb_binding'
        4: from app/adapters/instrumented_adapter.rb:141:in `find_all_of_model'
        3: from app/adapters/instrumented_adapter.rb:44:in `trace'
        2: from app/adapters/instrumented_adapter.rb:48:in `block in trace'
        1: from app/adapters/instrumented_adapter.rb:143:in `block in find_all_of_model'
Dry::Struct::Error ([Numismatics::Coin.new] [] (Array) has invalid type for :find_number violates constraints (can't convert Array into Integer failed))
```

hackartisan:
I took a look at the data and find number definitely can have alpha characters

hackartisan:
reopening until we've run the migration

hackartisan:
migration is finished

PR #4759, opened 2021-10-29
Coerce numismatics numbers to integers

hackartisan:
Make them optional because in at least one case there was a nil value
that raised an error.

refs #4728

### DIFF
--- a/app/resources/numismatics/accessions/numismatics/accession.rb
+++ b/app/resources/numismatics/accessions/numismatics/accession.rb
@@ -11,7 +11,7 @@ module Numismatics
     attribute :numismatic_citation, Valkyrie::Types::Array.of(Numismatics::Citation).meta(ordered: true)
 
     # descriptive metadata
-    attribute :accession_number, Valkyrie::Types::Integer
+    attribute :accession_number, Valkyrie::Types::Coercible::Integer.optional
     attribute :account
     attribute :cost
     attribute :date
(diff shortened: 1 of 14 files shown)


app/resources/numismatics/accessions/numismatics/accession.rb
```diff
@@ -11,7 +11,7 @@ class Accession < Resource
     attribute :numismatic_citation, Valkyrie::Types::Array.of(Numismatics::Citation).meta(ordered: true)
 
     # descriptive metadata
-    attribute :accession_number, Valkyrie::Types::Integer
+    attribute :accession_number, Valkyrie::Types::Coercible::Integer.optional
```
tpendragon, line 14: Question: Should we be using BetterParamsInteger for this and the other instances where we've replaced it with `Coercible::Integer.optional`?

tpendragon (changes requested):

hackartisan:
@tpendragon The other 3 properties were all using the AutoIncrementValidator, so they don't need the BetterParamsInteger. I added a comment referencing our discussion yesterday, about the possibility of using a validator for the coin_number. But I don't think we need any further code changes. Please take a look when you get a chance! Assuming it's ready, when should we run the migration?

tpendragon (approved):

tpendragon:
@hackartisan Whenever should be fine. Maybe kick it off some time today if we get this deployed?

PR #4767, opened 2021-11-03
Add "Rarely Used Terms" feature to Ephemera Fields

### LINKED ISSUE #4647: Add a way to select terms which should be grouped as "Rarely Used" for a given project.
In each project there should be a drop-down for each Controlled Vocabulary to select "rarely used" terms. Those should group in their own category at the bottom of the EphemeraFolder ingest.

Newly added terms to a vocabulary should not be marked as "rarely used"

## Implementation

I think we can implement this the same way we implement "favorite terms", which is just store a list of ephemera term IDs in the EphemeraField. See https://github.com/pulibrary/figgy/pull/1961/files for how we implemented that.

tpendragon:
Closes #4647 

Screen Shot 2021-11-03 at 11 51 15 AM

### DIFF
--- a/app/resources/ephemera_fields/ephemera_field_decorator.rb
+++ b/app/resources/ephemera_fields/ephemera_field_decorator.rb
@@ -25,6 +25,42 @@ class EphemeraFieldDecorator < Valkyrie::ResourceDecorator
     wayfinder.decorated_rarely_used_terms
   end
 
+  def sorted_terms
+    return subject_sorted_categories if attribute_name == "subject"
+    non_favorite_terms = vocabulary.terms.select { |x| !favorite_term_ids.include?(x.id) && !rarely_used_term_ids.include?(x.id) }
+    favorite_terms + non_favorite_terms + rarely_used_terms
+  end
+
+  def subject_sorted_categories
+    [favorite_category] + filtered_categories + [rarely_used_category]
+  end
+
+  def filtered_categories
+    vocabulary.categories.map do |category|
+      terms = category.terms.select do |term|
+        !rarely_used_term_ids.include?(term.id)
+      end
+      OpenStruct.new(
+        label: category.label,
+        terms: terms
+      )
+    end
+  end
+
+  def favorite_category
+    OpenStruct.new(
+      label: "Favorites",
+      terms: favorite_terms
+    )
+  end
+
+  def rarely_used_category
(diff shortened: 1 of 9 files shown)


tpendragon:
Converting to draft to do some refactoring of this load_fields business.

tpendragon:
Ready for review - I think there's more refactoring that could be done in EphemeraFieldDecorator, but I'm unsure of which pattern would make the code actually more readable.

app/resources/ephemera_fields/ephemera_field_decorator.rb
```diff
@@ -21,6 +21,50 @@ def favorite_terms
...
+  # terms at the beginning.
+  def sorted_terms_or_categories
+    return sorted_categories if attribute_name == "subject"
+    non_favorite_terms = vocabulary.terms.select { |x| !favorite_term_ids.include?(x.id) && !rarely_used_term_ids.include?(x.id) }
+    favorite_terms + non_favorite_terms + rarely_used_terms
+  end
+
+  # Returns categories where favorite terms appear at the top and rarely used
+  # terms are removed from categories and placed at the bottom.
+  def sorted_categories
```
hackartisan, line 38: Should we make these private?

app/resources/ephemera_fields/ephemera_field_decorator.rb
```diff
@@ -21,6 +21,50 @@ def favorite_terms
...
+
+  # Returns categories or terms with rarely used terms at the end and favorite
+  # terms at the beginning.
+  def sorted_terms_or_categories
+    return sorted_categories if attribute_name == "subject"
+    non_favorite_terms = vocabulary.terms.select { |x| !favorite_term_ids.include?(x.id) && !rarely_used_term_ids.include?(x.id) }
+    favorite_terms + non_favorite_terms + rarely_used_terms
+  end
+
+  # Returns categories where favorite terms appear at the top and rarely used
```
hackartisan, line 36: So favorite terms appear in both places. It might help to add this to the comment.

hackartisan (changes requested):
I think it looks great. Seems clear and I don't think we need more refactoring. I made a couple of very small comments.

tpendragon:
@hackartisan Thank you! Should be ready for re-review.

hackartisan (approved):

Issue #4755, opened 2021-10-28
CDL Early Return is throwing errors on a regular basis.


tpendragon:
## Backtrace

    line 75 of [PROJECT_ROOT]/app/adapters/instrumented_adapter.rb: block in save
    line 48 of [PROJECT_ROOT]/app/adapters/instrumented_adapter.rb: block in trace
    line 44 of [PROJECT_ROOT]/app/adapters/instrumented_adapter.rb: trace
    
## Params

```ruby
{
  "id" => "4494f3c3-9535-4445-b08a-4dd15c270e4b",
  "controller" => "cdl/cdl",
  "action" => "charge"
}
```

## Full Backtrace

```
app/adapters/instrumented_adapter.rb:75:in `block in save'
app/adapters/instrumented_adapter.rb:48:in `block in trace'
app/adapters/instrumented_adapter.rb:44:in `trace'
app/adapters/instrumented_adapter.rb:74:in `save'
app/change_set_persisters/change_set_persister.rb:99:in `save'
app/resources/cdl/charge_manager.rb:49:in `create_charge'
app/resources/cdl/cdl_controller.rb:13:in `charge'
```

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/67197483)

tpendragon:
In the most recent case it looks like someone hit the checkout button twice.

tpendragon:
It seems like someone's slamming that button a bunch too and managing to take up all the available item charges for themselves.

tpendragon:
We may have solved this. I've resolved the CDL ticket and I'm marking this closed. We should reopen this if honeybadger errors about it again.

PR #4791, opened 2021-11-17
Don't create two charge entries for the same person.

tpendragon:
Discovered via #4755

### DIFF
--- a/app/resources/cdl/charge_manager.rb
+++ b/app/resources/cdl/charge_manager.rb
@@ -39,6 +39,8 @@ module CDL
     end
 
     def create_charge(netid:)
+      existing_charge = resource_charge_list.charge_for(netid: netid)
+      return existing_charge if existing_charge.present?
       raise CDL::UnavailableForCharge unless available_for_charge?(netid: netid)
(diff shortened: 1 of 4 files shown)


hackartisan (approved):

app/resources/cdl/charge_manager.rb
```diff
@@ -9,7 +9,7 @@ class ChargeManager
     include ActionView::Helpers::DateHelper
     attr_reader :resource_id, :eligible_item_service, :change_set_persister
     attr_writer :resource_charge_list
-    delegate :charged_items, :pending_or_active_holds, :active_hold?, :hold?, :expired_holds, to: :resource_charge_list
+    delegate :charged_items, :pending_or_active_holds, :active_hold?, :hold?, :expired_holds, :active_charge?, to: :resource_charge_list
```
hackartisan, line 12: It seems like you added this and then didn't use it? Is it left over from an amended commit? It doesn't bother me per se but we probably could take it back out.

hackartisan (changes requested):

hackartisan (changes requested):
@tpendragon it looks like there's a rebase issue here? There are extra commits.

tpendragon:
@hackartisan Thanks for catching that, that's what I get for switching machines.

hackartisan (approved):

Issue #4712, opened 2021-10-11
Refactor LinkedData::Resource.for


tpendragon:
It's ignoring a bunch of cops and is pretty complicated.

eliotjordan:
Working Refactorama branch: https://github.com/pulibrary/figgy/compare/4712-refactor-linked-data-resource

hackartisan:
For next time:

- Should ScannedResource handle the simple resource change set situation
- Once it's a single line do we still need a factory method?

PR #4813, opened 2021-11-29
[Refactorama] Move linked_resource tests.

tpendragon:
Refs #4712


hackartisan:
We could move the test for `#resource` to https://github.com/pulibrary/figgy/blob/main/spec/shared_specs/models/linked_data/linked_resource.rb

hackartisan (approved):

PR #4863, opened 2022-01-05
Provide raster visibility in mosaic endpoint response

eliotjordan:
- Adds visibility to rasterset and raster documents
- Updates the routes to the more generic `titiler`. Non-set rasters need a visibility value as well and renaming avoids confusion.

### DIFF
--- a/app/resources/raster_resources/raster_resources_controller.rb
+++ b/app/resources/raster_resources/raster_resources_controller.rb
@@ -19,12 +19,14 @@ class RasterResourcesController < ResourceController
       mosaic_path = MosaicService.new(resource: resource).path
       respond_to do |f|
         f.json do
-          render json: { uri: mosaic_path }
+          render json: { uri: mosaic_path, visibility: resource.visibility&.first }
         end
       end
     else
       respond_to do |f|
-        f.json { head :not_found }
+        f.json do
+          render json: { visibility: resource.visibility&.first }
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


tpendragon:
How are we planning to use this visibility key? It's not clear to me what the use cases are here and how we plan to handle each.

eliotjordan:
I guess I was thinking, at a basic level we could IP restrict the data. The edge lambda would check if it's Princeton-only and then compare the client IP with a list of Princeton IP ranges. Reject if not in range.

tpendragon:
I think I'd want to split out the use cases here and come up with some solutions for them. There's nothing really wrong with this, but I don't want to get too far down the line of supporting restricted datasets until we have open ones well supported. It's likely we'd need some form of callback to Figgy and I want to have a chance to think through the auth and caching pieces.

eliotjordan:
I get that, and I can put this PR in draft, but from my experience from the last few days, a callback to figgy on every tile call just isn't going to work. Far too slow for a good user experience. 

Also, from what I've seen, we really don't have many public raster datasets (hence the weird sample dataset that was provided). We're going to have to figure out the restricted data issue pretty soon IMO.

tpendragon:
@eliotjordan Thank you. I just don't want to go too far down the restricted content path when we need to tighten up the easy path first. There's no guarantee we'll have the cycles to have a good "open" dataset path and still have time to do restricted ones, so I want to make sure we nail this and get the gotchas and then have a good sit down to properly architect restricted content when we know everything we can know about performant open content display.

hackartisan (changes requested):
In a way `titiler` is more generic than `mosaic` but in a way it's also more specific because it references a specific service. Is the response very unique to that service, or should we callit something like `tiles`?

PR #4872, opened 2022-01-08
Fingerprint mosaic.json files

### LINKED ISSUE #4859: Add RasterSet fingerprinting to mosaic service
See #4770

### LINKED ISSUE #4873: Run mosaic.json creation job in raster derivative service
1. Job should be put on the queue after cloud copy is uploaded to S3.
1. Job should only be run if the RasterSet is "complete"

eliotjordan:
- Creates fingerprinted mosaic.json files
- Runs mosaic generation job after creation or cleanup of raster file set derivatives.

Closes #4859 
Closes #4873

### DIFF
--- a/app/services/mosaic_service.rb
+++ b/app/services/mosaic_service.rb
@@ -15,11 +15,13 @@ class MosaicService
     mosaic_path = Valkyrie::Storage::Disk::BucketedStorage.new(base_path: base_path).generate(resource: resource, original_filename: fingerprinted_filename, file: nil).to_s
     return mosaic_path if storage_adapter.find_by(id: mosaic_file_id)
   rescue Valkyrie::StorageAdapter::FileNotFound
-    # build default mosaic file
-    build_node(default_filename) if MosaicGenerator.new(output_path: tmp_file.path, raster_paths: raster_paths).run
+    raise Error unless MosaicGenerator.new(output_path: tmp_file.path, raster_paths: raster_paths).run
 
-    # build fingerprinted mosaic file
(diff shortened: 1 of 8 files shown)


app/services/mosaic_service.rb
```diff
@@ -10,13 +10,16 @@ def initialize(resource:)
...
-    mosaic_path = Valkyrie::Storage::Disk::BucketedStorage.new(base_path: base_path).generate(resource: resource, original_filename: "mosaic.json", file: nil).to_s
+    mosaic_path = Valkyrie::Storage::Disk::BucketedStorage.new(base_path: base_path).generate(resource: resource, original_filename: fingerprinted_filename, file: nil).to_s
     return mosaic_path if storage_adapter.find_by(id: mosaic_file_id)
   rescue Valkyrie::StorageAdapter::FileNotFound
-    build_node if MosaicGenerator.new(output_path: tmp_file.path, raster_paths: raster_paths).run
+    # build default mosaic file
+    build_node(default_filename) if MosaicGenerator.new(output_path: tmp_file.path, raster_paths: raster_paths).run
+
+    # build fingerprinted mosaic file
+    build_node(fingerprinted_filename) if MosaicGenerator.new(output_path: tmp_file.path, raster_paths: raster_paths).run
```
tpendragon, line 22: Does this run the mosaic generation a second time? Can we use the created mosaic from the previous step?

tpendragon (changes requested):
One small change and a question about optimization, then I think we're good.

app/services/mosaic_service.rb
eliotjordan: Good catch. We should be able to do that unless the temp file is deleted in between.

tpendragon (approved):

PR #4877, opened 2022-01-11
Exclude raster set members from search results

### LINKED ISSUE #4876: Generating a PDF the first time locks up a Passenger thread, eventually freezing up a Figgy machine.
If a user eventually tries again because it's slow (or maybe even tries several times), it can lock up an entire box (or multiple boxes.)

We'll need to come up with some way of backgrounding the generation of PDFs and notifying them when it's ready or something.

We discussed the options below and decided to go with option 5:
Generate in the background and provide a progress bar (see https://groups.google.com/g/samvera-tech/c/yThEBMzA4_o/m/QTZUz-JYCAAJ)

Prevent generating PDF twice, the job should have a guard that checks to see if the PDF is already being generated and if it is the job exits.

Steps:

- [x] #6278
- [x] #6279
- [x] #6280
- [x] #6281
- [x] #6282
- [x] #6283
- [x] #6284

hackartisan:
closes #4876

Screen Shot 2022-01-11 at 4 11 08 PM

### DIFF
--- a/app/indexers/member_of_indexer.rb
+++ b/app/indexers/member_of_indexer.rb
@@ -7,10 +7,14 @@ class MemberOfIndexer
   end
 
   def to_solr
-    return {} unless resource.is_a?(ScannedResource) || resource.is_a?(ScannedMap)
+    return {} unless index_as_member?
     {
       "member_of_ssim" => parents.map { |x| "id-#{x.id}" }.to_a
     }
+  end
+
+  def index_as_member?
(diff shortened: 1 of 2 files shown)


tpendragon (dismissed):

hackartisan:
@tpendragon thanks for the review. I changed the name of the method I added from `index_as_child?` to `index_as_member?` -- can you review again?

tpendragon (approved):

PR #4943, opened 2022-01-27
Generate tile metadata for MapSets with raster children

### LINKED ISSUE #4934: Allow tilemetadata route to generate mosiacs for ScannedMaps with Raster grandchildren
https://github.com/pulibrary/figgy/blob/main/app/controllers/tile_metadata_controller.rb#L7

eliotjordan:
Closes #4934

### DIFF
--- a/app/controllers/tile_metadata_controller.rb
+++ b/app/controllers/tile_metadata_controller.rb
@@ -1,19 +1,18 @@
 # frozen_string_literal: true
 class TileMetadataController < ApplicationController
   attr_reader :resource
+  rescue_from MosaicService::Error, with: :not_found
 
   def metadata
     @resource = find_resource(params[:id])
-    if resource.is_a?(RasterResource) && resource.decorate.raster_set?
+    if resource.is_a?(RasterResource) || resource.is_a?(ScannedMap)
       respond_to do |f|
         f.json do
           render json: { uri: cached_mosaic_path }
         end
       end
     else
-      respond_to do |f|
-        f.json { head :not_found }
-      end
+      not_found
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/tile_metadata_controller.rb
```diff
@@ -19,6 +19,16 @@ def metadata
 
   private
 
+    def generate_mosaic?
+      if resource.is_a?(RasterResource) && resource.decorate.raster_set?
+        true
+      elsif resource.is_a?(ScannedMap) && resource.decorate.map_set?
```
hackartisan, line 25: Won't this give a false positive for map sets that don't have any raster children?

hackartisan (changes requested):

spec/factories/scanned_map.rb
```diff
@@ -67,5 +67,21 @@
     factory :complete_scanned_map do
       state "complete"
     end
+
+    factory :scanned_map_with_raster_children do
```
tpendragon, line 71: :heart:

app/controllers/tile_metadata_controller.rb
tpendragon: Also can we just generate this for all scanned maps/rasters? If it's a scanned map with no raster children will something blow up?

eliotjordan: It won't blow up, just returns an error. To me, seems better than running the mosaic file_set query multiple times per request. We could rescue that error and return a 404.

eliotjordan: Fixed

hackartisan (approved):

app/controllers/tile_metadata_controller.rb
```diff
@@ -1,19 +1,18 @@
 # frozen_string_literal: true
 class TileMetadataController < ApplicationController
   attr_reader :resource
+  rescue_from MosaicService::Error, with: :not_found
 
   def metadata
     @resource = find_resource(params[:id])
-    if resource.is_a?(RasterResource) && resource.decorate.raster_set?
+    if resource.is_a?(RasterResource) || resource.is_a?(ScannedMap)
```
hackartisan, line 8: Let's add a code comment here saying why we purposefully include things we know will not have a mosaic service.

hackartisan (changes requested):

hackartisan (approved):

PR #4950, opened 2022-01-28
Make FileSet service_targets property editable

### LINKED ISSUE #4928: FileSet#service_targets should be editable in the form
If a user uploads a file manually through the interface they should be able to mark it as a mosaic service target.

eliotjordan:
Closes #4928 
Screen Shot 2022-01-28 at 4 31 39 PM


tpendragon (changes requested):
Suggestion: Can we make this a dropdown with just "mosaic" and none as an option?

eliotjordan:
@tpendragon Converted to drop down:
Screen Shot 2022-01-28 at 5 05 53 PM

tpendragon (approved):

PR #4953, opened 2022-01-31
Add wmts service in geo document_builder

### LINKED ISSUE #4898: ScannedMaps with raster resources should have IIIF and mosaic web service endpoints.
To support new mosaic workflow.

Close this ticket when MapSet records have the mosaic web service indexed.

thanyabegum:
Closes #4898

### DIFF
--- a/app/resources/scanned_maps/scanned_map_decorator.rb
+++ b/app/resources/scanned_maps/scanned_map_decorator.rb
@@ -50,6 +50,7 @@ class ScannedMapDecorator < Valkyrie::ResourceDecorator
            :geo_metadata_members,
            :members,
            :parents,
+           :mosaic_file_set_count,
            to: :wayfinder
 
   def collection_slugs
--- a/app/services/geo_discovery/document_builder/wxs.rb
+++ b/app/services/geo_discovery/document_builder/wxs.rb
@@ -43,7 +43,7 @@ module GeoDiscovery
       # Returns the wmts server url.
       # @return [String] wmts server url
       def wmts_path
-        return unless visibility == "open" && (raster_set? || raster_file_set?)
+        return unless generate_wmts_path?
         id = resource_decorator.id.to_s.delete("-")
         "#{tileserver_path}/mosaicjson/WMTSCapabilities.xml?id=#{id}"
       end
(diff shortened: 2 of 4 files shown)


app/services/geo_discovery/document_builder/wxs.rb
```diff
@@ -81,6 +81,15 @@ def generate_wms_path?
           @config && visibility && file_set && (raster_file_set? || vector_file_set?)
         end
 
+        # Determines if the wmts path should be generated
+        # @return [Boolean]
+        def generate_wmts_path?
+          return false unless visibility == "open"
+          return true if raster_set?
+          return true if raster_file_set?
+          resource_decorator.object.decorate.try(:mosaic_file_sets).present?
```
tpendragon, line 90: Suggestion: Add `Wayfinder.for(resource).try(:mosaic_file_set_count)`, and then make that method use `custom_queries.find_deep_children_with_property(resource: resource, model: FileSet, property: :service_targets, value: "mosaic", count: true)` - that should be faster and more stable than serializing all the file sets into objects.

tpendragon (changes requested):
Just one suggested change. Open to discussion on it too.

app/services/geo_discovery/document_builder/wxs.rb
hackartisan: Sounds good. Do we want to just change the one we added and continue going through the resource decorator? Or we aren't trying to contain use of wayfinders? There's a lot of using them via the resource decorators.

tpendragon: Yeah that's a good call, let's continue using the decorator.

eliotjordan (dismissed):

app/resources/scanned_maps/scanned_map_decorator.rb
```diff
@@ -50,6 +50,7 @@ def attachable_objects
            :geo_metadata_members,
            :members,
            :parents,
+           :mosaic_file_sets,
```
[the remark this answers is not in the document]
hackartisan: @eliotjordan the builder just goes ahead and adds the service for anything that's a raster resource or raster fileset. Is that not the correct behavior?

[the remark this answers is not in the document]
hackartisan: We had a discussion about this, and whether the RasterSet case needs to also check for service type. @eliotjordan will create a new issue.

hackartisan:
superceded by #4958 and #4962

PR #4973, opened 2022-02-03
Fix BibID recognition for bulk ingest.

### LINKED ISSUE #4970: Save and Ingest for MVWs is not working for Digital Studio
### Summary
The digital studio is unable to utilize the save and ingest function (perhaps tied to bulk ingest) for MVWs.

### Impact 
This is hindering, but not ceasing work in the digital studio. It is certainly slowing work down.

### Priority recommendation

- [x] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

_Required if "asap" or "within the next 3 weeks" is checked_
The amount of MVWS needing ingest from the studio moves this up to a high priority due to daily workflows and needs for multiple departments. 

### Expected behaviour
Properly staged ingests from studio server should ingest as MVWs.
### Actual behaviour
As of 2/1, figgy ingests record but no images. Workaround is manual attachment of volumes to ingested record.
### Steps to reproduce behaviour
Stage MVW from DPUL directory on studio share and use "save and ingest".

tpendragon:
There were directories where each sub-directory for MVWs were named "1", "2" etc so that the volumes would be named "1", "2", etc. However, the system was seeing those as metadata IDs and then breaking when the IngestFolderJob tried to fetch metadata for them.

Closes #4970


app/services/pul_metadata_services/client.rb
```diff
@@ -38,6 +38,9 @@ def retrieve_aspace_pulfa_ead(id)
       # @param source_metadata_id [String] the remote metadata identifier
       # @return [Boolean]
       def bibdata?(source_metadata_id)
+        # 3506421 will be in all alma IDs, and old Voyager records are
```
hackartisan, line 41: We should update this comment. only 6421 are guaranteed precisely (although there will always be a 2-digit code at the beginning as well)

I just found this posted in slack by @mzelesky:

`99 Record type; 350 - Voyager source; 6421 - Princeton's Alma`

hackartisan (changes requested):
Can you update the PR description to explain what was happening? There were directories with short numbers for names and the code thought they were bib ids?

A small change request inline.

Thank you!

tpendragon:
@hackartisan Should be ready again.

hackartisan (changes requested):
👍🏻 

Do we want a ticket instead of a TODO here?

hackartisan (approved):

hackartisan:
issue: https://github.com/pulibrary/figgy/issues/4977

PR #4978, opened 2022-02-08
Begin separating out resources subdirectory

tpendragon:
We've had some trouble parsing the `resources` directory internally and figuring out where to put things.

This is a new proposal and the beginning of the separation, open for draft and discussion:

1. Put things in models if they persist to a database or support persisting to a database.
2. Move things in models that don't do that out to `values` or `services`
3. ChangeSets should be namespaced under the model they're attached to - `ScannedResource::SimpleChangeSet`, for example.
4. Wayfinders, similarly, should be namespaced under their model. For example, `ScannedResource::Wayfinder`.
5. Controllers, jobs, and mailers go in their normal location for Rails.

Included in this PR so far is moving controllers, moving things out of `models`, and separating the `resources/cdl` directory.


tpendragon:
@hackartisan I think the main difference between this and that world are that change sets are in `app/change_sets` - do you see another one?

I was hoping that by keeping them namespaced under the models they modify that we'd better be able to see the lenses under which a model can be seen. Would you prefer we do `app/change_sets`?

hackartisan:
I will check out this branch and take a closer look.

Thank you for moving us forward in a direction, regardless!!

hackartisan:
Okay I had somehow missed that you had moved the controllers out. I think my biggest use case was "I want to run all the controller tests" so this does make me happy.

I do worry a little that we still have `app/change_sets/`, `app/decorators`, `app/wayfinders` where not all the change sets decorators or wayfinders are found. But I like your point about keeping the change sets with the resource they apply to. That may help more with onboarding than having a more traditional file structure.

`app/resources` not being in `app/models` I guess is an amendment to point 1 on your list. Will we turn this into an ADR?

On the whole I am feeling good about this. Let's do it.

tpendragon:
Closed in favor of #5061

PR #4980, opened 2022-02-08
Notify on file appender error

### LINKED ISSUE #4946: Consider making FileAppender raise an error again, rather than swallowing them
If the FileAppender errors right now, we end up with a broken resource and nobody knows.

Problematic code:

https://github.com/pulibrary/figgy/blob/9c92d2359e441fb383afe3f30421bd3e4a2de06f/app/services/file_appender.rb#L131-L134

Ingests are failing and not fixing themselves, and in cases where there are hundreds of pages it's very easy to miss.

hackartisan:
closes #4946 

review: commits are split into refactors, work.


tpendragon (approved):
That's a nice refactor, good work!

Issue #4844, opened 2021-12-17
Update Ruby to 2.7


tpendragon:
Ruby 2.6 is EOL April 2022.

PR #5001, opened 2022-03-01
Upgrade to Ruby 2.7

tpendragon:
Refs #4844


tpendragon:
Confirmed this seems to work in staging.

app/controllers/application_controller.rb
```diff
@@ -23,7 +23,7 @@ def after_sign_in_path_for(resource_or_scope)
   end
 
   def guest_uid_authentication_key(key)
-    key &&= nil unless key.to_s =~ /^guest/
+    key &&= nil unless /^guest/.match?(key.to_s)
```
hackartisan, line 26: I wonder why this is preferred

hackartisan: answer: performance, apparently. https://www.rubydoc.info/gems/rubocop/0.47.1/RuboCop/Cop/Performance/RegexpMatch

app/services/bulk_ingest_service.rb
```diff
@@ -221,6 +220,17 @@ def to_a
         nodes
       end
 
+      def mime_type(file_path)
+        basename = File.basename(file_path)
```
hackartisan, line 224: do we need this line? It looks like we do this right before calling the method, and don't use file_path anywhere else in this method. just accept basename?

config/initializers/figgy.rb
```diff
@@ -30,15 +30,13 @@ def campus_ip_ranges
...
-          IPAddr.new(address.to_s)
-        end
-      rescue
-        Rails.logger.debug("Unable to resolve #{fqdn}")
-        nil
+      address = Dnsruby::Resolver.new.query(fqdn).answer[0]&.address
+      if address.present?
+        IPAddr.new(address.to_s)
       end
+                            rescue
```
hackartisan, line 37: is this really indented like this?

hackartisan (changes requested):
This looks good. The style changes seem good. I had a couple of very small inline.

hackartisan (approved):

PR #5016, opened 2022-03-17
Add multispectral use case.

tpendragon:
The technical implementation is up for discussion in this PR - the only things set in stone are the folder structure and file naming convention. What's presented here is a solution which means we can ingest these in a tight timeline with no development work.

### DIFF
--- a/docs/use_cases.md
+++ b/docs/use_cases.md
@@ -42,6 +42,10 @@ indicate values to be filled in, square brackets represent optional values)
       - 💿 Princeton_\<MMSID\>\_f\<sheet
       number\>\<r/v\>\[x/y/z\]\_###\<N/R/G\>\_\<shot_number\>\_\<R/F\>.tif
       (### is wavelength, N/R/G represent filter - none, red, or green)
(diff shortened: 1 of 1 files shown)


docs/use_cases.md
```diff
@@ -25,3 +25,32 @@ The scanned map and the georeferenced tiff need to stay affiliated with one anot
...
+      number\>\<r/v\>\[x/y/z\]\_###\<N/R/G\>\_\<shot_number\>\_\<R/F\>.tif
+      (### is wavelength, N/R/G represent filter - none, red, or green)
+
+For more on the meaning of the filename see the contractor's
+[README](https://docs.google.com/document/d/1rjVgnUizdSrR1EsxaTV5RyHIOEWyUze8kjyS9ICPkyM/edit?usp=sharing)
+
+These resources are ingested via bulk ingest or "Save and Ingest" and results in
+a multi-volume work resource with one volume per "page" and 25 images per side
+of page. If the resource is already digitized then the catalog will have two
+viewers - one for the multispectral images and one for the resource as a whole.
```
hackartisan, line 52: so 2 manifests?

tpendragon: Yes.

PR #5040, opened 2022-04-06
Allow download of original pdf from viewer

### LINKED ISSUE #5020: Allow download of original PDF
If the resource was uploaded as a PDF, downloading the pdf should short-circuit pdf generation and provide the original PDF.

Original question:

We have a large set of Chinese TOCs we would like to host somewhere, preferably Figgy.  But they are searchable PDFs and it’s important that this feature be preserved.  I know Figgy can do OCR for Chinese, but even at its best it will be a step down from the files in their original format.

This content isn’t really right for DSpace, so we are just trying to figure out where we can make these files available in their current format.

Sample file:
https://drive.google.com/drive/u/2/folders/1UCWTYy9OGpw2iUklm58ULjpPzGGed15N

Follow up with @tventimi 

## Success Criteria

When a user has permission to download a resource which was generated from a PDF they should have a link in the UV to download the Original PDF instead of the generated PDF.

hackartisan:
- Use catalog pdf route in manifest
- Fix path in scanned resource request spec
- Pass auth token through from catalog pdf route to concerns pdf route
- Allow original pdf download with token or auth

closes #5020

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -215,12 +215,14 @@ class CatalogController < ApplicationController
     end
   end
 
+  # This endpoint redirects to download an original pdf if present,
+  # or redirects to the resource pdf endpoint if one needs to be generated
   def pdf
     _, @document = fetch(params[:solr_document_id])
 
     source_pdf = Wayfinder.for(resource).source_pdf
-    if source_pdf && can?(:download, source_pdf)
-      redirect_to "#{request.base_url}#{helpers.fileset_download_path(source_pdf)}"
+    if source_pdf
+      redirect_to "#{request.base_url}#{helpers.fileset_download_path(source_pdf, auth_token_param)}"
     elsif helpers.pdf_allowed?(resource) && can?(:pdf, resource)
       redirect_to "#{request.base_url}#{polymorphic_path([:pdf, resource], auth_token_param)}"
     else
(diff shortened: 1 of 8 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -215,16 +215,25 @@ def lookup_manifest
...
 
+  # This endpoint redirects to download an original pdf if present,
+  # or redirects to the resource pdf endpoint if one needs to be generated
   def pdf
     _, @document = fetch(params[:solr_document_id])
 
     source_pdf = Wayfinder.for(resource).source_pdf
-    if source_pdf && can?(:download, source_pdf)
-      redirect_to "#{request.base_url}#{helpers.fileset_download_path(source_pdf)}"
+    if source_pdf
```
tpendragon, line 224: Does the removal of the download permission check here change anything? I'm trying to think if there were cases where I could PDF but not download, is that a thing?

hackartisan: Where you could download the pdf but not the original files? I'm not aware of any and it seems unlikely. but not outside the realm of possibility. I'm not really sure how to figure this out.

the downloads controller does authorize before allowing the download, which is why it seemed like removing this was okay.

https://github.com/pulibrary/figgy/blob/b03688010aab8a803c32750404bf8f8e05a9251c/app/controllers/downloads_controller.rb#L91-L94

Definitely appreciate confirmation on whether this seems right.

PR #5041, opened 2022-04-06
Create rake task for migrating rasters

### LINKED ISSUE #4960: Migrating existing rasters
1. Re-run the derivates to generate a cloud-raster
2. Add a service target ("tiles") See: https://github.com/pulibrary/figgy/issues/4959

thanyabegum:
Refs #4960 -- don't close that issue until the migration has been run on prod.

### DIFF
--- a/app/services/migrations/add_service_target_migrator.rb
+++ b/app/services/migrations/add_service_target_migrator.rb
@@ -10,12 +10,12 @@ module Migrations
 
     def run
       query_service.find_all_of_model(model: RasterResource).each do |raster_resource|
-        Wayfinder.for(raster_resource).file_sets.each do |file_set|
-          change_set = ChangeSet.for(file_set)
-          change_set.validate(service_targets: ["tiles"])
-          change_set_persister.save(change_set: change_set)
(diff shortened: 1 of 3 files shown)


app/services/migrations/add_service_target_migrator.rb
```diff
@@ -0,0 +1,28 @@
...
+  class AddServiceTargetMigrator
+    def self.call
+      new.run
+    end
+
+    delegate :query_service, to: :change_set_persister
+
+    def run
+      query_service.find_all_of_model(model: RasterResource).each do |raster_resource|
+        Wayfinder.for(raster_resource).file_sets.each do |file_set|
```
tpendragon, line 13: We have a mosaic in production now with uncropped tiles, so I think we only want to migrate rasters where there's only one FileSet underneath the Raster.

Maybe something like:

```ruby
  file_sets = Wayfinder.for(raster_resource).file_sets
  if file_sets.length == 1
    # migrate
  end
```

tpendragon (dismissed):

PR #5051, opened 2022-04-08
Add GeoTiff to the manifest download pane.

### LINKED ISSUE #5015: ScannedMaps with Rasters should have a "Rendering" in the manifest for downloading the GeoTIff
## Reasoning

Right now there's no way to get the GeoTiff when looking at a MapSet in the viewer. Adding a rendering will make it show up in the download pane of the UV.

tpendragon:
Closes #5015


spec/services/manifest_builder_spec.rb
```diff
@@ -981,4 +981,21 @@ def logical_structure(file_set_id)
       expect(media_sequence["elements"][0]["label"]).to eq scanned_resource.title.first
     end
   end
+
+  context "when given a MapSet with Vector children" do
```
hackartisan, line 985: title should say "with Raster children"

hackartisan (changes requested):
This looks good to me. One spec title needs to be fixed.

hackartisan (approved):

PR #5061, opened 2022-04-11
Get rid of `app/resources`

tpendragon:
This is a much simpler version of #4978 which was actually completable.

This just separates out change_sets/decorators/wayfinders/controllers into their own directories again.


spec/change_sets/recording_change_set_spec.rb
```diff
@@ -2,106 +2,31 @@
 require "rails_helper"
 
 RSpec.describe RecordingChangeSet do
-  subject(:change_set) { described_class.new(form_resource) }
```
hackartisan, line 5: Why are there a bunch of tests removed? Are they duplicates or something? I wonder if this should be in a separate PR.

hackartisan (changes requested):
One question but once the test suite passes I'm 👍🏻

spec/change_sets/recording_change_set_spec.rb
tpendragon: Good catch - we had two recording_change_set specs, so my renaming killed one of them. They're combined now.

eliotjordan (approved):

Issue #5056, opened 2022-04-08
Restricted maps should have a copyrighted rights statement


tpendragon:
Find all the maps that are restricted and mark them as copyrighted.

## Sudden Priority Justification

Users are sending us (and the map lab) emails because the rights statement says there's no copyright, when there is.

PR #5070, opened 2022-04-13
Adds migration rake task for updating rights statements ... 

sdellis:
... for maps with no known copyright and campus-only access

refs #5056

### DIFF
--- a/spec/services/migrations/map_copyright_migrator_spec.rb
+++ b/spec/services/migrations/map_copyright_migrator_spec.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Migrations::MapCopyrightMigrator do
+  describe ".call" do
+    context "when a map is restricted" do
+      it "updates the rights statement to be \"in copyright\"" do
+        stub_ezid(shoulder: "99999/fk4", blade: "123456")
+        scanned_map = FactoryBot.create_for_repository(:complete_campus_only_scanned_map)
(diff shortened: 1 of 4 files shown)


eliotjordan (dismissed):
This looks good. 👍

spec/services/migrations/map_copyright_migrator_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+require "rails_helper"
+
+RSpec.describe Migrations::MapCopyrightMigrator do
+  describe ".call" do
+    context "when a map is restricted" do
+      it "updates the rights statement to be \"in copyright\"" do
+        stub_ezid(shoulder: "99999/fk4", blade: "123456")
+        scanned_map = FactoryBot.create_for_repository(:complete_campus_only_scanned_map)
+        raster_map = FactoryBot.create_for_repository(:complete_campus_only_raster_resource)
+        vector_map = FactoryBot.create_for_repository(:complete_campus_only_vector_resource)
```
hackartisan, line 12: I think it would be good to add one geo resource that's open and then an expectation that its rights statement has not changed.

hackartisan (changes requested):

spec/services/migrations/map_copyright_migrator_spec.rb
sdellis: Good point, @hackartisan . I just pushed a commit that adds this expectation.

hackartisan (approved):
Great! I think we should try it on staging and then merge / run it on prod.

PR #5111, opened 2022-05-06
Restrict refresh_remote_metadata route

### LINKED ISSUE #5108: Add authorization to resource metadata update endpoint
We shouldn't allow all systems to issue a resource metadata update.

## Implementation Proposal

Add a permission for `can :refresh_metadata, Valkyrie::Resource`, give it to everyone in the group `metadata_refreshers`. In the Figgy interface create an auth token for Pulfalight which gives the `metadata_refreshers` group, then add that auth token to the Pulfalight request.

eliotjordan:
Closes #5108 

Tested on staging.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -11,6 +11,7 @@ class ResourcesController < ApplicationController
   skip_before_action :verify_authenticity_token, only: [:refresh_remote_metadata]
 
   def refresh_remote_metadata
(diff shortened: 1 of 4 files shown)


tpendragon (approved):

tpendragon (changes requested):
Looks great, just blocking this so we can get the pulfalight bit in first. Good work!

tpendragon (approved):

PR #5117, opened 2022-05-16
Don't raise after a bulk update notification.

### LINKED ISSUE #5116: Don't retry a bulk update after sending email notification
We want to swallow the error so the job doesn't retry, error again, and spam the user.

hackartisan:
Raising means the job re-runs and sends another email.
closes #5116

### DIFF
--- a/app/jobs/bulk_update_job.rb
+++ b/app/jobs/bulk_update_job.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 class BulkUpdateJob < ApplicationJob
+  # rubocop:disable Lint/NonLocalExitFromIterator
   def perform(ids:, email:, args:, time:, search_params:)
     change_set_persister.buffer_into_index do |buffered_change_set_persister|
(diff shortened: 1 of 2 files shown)


hackartisan:
I added the rubocop disable because I think at some point we may want to accumulate all errors and send them in a single email. if/when we do that we won't need the return anymore.

hackartisan:
Make bulk update a database object?

tpendragon:
Or manually requeue with an `alerted:` flag? I'm open to retrying not being useful at all here though, I just realized it'd only happen if something didn't validate.

hackartisan:
It's already passing around a lot of arguments.

hackartisan:
Yeah I don't know that there's a scenario where a retry succeeds without intervention. And if the person intervenes and then starts up the bulk update job again we still have the other one retrying repeatedly.

tpendragon (approved):

tpendragon:
Works for me.

PR #5122, opened 2022-05-19
Refactor pulfa remote metadata

### LINKED ISSUE #5042: Adding a collection with a nonexistent source_metadata_id breaks the scanned resource form
To replicate
* create a collection and give it a source metadata id that looks like a component id but isn't a real component id. don't forget you have to fill in the slug field.
* Go to the new scanned resource form

This error was discovered when a collection was made from an unpublished finding aid. Once the finding aid was published and the metadata synced it resolved. https://figgy.princeton.edu/catalog/2c11ef96-c7ba-4756-b7c5-d30a97e13193

Resolve either by not allowing the object to be created (this was expected behavior and may be preferred) or by ensuring collections with nil titles are filtered out just before the location of the error.

Honeybadger:
https://app.honeybadger.io/projects/53391/faults/84371050

```

1: <% if !resource_has_parents?(f.object.resource, params) || f.object.kind_of?(EphemeraFolderChangeSetBase) %>
    2:   <%= f.input :member_of_collection_ids,
    3:               collection: @collections.sort_by(&:title),
    4:               label_method: :title,
    5:               value_method: :id,
    6:               input_html: { multiple: f.object.multiple?(key) },
```
```
app/views/records/edit_fields/_member_of_collection_ids.html.erb:3:in `sort_by'
app/views/records/edit_fields/_member_of_collection_ids.html.erb:3:in `_app_views_records_edit_fields__member_of_collection_ids_html_erb__1137239044168021665_605200'
app/views/base/_form_metadata_inner.html.erb:3:in `block in _app_views_base__form_metadata_inner_html_erb__3945062815367994064_603580'
app/views/base/_form_metadata_inner.html.erb:2:in `each'
app/views/base/_form_metadata_inner.html.erb:2:in `_app_views_base__form_metadata_inner_html_erb__3945062815367994064_603580'
app/views/base/_form_metadata.html.erb:4:in `_app_views_base__form_metadata_html_erb__2340373240580214423_603560'
app/views/base/_form.html.erb:7:in `block in _app_views_base__form_html_erb__2554498001300087046_603540'
app/views/base/_form.html.erb:2:in `_app_views_base__form_html_erb__2554498001300087046_603540'
app/views/base/new.html.erb:1:in `_app_views_base_new_html_erb___4234022672507813375_42183560'
```

eliotjordan:
Closes #5042

- Refactor pulfa remote metadata
- Rename stub_aspace method
- Update findingaids config url
- Remove old pulfa fixtures
- Update stub_aspace_error method name


spec/characterization_services/external_metadata_characterization_service_spec.rb
```diff
@@ -42,7 +42,7 @@
   end
 
   context "with an non-geo metadata file" do
-    let(:file) { fixture_file_upload("files/pulfa/MC016/c9616.xml", "application/xml") }
+    let(:file) { fixture_file_upload("files/geo_metadata/non-geo-metadata.xml", "application/xml") }
```
escowles, line 45: Just reading through to see where this is headed, I noticed this change looks like it's different from the description above. Should this remain a non-geo metadata file, or should we update the description?

eliotjordan: The test used `c9616.xml` to test that Figgy doesn't add the `schema` mime type parameter.  to non-FGDC and non-ISO XML files. The file was chosen at a random. As far as I can tell the spec description is still accurate. I suppose we could move that file out of the /geo_metada/ directory to make it less confusing. Or use an another random xml file from fixtures.

escowles: That's not necessary — just making sure I was missing something (which I was).

tpendragon (approved):

tpendragon:
This is amazing and such a tough cleanup. Really great work @eliotjordan

PR #5142, opened 2022-06-03
Upgrades ordermanager to Bootstrap 4

### LINKED ISSUE #5126: Bootstrap 4: check / fix order manager
On the rails-6-again long-running branch

sdellis:
Fixes #5126

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -63,13 +63,8 @@ GIT
       ttfunk
 
 GEM
-  remote: https://gems.contribsys.com/
-  specs:
-    sidekiq-pro (3.7.1)
-      sidekiq (>= 4.1.5)
-
-GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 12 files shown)


hackartisan (changes requested):
I think we need to rebase this PR and add back in one or more of the helpers that were removed because they weren't being used (see https://github.com/pulibrary/figgy/pull/5138/commits/db95cffccdb7aee279563ff925492a091e1fa01a)

eliotjordan:
@hackartisan Page title generation no longer uses those methods. They can be removed.

sdellis:
It looks like the coverage went down on this one. I imagine it's from adding back in the helper methods? I don't see any tests for those helper methods in the previous code. Is there any way to tell what code is not covered @hackartisan ?

hackartisan (approved):

PR #5159, opened 2022-06-13
Remove Uglifier, allow deploys.

tpendragon:


config/environments/production.rb
```diff
@@ -5,7 +5,7 @@
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
   config.public_file_server.enabled = true
-  config.assets.js_compressor = :uglifier
+  # config.assets.js_compressor = Uglifier.new(harmony: true)
```
eliotjordan, line 8: question: do you think we'll be adding this back? If not, we should delete the line.

eliotjordan (changes requested):

PR #5216, opened 2022-06-27
Enable ajax selct on accessions edit form

eliotjordan:
Fixes AJAX select widgets on accessions edit form.

Adding `class=lux` for each div does not work - and I'm not sure why. The select widget is duplicated multiple times. The solution fI came up with for accessions is to add the `lux` class once by overriding the form partial.


hackartisan (changes requested):
Could you please add to the description a brief summary of what was broken and how this fixes it?

hackartisan:
Thank you!

PR #5228, opened 2022-07-08
Try out new select for Ephemera.

tpendragon:


hackartisan (changes requested):
It's a bit tight at the top when the dropdown appears. It looks like we can add a padding-top on the `.select2-results__options`

tpendragon:
@hackartisan Can you screen shot what you mean by tight at the top?

hackartisan:
Screen Shot 2022-07-08 at 3 17 23 PM

tpendragon:
Uh, weird, I don't think that's how it looks to me.

hackartisan:
@tpendragon It looks like the issue is that it opens up already scrolled to the bottom. can you replicate that on this object? https://figgy-staging.princeton.edu/concern/ephemera_folders/040473cf-c141-4be9-b7f6-58ffd947df59/edit

tpendragon:
Yup, weird, doesn't seem to be the case for new items.

hackartisan:
It's still better than without this fix. I won't block on it if you want to ticket and mergre.

tpendragon:
I think this broke Numismatics, per @eliotjordan 

```
Unforunately, it seems to have broken Issue and Coin forms. See Metal, Shape, Color: https://figgy-staging.princeton.edu/concern/numismatics/issues/new
```

tpendragon:
This wasn't what broke numismatics, but I had to enable lux for all the Ajax-Select'y fields to get this to work.

eliotjordan (approved):

tpendragon:
Created https://github.com/pulibrary/figgy/issues/5234

PR #5262, opened 2022-07-21
Remove Hydra Editor

### LINKED ISSUE #5264: Remove Hydra Editor
It's preventing us from upgrading to Rails 6.1 and we don't use much of it.

tpendragon:
Closes #5264


hackartisan (changes requested):
This looks great my only question is whether we need some new tests.

Even just some feature tests, maybe, that check for the inputs and exercise the add / remove additional values buttons?

hackartisan (approved):

eliotjordan (approved):
👍

PR #5290, opened 2022-07-26
Fix Numismatics Forms

### LINKED ISSUE #5273: Numismatics - Accessions - new / edit form duplicated drop downs

### LINKED ISSUE #5274: Numismatics - Coin - new/edit form “numismatics collection” field is non-functional

### LINKED ISSUE #5276: Numismatics - Coin - new/edit page has repeating select dropdowns

### LINKED ISSUE #5277: Numismatic Issue edit form - artist:person and citation:reference have multiple copies of the select dropdown

tpendragon:
Removes all the lux tags, changes where we initialize the JS, and uses a single lux tag for the application.

The key here seems to be that you need to run non-vue JS -after- the vue is mounted, or it doesn't work. When I swapped everything to webpacker I made vue happen at the end on accident, which is what broke everything and necessitated controlled lux tags.

Previously we had two lux tags, one for the header and one for the body. It turned out that the numismatics code depended on that - the first lux tag would initialize vue, and add vue tags to the page in the HTML, and then the second would initialize vue and actually mount those components. To have one lux tag we have to add vue tags in beforeCreate instead so the mounting process will catch them, and then bind the cocoon helpers after mounting so the JS will work.

Closes #5276 
Closes #5273 
Closes #5274 
Closes #5277

### DIFF
--- a/app/javascript/helpers/setup_ajax_select.js
+++ b/app/javascript/helpers/setup_ajax_select.js
@@ -34,13 +34,15 @@ function handleCocoonAfterInsert (event, elements) {
 
 // Inserts an ajax-select component before every input with a ajax_select_type
 // attribute. Example:  `<input ajax_select_type="Place"/>`
-export default function setupAjaxSelect () {
+export function setupAjaxSelect () {
   const ajaxInputs = document.querySelectorAll('input[ajax_select_type]')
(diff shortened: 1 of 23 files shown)


hackartisan (approved):
nice work!! 🎉

PR #5294, opened 2022-07-26
Give render_file an absolute path to the 404.

### LINKED ISSUE #5269: DEPRECATION WARNING: render file
DEPRECATION WARNING: render file: should be given the absolute path to a file (called from show at /home/circleci/figgy/app/controllers/downloads_controller.rb:10)

tpendragon:
Closes #5269

Note: I'm not convinced this is the right thing to do, https://github.com/samvera/hydra-head/pull/548 might be instead.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -69,6 +69,13 @@ class DownloadsController < ApplicationController
     nil
   end
 
+  def render_404
+    respond_to do |format|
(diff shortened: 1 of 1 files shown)


hackartisan:
Did you change this after you left the comment about how it should be done? I am not seeing the distinction.

tpendragon:
The distinction is it might be better to patch upstream rather than override the method

hackartisan:
Aha. That does seem like a nice thing to do.

PR #5304, opened 2022-07-28
Fix launching rails server.

tpendragon:
Initializers shouldn't call "require" and we have to avoid the mini
profiler patch problem causing an infinite loop.


hackartisan (approved):

hackartisan:
yes this fixes it

PR #5337, opened 2022-08-03
Add local parallel testing

tpendragon:
Cuts the test suite locally in half (on my machine).


eliotjordan (commented):
I'm getting some intermittent failures after each test run. The most recent:

```

rspec ./spec/exporters/hathi/submission_information_package_spec.rb:33 # Hathi::SubmissionInformationPackage.deposit creates a SIP directory on disk
rspec ./spec/jobs/export_hathi_sip_job_spec.rb:32 # ExportHathiSipJob.perform exports the object
rspec ./spec/exporters/hathi/submission_information_package_spec.rb:40 # Hathi::SubmissionInformationPackage sip contents contains the right content
rspec ./spec/services/tile_metadata_service_spec.rb:56 # TileMetadataService#path when the file already exists on the storage adapter returns the path
rspec ./spec/adapters/bagit/storage_adapter_spec.rb:6 # Bagit::StorageAdapter behaves like a Valkyrie::StorageAdapter doesn't leave a file handle open on upload/find_by
```

eliotjordan:
The test that fails consistently for me is: 

`rspec ./spec/adapters/bagit/storage_adapter_spec.rb:6`

tpendragon:
> rspec ./spec/adapters/bagit/storage_adapter_spec.rb:6

This fails for me with `rspec spec` too - it's the open files test right?

tpendragon:
@hackartisan I think increasing the timeouts fixed the solr connection thing, can you give it a shot?

hackartisan:
@tpendragon can you rebase it to get the node update in .tool-versions? node is all weird for me.

tpendragon:
@hackartisan Done

eliotjordan (dismissed):

README.md
```diff
@@ -96,6 +96,18 @@ destroy` or turn off all services with `lando poweroff`.
...
 
+### Parallel Tests
+
+If you'd like to run the test suite in parallel do the following:
+
+1. `bundle exec rake servers:start`
+1. `PARALLEL_TEST_FIRST_IS_1=true RAILS_ENV=test rake parallel:setup`
+1. `./bin/turbo_test`
+
+To run in parallel with working coverage, but worse output, do:
```
hackartisan, line 107: What does "worse output" mean? Output looks as expected to me, but I have test format set in a global .gitconfig file. Actually it's the other one that doesn't respect that format setting.

README.md
```diff
@@ -96,6 +96,18 @@ destroy` or turn off all services with `lando poweroff`.
...
+
+1. `bundle exec parallel_rspec --first-is-1 --runtime-log tmp/turbo_rspec_runtime.log`
```
hackartisan, line 109: Coverage is a primary benefit of running the full suite so I expect I'll always prefer this method. A `.bin` script that runs it would be nice.

README.md
```diff
@@ -96,6 +96,18 @@ destroy` or turn off all services with `lando poweroff`.
      - Or run services separately as shown in [[https://github.com/pulibrary/figgy/blob/master/Procfile]]
    - Access Figgy at http://localhost:3000/
 
+### Parallel Tests
+
+If you'd like to run the test suite in parallel do the following:
+
+1. `bundle exec rake servers:start`
+1. `PARALLEL_TEST_FIRST_IS_1=true RAILS_ENV=test rake parallel:setup`
```
hackartisan, line 104: This is only needed once? Should it be integrated into `.bin/setup`?

It's not clear from the layout whether this is a prerequisite for both `./bin/turbo_test` and `parallel_rspec`. The `parallel_rspec` list item starts over at 1 again due to the text between the two list items. Is that intended? Should the third list item have 2 sub-items with the 2 different ways to run the tests?

README.md
hackartisan: Is it not showing the full test run time all together?

hackartisan (commented):
For me the turbo tests bin script is now working and tests pass. 

The parallel_rspec method has > 200 failures.

hackartisan (dismissed):

hackartisan (approved):
I think we should squash these commits

PR #5338, opened 2022-08-04
Update service target migrator to handle multiple file sets

eliotjordan:
- Update service target migrator to handle multiple file sets
- Update Metrics/LineLength cop name

Works towards #4960

### DIFF
--- a/app/services/geo_discovery/geoblacklight_document.rb
+++ b/app/services/geo_discovery/geoblacklight_document.rb
@@ -66,7 +66,7 @@ module GeoDiscovery
           rights_statement_s: rendered_rights_statement
         }
       end
-      # rubocop:enable Metrics/LineLength, Metrics/AbcSize
+      # rubocop:enable Metrics/MethodLength
 
       def document_hash_required
         {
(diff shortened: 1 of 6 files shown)


app/services/geo_discovery/geoblacklight_document.rb
```diff
@@ -66,7 +66,7 @@ def document_hash_optional
           rights_statement_s: rendered_rights_statement
         }
       end
-      # rubocop:enable Metrics/LineLength, Metrics/AbcSize
+      # rubocop:enable Layout/LineLength, Metrics/AbcSize
```
hackartisan, line 69: also change line 45?

hackartisan (changes requested):
this looks good! I think just one more Metrics fix is needed.

app/services/geo_discovery/geoblacklight_document.rb
eliotjordan: Fixed.

hackartisan (approved):

Issue #5229, opened 2022-07-08
Add "Refresh metadata from PULFA/Voyager" to bulk edit functionality


kelea99:
### Summary
As a figgy user, I would like to "Refresh metadata from PULFA/Voyager" via the bulk edit feature in figgy. Sometimes I may hear of changes to the MARC or FA records for multiple items in a collection and manually editing each item would be cumbersome.
This is due to actual use cases:

- EAL #4652 
- Lapidus #5227 
- Middle East Manuscripts (slack communications) -Joyce Bell is doing massive remediation on a subset of Manuscripts from the Islamic World.

### Impact 
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._
This is not hindering work in figgy itself, but it is contributing to outdated/incorrect metadata being displayed to our users in our applications.

### Priority recommendation

- [ ] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "Maintenance/Research" labels_

tpendragon:
Status update: There's a PR #5230 for adding this, but it was determined that it had potential to use more Alma API hits than DACS was comfortable with. So once #3072 is closed we'll pull JSON-LD from Orangelight instead,  which will make no hits required.

hackartisan:
Are we sure the orangelight json page doesn't hit the alma api?

tpendragon:
I think the proposal is to add a new JSON-LD (.jsonld) endpoint which just uses the solr document.

PR #5343, opened 2022-08-08
Use catalog JSON-LD instead of Bibdata.

tpendragon:
Work towards #5229


app/models/processed_event.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
-# This resource represents a single bibdata dump and tracks whether it has been
+# This resource represents a single catalog dump and tracks whether it has been
 #   processed to refresh metadata from voyager
-# We need to do this because bibdata provides an ongoing list of dump events,
+# We need to do this because catalog provides an ongoing list of dump events,
```
hackartisan, line 4: Wait, catalog doesn't have dumps

hackartisan (changes requested):
you re-pulled all these fixtures?

hackartisan:
this is pretty amazing that it just drops in like that!

tpendragon:
Nope, didn't repull them, find and replace just got ahead of itself. I'll patch it later, thanks.

hackartisan (changes requested):
This was clearly a big job, thank you for doing it.

It seems like the change of title value may be pretty noticeable to users. It looks like maybe we used to pull just the `245a` (or maybe `a` and `b`) and now we have the whole `245` including medium (`245h`) and statement of responsibility (`245c`). Given that this will then percolate out to, e.g. DPUL, I'm a little concerned this isn't what we want. How can we get assurances that this will meet user expectations? Or do we need to request an update to the json from the catalog?

tpendragon:
@hackartisan You're talking about for titles? This changed a long time ago - we just never updated our fixtures. See for instance https://bibdata.princeton.edu/bibliographic/991234563506421/jsonld

hackartisan:
Okay, great!

hackartisan (approved):

Issue #5385, opened 2022-08-24
When a resource fails to validate in Bulk Ingest, the ingester should be notified.


tpendragon:
Email the depositor when bulk ingesting fails to create a resource - include the reason why validation failed, probably `resource.errors.full_messages`.

PR #5387, opened 2022-08-24
Add support for figgy_metadata.json 

### LINKED ISSUE #5374: Add a way to set Figgy metadata as part of bulk ingest.
## Use Case

When bulk ingesting ETDs every year each record will have a different visibility and embargo date. There should be a format so we can still use our existing bulk ingest UI, but every record has unique data for those properties.

## Proposal

Add support for a `figgy_metadata.json` that allows users to provide ChangeSet attributes. For example:

```json
{
  "visibility": "open",
  "embargo_date": "2022-01-02"
}
```

Part of #5367 

Properties in the bulk ingest interface should apply first, then those in the file should override those on a per-resource basis.

## Considerations

We considered doing YML and decided against it because YML tends to have more security problems.

We considered `resource_metadata.yml`, but felt `figgy_metadata.yml` was more descriptive and evokes thoughts of the types of resources in Figgy.

## Thoughts

We're not sure how we would handle invalid resources that were caused by putting bad metadata in the JSON file.

### Proposal for handling invalid resources

1. JSON Schema validation?
2. Capture failing resources & shoot them into their own job, notify somehow?
    * Maybe no job, just email the ingester that there were some resources that failed validation?

tpendragon:
Closes #5374 

This also propagate depositor for bulk ingest to help #5385, and secures that action (oops)


hackartisan (changes requested):
This looks good -- are we sure we don't want to keep these json files?

tpendragon:
> This looks good -- are we sure we don't want to keep these json files?

What's the use case for keeping them?

hackartisan:
Use case for keeping them would be:
- we have the original values if there is some bug during ingest
- we have the original values to validate ingest was correct if someone suggests there was a bug during ingest but there wasn't

tpendragon:
So the use case is:

I bulk ingest
I delete my files
I notice a bug
I blame Figgy

Yeah? Do you think we should attach the file itself? Store the values in a property somewhere?

eliotjordan (approved):

PR #5401, opened 2022-08-26
Add embargo date field

### LINKED ISSUE #5372: Add embargo_date property to all resources
There should be a spot in the form to edit it along with a datepicker. It should store ISO8601 dates that coordinate with midnight on a given day in the eastern time zone.

For instance, if I put in 2022-01-02, that means the date stored should be midnight on the January 2nd 2022 in the eastern time zone.

We might have to configure Rails' default time zone to Eastern to do this. See https://stackoverflow.com/questions/6118779/how-to-change-default-timezone-for-active-record-in-rails.

If that doesn't work, another option might be wrapping code in `Time.use_zone("Eastern Time (US & Canada)") do`, but this might get complicated.

Part of #5367

hackartisan:
closes #5372 

Next steps here are
- add the shared change set spec to all the other resources
- there are probably some opportunities for refactoring
- ensure the date picker is auto populating with the stored value (and are we sure we want that change set accessor to return a string? Or do we want it to have some other name that we call explicitly when setting up the lux element?)


tpendragon:
Superceded by #5404

PR #5408, opened 2022-09-01
Add visibility label for embargoed resources

### LINKED ISSUE #5406: Add some sort of notification on the Figgy show page that an item is embargo'd
Right now it's not clear that something's not showing up because it's under embargo. Maybe a banner, or adjusting the visibility label like we do with state?

eliotjordan:
Closes #5406 

- Adds visibility label for embargoed resources
- Adds an embargo service

Screen Shot 2022-09-01 at 12 13 05 PM

### DIFF
--- a/app/decorators/valkyrie/resource_decorator.rb
+++ b/app/decorators/valkyrie/resource_decorator.rb
@@ -45,7 +45,7 @@ class Valkyrie::ResourceDecorator < ApplicationDecorator
 
   def visibility
     Array(super).map do |visibility|
-      h.visibility_badge(visibility, public_readable_state?)
+      h.visibility_badge(visibility, public_readable_state?, embargoed?)
     end
   end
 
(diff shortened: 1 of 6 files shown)


app/decorators/valkyrie/resource_decorator.rb
```diff
@@ -187,6 +187,12 @@ def public_readable_state?
     workflow_class.public_read_states.include? Array.wrap(state).first.underscore
   end
 
+  # Is the resource embargoed?
+  # @return [TrueClass, FalseClass]
+  def embargoed?
+    EmbargoService.resource_embargoed?(model)
```
tpendragon, line 193: Question: This is interesting - what led you to extract a class here instead of putting the function here and using the decorator in the ability?

tpendragon (changes requested):
I'm not really blocking on this, but I'm interested in the design decision.

app/decorators/valkyrie/resource_decorator.rb
tpendragon: Or even putting it in Resource, although I'm not positive that has enough coverage.

eliotjordan: The decorator is better place for it. I was initially thinking I would add more methods to dry up date processing in a helper, but that doesn't seem necessary. Moved the method over.

tpendragon (approved):

PR #5409, opened 2022-09-01
Ingest ArchivalMediaBags via Bulk Ingest

### LINKED ISSUE #5307: Add a bulk ingest screen for Recordings from a Bag
### Summary

Archival Media Bags need to be able to be ingested into collections which aren't "Archival Media Collections." We decided on building a bulk ingest screen like our existing one, but for Recordings, which will allow users to select an AV bag.

### Impact 

Without this we're unable to ingest bags of digitized audio material that are patron requests.

### Priority recommendation

- [ ] asap
- [x] within the next 3 weeks
- [ ] PO will prioritize

Patron requests are a core service, and Kelly Bolding is unable to fulfill them without this feature.

tpendragon:
Closes #5307


tpendragon:
I'm marking this ready for review.

The wireframes proposed putting a "Bulk Ingest" button next to recordings. However, the heading for that section is "Media Reserves", adding a feature that actually made it so you can bulk ingest resources would mean adjusting the BulkIngestController to handle change sets, and just adding bag support to BulkIngest generally solves this use case.

This would mean Kelly would just hit the bulk ingest button next to "Scanned Resources" to ingest bags.

hackartisan:
> This would mean Kelly would just hit the bulk ingest button next to "Scanned Resources" to ingest bags.

It's not just Kelly, though, right? Is there anything we can do to make it a little more obvious where to go? Maybe add a new button that goes to the same form?

tpendragon:
> It's not just Kelly, though, right? Is there anything we can do to make it a little more obvious where to go? Maybe add a new button that goes to the same form?

It's not just Kelly, but the whole workflow is managed through a documentation process right now. I'm not sure a "Bulk Ingest" button next to Recording is any easier, and I think closer to what we'd need is a full rework of how accessing and interfacing with "Bulk Ingest" is done so that it's obvious that's where you want to go and what your options are when you're in there.

Maybe I'm wrong though?

tpendragon:
On home page:

Change "Media Reserves" to say "Audio", add a new box which says "Vendor Bags" and has a button that goes to the scanned resource Bulk Ingest screen. Adjust the title of the scanned resource Bulk Ingest screen to say "Bulk Ingest Scanned Resources or Vendor Bags"

tpendragon:
Screen Shot 2022-09-08 at 12 28 07 PM
Screen Shot 2022-09-08 at 12 28 29 PM

hackartisan (changes requested):
I see where you're recommending that the archival media collection go away. Questions:

1. Does this mean the workflow becomes:
   - Create a collection
   - bulk ingest the bag into that collection
1. In the meantime does that form still work? i.e. it will still create a new collection for you?


Also, can we have Kelly at our wrapup to demo this for her?

Code-wise it looks pretty good to me. There were a lot of detailed changes to make. I like that the code becomes more general, feels like we're moving in the right direction.

Blocking for questions not changes.

tpendragon:
> * Does this mean the workflow becomes:
>   
>   * Create a collection
>   * bulk ingest the bag into that collection

Yes

> * In the meantime does that form still work? i.e. it will still create a new collection for you?

Uhh. I think it won't add it to a collection if they use the AMC method. Maybe we should remove it from the home page?

hackartisan:
I think we should remove it if it won't behave the same. I also think it would be good to have a chat with Kelly about the shape these changes have taken.

hackartisan:
> I think we should remove it if it won't behave the same. I also think it would be good to have a chat with Kelly about the shape these changes have taken.

not necessarily in that order

PR #5425, opened 2022-09-14
Always return "unauthorized" for Reading Room resources.

tpendragon:
Discover now means "someone could get the digital content for this potentially." For Reading Room material they could always participate in the OARSC program, so always mark reading room resources as discoverable. This will make the OARSC "yellow box" appear in Pulfalight.

Closes pulibrary/pulfalight#1096


hackartisan (approved):

hackartisan:
Nice! I feared this would be more complicated.

PR #5443, opened 2022-09-27
Consistency in NotFoundErrors

### LINKED ISSUE #5442: [Figgy/production] Blacklight::Exceptions::RecordNotFound: The solr permissions search handler didn't return anything for id "dbeaedd8-acf5-42f4-a85d-76575569dae9"
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/87085726)

hackartisan:
Treat Blacklight::Exceptions::NotFoundError the same as
Valkyrie::Persistence::ObjectNotFoundError

closes #5442

I'm not really sure about this. The other option would be to accept the current behavior and ignore the error in honeybadger. The error does show up like a regular 404 to the browser. Which is a default rails page :(

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -57,7 +57,7 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  rescue_from Valkyrie::Persistence::ObjectNotFoundError, with: :resource_not_found
+  rescue_from Valkyrie::Persistence::ObjectNotFoundError, Blacklight::Exceptions::RecordNotFound, with: :resource_not_found
   def resource_not_found(_exception)
     respond_to do |format|
       format.json { head :not_found }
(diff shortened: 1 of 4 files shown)


hackartisan:
Maybe it would be better to style the 404 page and make Valkyrie::Persistence::NotFoundError go there instead. It doesn't really seem right to return a 302 for a not found.

hackartisan:
Here's what it looks like now. I do think it's good to make this consistent with current behavior for the other not found error.

Screen Shot 2022-09-28 at 10 02 39 AM

tpendragon:
Is this error appearing when there's an indexing disconnect, and if someone hits the search results and this error happens it forces them out of search results?

If so, will we know when that happens now?

hackartisan:
Good questions. The error seems to be coming from EZID link checker; https://ezid.cdlib.org/. They started on August 1, when we deployed the rails / blacklight upgrade. I don't know what previous behavior was. I don't think we do anything with these reports or have ever done anything with anything like them. There have been > 50,000 of them in the 2 or so months since they started.

If we want to be able to check on this sort of thing on demand, we could write a log message. But we may just as well put that off until it seems like we need it.

tpendragon:
Ah, these are requests to the record pages, not the search results?

hackartisan:
Yes. It's deployed on staging -- See, e.g. figgy.princeton.edu/catalog/notathing vs figgy-staging.princeton.edu/catalog/notathing

tpendragon (approved):

PR #5460, opened 2022-10-06
I5370 graphql error alerts

sdellis:


sdellis:
We need a test that creates a new resource with a SAVED_STATE property that has a value of "ERROR". Then we can test for the existence of the alert message.

PR #5479, opened 2022-10-19
Add migration to regenerate derivatives on filesets with too many

### LINKED ISSUE #5352: fileset with too many derivatives
Is this right? https://figgy.princeton.edu/catalog/99bd0a3f-488c-4c36-8519-d75fb0a87b44

found while investigating errors from 8/4 on https://app.honeybadger.io/projects/53391/faults/84134442/01G9MZNGJDN3SD9FEJP9Z0FZ72#notice-summary

There were 587 errors on 8/4; one was a scanned resource image that needed a regenerated derivative.

Here's a list of ids: https://gist.github.com/hackartisan/9ee3224cc8b45cd9c4e1bb6e00b32317

hackartisan:
closes #5352

### DIFF
--- a/app/services/migrations/remove_extra_derivatives.rb
+++ b/app/services/migrations/remove_extra_derivatives.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Migrations
+  class RemoveExtraDerivatives
+    def self.call
+      new.run
+    end
+
+    def run
+      progress_bar
+      query_service.find_all_of_model(model: RasterResource).each do |resource|
+        Wayfinder.for(resource).file_sets.each do |file_set|
+          next unless extra_thumbnails?(file_set)
+
+          RegenerateDerivativesJob.perform_later(file_set.id.to_s)
+        end
+        progress_bar.progress += 1
+      end
+    end
+
+    private
+
+      def progress_bar
+        @progress_bar ||= ProgressBar.create format: "%a %e %P% Resources Processed: %c of %C", total: query_service.custom_queries.count_all_of_model(model: RasterResource)
+      end
+
+      # we only check for extra thumbnails because lots of resources might have
+      # extra derivatives (one jp2 and one tiff) from when we migrated to
+      # pyramidal tiffs, and the bug we're cleaning up from was failing after
+      # generating both the derivative and the thumbnail
+      def extra_thumbnails?(file_set)
+        file_set.file_metadata.count(&:thumbnail_file?) > 1
+      end
+
+      def query_service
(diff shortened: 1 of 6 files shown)


app/services/migrations/remove_extra_derivatives.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    private
+
+      def progress_bar
+        @progress_bar ||= ProgressBar.create format: "%a %e %P% Resources Processed: %c of %C", total: query_service.custom_queries.count_all_of_model(model: FileSet)
+      end
+
+      def extra_derivatives?(file_set)
+        file_set.file_metadata.count(&:derivative?) > 1
```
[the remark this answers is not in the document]
tpendragon: The above doesn't really work. Query to get file set IDs with more than two derivatives:

```sql
select a.id, COUNT(file_metadata_element) FROM orm_resources a, jsonb_array_elements(a.metadata->'file_metadata') AS file_metadata_element WHERE file_metadata_element @> '{"use": [{"@id": "http://pcdm.org/use#ServiceFile"}]}' GROUP BY a.id HAVING(count(file_metadata_element)) > 2 LIMIT 10;
```

It's pretty slow, getting 10 takes a while.

app/services/migrations/remove_extra_derivatives.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module Migrations
+  class RemoveExtraDerivatives
+    def self.call
+      new.run
+    end
+
+    def run
+      progress_bar
+      query_service.find_all_of_model(model: FileSet).each do |file_set|
```
tpendragon, line 11: This isn't memory efficient - it'll load every file set into memory. See https://github.com/samvera-labs/valkyrie-sequel/blob/main/lib/valkyrie/sequel/query_service.rb#L43-L47.

We created the MemoryEfficientAll query for reindexing, but that doesn't let us filter to a specific model - it just has a deny list.

app/services/migrations/remove_extra_derivatives.rb
tpendragon, line 26: It's looking like > 2 is a decent check, I'm running a longer query to see if there's any that we wouldn't want to regenerate that have that count.

tpendragon (changes requested):

app/services/migrations/remove_extra_derivatives.rb
hackartisan: we could try limiting it to file sets on raster resources, the bug was in the raster resource derivative service

tpendragon (approved):

PR #5484, opened 2022-10-25
Allow bulk ingest of jpegs

### LINKED ISSUE #5455: Figgy should allow bulk ingest of jpegs
Use case:
bcadmin\publicpolicy\MC147\ML.2022.013\AAIA

The images were scanned as part of an agreement with an e-resource vendor, who provided us with the images in exchange for an exclusive sales contract for a certain number of years.

We will be able to provide these images freely in a couple of years. In the meantime we'd like them safe in figgy. They are all jpgs. There are almost 6000 resources, the collection is "AAIA (Gale)": https://figgy.princeton.edu/catalog/3c3d2cb3-bc58-4782-84c6-d70df7e36bde. Roel estimates 30,000 images.

Here's the line where we only get tiffs:
https://github.com/pulibrary/figgy/blob/3abfe2c68d4816499759b6a11603a59bb18985b0/app/services/bulk_ingest_intermediate_service.rb#L75

Current workaround is to have the studio convert the jpgs to tiffs and then ingest them.

We expect to have more use cases like this, where we are provided jpgs and need them in figgy.

## Sudden priority justification
It's must slower to have to convert things to tiffs, and involves more peoples' work time. We think this is a small change we can make in code.

taylor-yamashita:
closes #5455


hackartisan (commented):
This seems right. It would be good to try bulk ingesting some jpgs in a local or staging instance to validate.

tpendragon (changes requested):
This is a great change, but I'm not sure it will work for normal bulk ingest.

Can you try out a test like https://github.com/pulibrary/figgy/blob/eb1f6d1310279ff528094ca7d265d0a9c7126388/spec/jobs/ingest_folder_job_spec.rb#L48-L82, but for jpegs, and see where that leads you?

app/services/bulk_ingest_intermediate_service.rb
```diff
@@ -51,13 +51,13 @@ def background?
     end
 
     # Parse and extract a numeric index from a file path
-    # @param [String] tiff_entry the file path
+    # @param [String] image_entry the file path
     # @return [String]
-    def extract_index(tiff_entry)
-      index_m = /\/(\d+)\./.match(tiff_entry)
+    def extract_index(image_entry)
```
tpendragon, line 56: Renaming variables is important! Thank you!

app/jobs/ingest_folder_job.rb
```diff
@@ -23,7 +23,7 @@ def perform(directory:, property: nil, change_set_param: nil, class_name: "Scann
     def typed_file_filter(class_name)
       case class_name
       when "ScannedResource"
-        [".tif", ".wav", ".pdf", ".zip"]
+        [".tif", ".wav", ".pdf", ".zip", ".jpg"]
```
tpendragon, line 26: Good find!

tpendragon (approved):
Awesome work, glad to see the test could drive the implementation.

I'm not merging because you've requested a couple other reviewers. If when you get in again this seems fine to you, you should feel free to merge it.

PR #5497, opened 2022-11-01
Import content_warning from Pulfalight

### LINKED ISSUE #5490: Import content_warning from remote metadata sources
Related to https://github.com/pulibrary/pulfalight/issues/1195

When there's a content_warning in a remote source, we should import it.

tpendragon:
Closes #5490 

I refuse to make change_set_persister_spec bigger, so this also moves a bunch of the "remote record" tests to its own file, which is where I added the new test.


hackartisan (approved):
👍🏻 👍🏻

PR #5500, opened 2022-11-02
Add a remove from collection option to bulk edit

### LINKED ISSUE #5101: Add "Remove from a Collection" to bulk edit functionality
### Summary
_As a figgy user, I would like to remove items from a collection via the bulk edit feature in figgy. Sometimes I may accidentally add a lot of items to a collection  i did not intend to and manually editing each item would be cumbersome._ 
This is due to an actual use  case.  A user accidentally added a large number of items to a collection in figgy that they would like  removed. It would  be useful to give users the ability to  remove items from collections as well via the bulk edit feature. 
### Impact 
There are no hard deadlines for this
### Priority recommendation

- [ ] asap
- [ ] within the next 3 weeks
- [x] PO will prioritize - Will discuss with DLS to prioritize as a team

eliotjordan:
Closes #5101 

Screen Shot 2022-11-02 at 2 33 42 PM


hackartisan (changes requested):
The code looks good to me.

I wonder if it should be placed next to the "add to collection" form element. Does it make sense to even try to group these in some way?

We talked about having a warning that items removed from a collection can no longer be bulk edited via that collection. I'm wondering if there is some usability testing we could or should do here.

eliotjordan:
Update:

- "Remove from Collect" moved below "Add to Collection"
- The collection used to query the bulk edit is filtered out of the removable collection list

There was a suggestion to limit the removable collections to the collections the resources are a member of. I did not implement this at it would add quite a bit of complexity and is not really needed. It can be added, though, if that is the consensus.

hackartisan (approved):

PR #5501, opened 2022-11-03
Add specific content warnings when they're available.

### LINKED ISSUE #5492: When content_warning is populated from remote metadata, display that content warning in the viewer
See mockup:

https://docs.google.com/presentation/d/1eNAiwQB4eMog7BQffaGQXmIPIPb-3UEEEmqXGmzLW4A/edit#slide=id.g16995e60172_3_0

Related to https://github.com/pulibrary/pulfalight/issues/1195 and https://github.com/pulibrary/figgy/issues/5490

## Implementation Details

We should populate `notice` in the graphql with this information when it's available. Probably in https://github.com/pulibrary/figgy/blob/harmful_warnings/app/graphql/types/resource.rb#L111-L116

tpendragon:
Closes #5492 

This includes a bunch of refactors @hackartisan and I had discussed before. Doing them made it easier to add a vocab that's not in ControlledVocabulary.all, but which has config for the visual display colocated with the ones which are.

There's also some refactors which utilizes polymorphism.

Screen Shot 2022-11-02 at 5 26 27 PM

### DIFF
--- a/app/graphql/types/resource.rb
+++ b/app/graphql/types/resource.rb
@@ -109,10 +109,7 @@ module Types::Resource
   end
 
   def notice
-    return if object.try(:notice_type).blank?
-    ControlledVocabulary.for(:notice_type).find(object.notice_type.first).to_graphql.tap do |notice_graphql|
-      notice_graphql[:text_html] = notice_graphql[:definition]
-    end
+    ControlledVocabulary.for(:notice_type).for(object)&.to_graphql
   end
 
   def query_service
--- a/app/models/controlled_vocabulary.rb
+++ b/app/models/controlled_vocabulary.rb
@@ -324,11 +324,35 @@ class ControlledVocabulary
       @notice_authority_config ||= YAML.safe_load(File.read(Rails.root.join("config", "authorities", "notices.yml")), [Symbol])
     end
 
+    # Return the notice for a given resource.
+    # @return [Term]
+    def for(resource)
+      if resource.try(:primary_imported_metadata)&.content_warning.present?
+        NoticeTerm.new(value: "specific_harmful_content", definition: resource.primary_imported_metadata.content_warning)
+      elsif resource.try(:notice_type).present?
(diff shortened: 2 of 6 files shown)


app/graphql/types/resource.rb
```diff
@@ -109,10 +109,7 @@ def embed
   end
 
   def notice
-    return if object.try(:notice_type).blank?
-    ControlledVocabulary.for(:notice_type).find(object.notice_type.first).to_graphql.tap do |notice_graphql|
-      notice_graphql[:text_html] = notice_graphql[:definition]
-    end
+    ControlledVocabulary.for(:notice_type).for(object)&.to_graphql
```
hackartisan, line 112: I like how this invocation turned out

app/models/controlled_vocabulary.rb
```diff
@@ -324,11 +324,35 @@ def self.authority_config
       @notice_authority_config ||= YAML.safe_load(File.read(Rails.root.join("config", "authorities", "notices.yml")), [Symbol])
     end
 
+    # Return the notice for a given resource.
+    # @return [Term]
+    def for(resource)
+      if resource.try(:primary_imported_metadata)&.content_warning.present?
+        NoticeTerm.new(value: "specific_harmful_content", definition: resource.primary_imported_metadata.content_warning)
```
hackartisan, line 331: It's a little confusing that it's loaded in as the definition here, then on line 349 (via the method on 353) it's referred to as the message, while actually getting the full "definition" from the i18n file.

I wonder if we could use "message" in the i18n file and keep the use of "definition" just for the `specific_harmful_content` terms? Or maybe there's some other way to greater clarity here.

hackartisan (changes requested):
I think this works well! Just one bit that I think we could make more clear.

tpendragon:
@hackartisan Good suggestion, does that look better?

hackartisan (approved):
👍🏻

PR #5513, opened 2022-11-07
Change version.txt cap task to fetch SHA from revisions.log

m-ichelleliu:


hackartisan (changes requested):
We also need to delete the `after "deploy:published", "write_version"` on line 77.

hackartisan (approved):

PR #5514, opened 2022-11-08
Allow access to JSON-LD by overriding access controls

### LINKED ISSUE #1367: Staff users can't delete FileSets they have created

eliotjordan:
Closes #1367


hackartisan:
close in favor of #5522

PR #5523, opened 2022-11-11
Use wayfinders instead of decorators where possible

### LINKED ISSUE #2689: Use wayfinders instead of decorators where possible
There will be lots of places in the code that need this change. But it's important to do so we can get to a consistent state.

taylor-yamashita:
#2689

### DIFF
--- a/app/change_set_persisters/change_set_persister/generate_mosaic.rb
+++ b/app/change_set_persisters/change_set_persister/generate_mosaic.rb
@@ -23,11 +23,12 @@ class ChangeSetPersister
       end
 
       def mosaic?(resource)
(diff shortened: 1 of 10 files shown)


hackartisan (changes requested):
This is looking great. I think we should look to merge it and start a new branch if there's more work to do, so this doesn't sit around too long.

taylor-yamashita:
> This is looking great. I think we should look to merge it and start a new branch if there's more work to do, so this doesn't sit around too long.

Sounds good! This one is ready to merge. I'll start another branch for additional changes.

hackartisan (approved):

Issue #5399, opened 2022-08-26
Better handle single raster images


eliotjordan:
We have done a lot of work on mosaicked GeoTIFFS, but have not implemented single GeoTIFF services fully.

The WMTS and XYZ service URLs for single GeoTIFFs should be of the form:
`https://map-tiles.princeton.edu/cog/tiles/WebMercatorQuad/{z}/{x}/{y}...`

sdellis:
@eliotjordan will add a link to an example and state what it should look like to close this ticket.

eliotjordan:
Types of resources that should have a viewer and an XYZ link the GeoBlacklight document.

Single raster:
https://figgy.princeton.edu/catalog/006ab85b-6924-446e-8eed-b8f107e25351

Single scanned map with a single child raster:
Need example

PR #5527, opened 2022-11-15
Allow tile metadata service to generate tilejson for scanned map with single raster

eliotjordan:
Works towards #5399 

Screen Shot 2022-11-15 at 1 54 15 PM

### DIFF
--- a/app/services/tile_metadata_service.rb
+++ b/app/services/tile_metadata_service.rb
@@ -11,6 +11,7 @@ class TileMetadataService
   end
 
   def path
+    raise Error if raster_file_sets.empty?
     if mosaic?
       # Path to mosaic.json file
       mosaic_path
@@ -21,13 +22,13 @@ class TileMetadataService
   end
 
   def mosaic?
-    return true if resource.is_a?(RasterResource) && resource.decorate.raster_resources_count.positive?
+    # A mosaic is single service comprised of multiple raster datasets.
+    # This tests if there are multiple child raster FileSets.
     return true if raster_file_sets.count > 1
     false
   end
 
   def mosaic_path
-    raise Error if raster_file_sets.empty?
     document_path = Valkyrie::Storage::Disk::BucketedStorage.new(base_path: base_path).generate(resource: resource, original_filename: fingerprinted_filename, file: nil).to_s
     return document_path if storage_adapter.find_by(id: mosaic_file_id)
   rescue Valkyrie::StorageAdapter::FileNotFound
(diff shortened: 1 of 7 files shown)


app/values/tile_path.rb
```diff
@@ -48,13 +48,20 @@ def tileserver
...
       file_count = resource.decorate.try(:mosaic_file_count)
       return false unless file_count&.positive?
-      return false if resource.is_a?(ScannedMap) && resource.decorate.scanned_maps_count <= 1
       true
     end
 
     def mosaic?
-      return true unless resource.is_a?(RasterResource)
-      return true if resource.decorate.decorated_raster_resources.count.positive?
+      return true if resource.is_a?(RasterResource) && resource.decorate.raster_resources_count > 1
```
hackartisan, line 55: I think we'd like to prefer `Wayfinder.for(resource).raster_resources_count` over `resource.decorate.raster_resources_count`

app/services/tile_metadata_service.rb
```diff
@@ -21,8 +21,8 @@ def path
   end
 
   def mosaic?
-    return true unless resource.is_a?(RasterResource)
-    return true if resource.decorate.raster_resources_count.positive?
+    return true if resource.is_a?(RasterResource) && resource.decorate.raster_resources_count.positive?
```
hackartisan, line 24: Prefer using a wayfinder rather than a decorator here as well

hackartisan (changes requested):
It's not obvious to me why we're checking for positive or `> 1` in any given case. Maybe you could add comments to say why these are the checks we use?

eliotjordan:
@hackartisan I simplified the mosaic method, added some comments, and fixed some logic that was incorrect in our original implementation. A "mosaic" should have two or more raster files.

spec/services/tile_metadata_service_spec.rb
```diff
@@ -63,14 +65,6 @@
         expect(path).to eq(cloud_path.join("33", "1d", "70", "331d70a54bd94a6580e4763c8f6b34fd", "mosaic-#{fingerprint}.json").to_s)
       end
     end
-
-    context "when there aren't any files on the raster members" do
```
hackartisan, line 67: Why don't we want a test for this case anymore?

hackartisan (changes requested):
Thank you that is much more clear! I just have a question about this deleted test. Does it does something else now? Should we update it? Not sure why we want to delete it.

spec/services/tile_metadata_service_spec.rb
eliotjordan: Good catch. We do, I just needed to move the `raise Error` elsewhere.

hackartisan (approved):

PR #5564, opened 2022-12-05
Don't allow deleting empty collections.

### LINKED ISSUE #5529: Only empty collections should be able to be deleted
Reasoning: It would be difficult to support restoring a collection from preservation and making the members re-attach. So to prevent having to do that, we can instead make it so collections with members can't be deleted. Users would have to bulk move members to another collection, and then delete it.

tpendragon:
Closes #5529

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -74,11 +74,19 @@ class ResourcesController < ApplicationController
   def destroy
     @change_set = ChangeSet.for(find_resource(params[:id]))
     authorize! :destroy, @change_set.resource
-    change_set_persister.buffer_into_index do |persist|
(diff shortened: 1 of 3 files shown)


app/controllers/resources_controller.rb
```diff
@@ -74,11 +74,19 @@ def after_create_success(obj, change_set)
...
+      after_delete_success
     end
-    flash[:alert] = "Deleted #{@change_set.resource}"
-    after_delete_success
+  end
+
+  # Overridable method to enable controlling delete behavior in inherited
+  # controllers. @change_set is accessible here.
+  def around_delete_action
+    yield
```
hackartisan, line 89: 👍🏻 Okay!

hackartisan (approved):

PR #5567, opened 2022-12-05
Return 404 for objects and files not found, instead of 302

m-ichelleliu:


hackartisan (changes requested):
The coverage report says the errors_controller has no test coverage. It looks like we could remove the controller and its route, but keep the template for rendering from the application controller.

hackartisan (approved):

PR #5576, opened 2022-12-07
Use primary_file over original_file

### LINKED ISSUE #5538: CheckFixityJob should use `primary_file` instead of `original_file`
This might be causing a bug where primary_files aren't having fixity status stored, might be worth checking.

### LINKED ISSUE #5555: Audit use of `original_file` throughout figgy
We've found several places where `original_file` should be replaced by `primary_file` and are probably causing bugs.

Audit all uses of `original_file` and change any that should use `primary_file`.

tpendragon:
Closes #5555 and closes #5538

### DIFF
--- a/app/services/manifest_builder/canvas_rendering_builder.rb
+++ b/app/services/manifest_builder/canvas_rendering_builder.rb
@@ -133,7 +133,7 @@ class ManifestBuilder
       end
 
       def original_file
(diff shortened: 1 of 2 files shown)


eliotjordan (approved):
👍

hackartisan (approved):

Issue #3592, opened 2020-01-16
Ensure only 1 PreservationObject is made per Resource.


tpendragon:
We saw a second one get created when re-saving a resource. We cleaned it up already though - just check to ensure this is the case.

## Success Criteria

Write a query to see if there are more than one preservation object with the same resource ID. If so, we have to clean them up and figure out why they're being created.

eliotjordan:
Query run on the console:
```
SELECT metadata #>> '{"preserved_object_id",0,"id"}' as rid, COUNT(*)
FROM orm_resources orm 
WHERE internal_resource = 'PreservationObject'
GROUP BY rid
HAVING orm.count > 1
```

Found 8609 resources with two or more PreservationObjects.
https://docs.google.com/spreadsheets/d/1nhuTwoFj0w8Bsv1OA9Lw2-UK3qDW54nCpqc4Iv4JsUA

eliotjordan:
Example: This [FileSet](https://figgy.princeton.edu/catalog/0002c3b1-ae6a-4a79-a75a-658841356189) has two related PreservationObjects.

b60878bf-05db-4b07-9204-8d18a3eb987d 
dcd45a8e-79e2-4899-9037-a16f2cce28db

eliotjordan:
Query to get created dates in descending order:
```
WITH ids AS (
	SELECT metadata #>> '{"preserved_object_id",0,"id"}' as rid, COUNT(*)
	FROM orm_resources orm 
	WHERE internal_resource = 'PreservationObject'
	GROUP BY rid
	HAVING orm.count > 1
)
SELECT orm.id, orm.created_at, ids.rid
FROM (
	SELECT * FROM orm_resources
	WHERE internal_resource = 'PreservationObject'
	) as orm
INNER JOIN ids	
ON ids.rid = orm.metadata #>> '{"preserved_object_id",0,"id"}'
ORDER BY orm.created_at DESC
```

eliotjordan:
May have found the problem:
https://github.com/pulibrary/figgy/blob/main/app/services/preserver.rb#L78

The wayfinder method is `preservation_objects`, plural. Looks like a new preservation object is created with each save.

eliotjordan:
Most multiple PreservationObjects created for the same Resource resource link to this Coin:
https://figgy.princeton.edu/catalog/6c6994a7-4abd-4b6b-8488-38b128fb70c2

The PreserveResourceJob was queued 5 times in a 4 minute span.

[...]

PR #5584, opened 2022-12-09
Add optimistic lock token checking to PreserveResourceJob

eliotjordan:
- Adds optimistic lock property to (most) resources and change sets.
- Disables optimistic locking functionality by default.
- In PreserverResourceJob, adds logic to compare lock tokens passed as a param with current lock token values and exit early if they don't match.

refs #3592


eliotjordan:
Split into two PRs:
https://github.com/pulibrary/figgy/pull/5585
https://github.com/pulibrary/figgy/pull/5586

Issue #5589, opened 2022-12-13
Add 'current' property to Event


tpendragon:
The `current` property is true/false and says whether that event reflects current state or has been overridden by a newer event.

See https://github.com/pulibrary/figgy/issues/5545#issuecomment-1349529270 for context.

Code should
- [x] add the new property
- [x] When adding a new event that has the same `type`, `resource_id`, and `child_id`, mark the previously current event with that type, resource_id, and child_id as current: false.
- [x] Add a migration to set current true/false to existing events. There's over a million of them.

Rollout:
- [x] Turn off all fixity jobs: Pause the "super_low" queue to keep local checks from running after derivative generation via CheckFixityJob and cloud checks from running via UpdateFixityJob.
- [x] deploy code and run the migration
- [x] turn fixity jobs back on

## Implementation notes
Given the number of Events in the database it may be impractical to run the migration in ruby. How will we know if this is impractical? If it is impractical, we might use the first query / a variation thereof in the comment here https://github.com/pulibrary/figgy/issues/5545#issuecomment-1349550355 to do an update query for the migration.

PR #5595, opened 2022-12-13
Use events for tracking local fixity checks

### LINKED ISSUE #2431: CheckFixityJob should create Fixity Events instead of storing status as part of the FileSet
## Success Criteria

CheckFixityJob creates the same kind of events as cloud fixity UpdateFixityJob, and the appropriate displays are powered off of those.

We'll need either a data migration to create events from the succeeded fixity dates stored on FileSets, or have displays do a fallback check.

related: #1973, #2674, #3329

hackartisan:
closes #2431 

TODO:
- [x] have local fixity check create an event
- [x] Update show page local fixity displays to use events, fall back to using fields on file set.
- [x] Update Fixity Dashboard
- [x] rename jobs from `check_fixity_job` and `update_fixity_job` to `check_local_fixity_job` and `check_cloud_fixity_job`
- [x] Add a comment that the file_metadata fixity fields are deprecated

Dashboard update implementation notes:
* There are no failed local fixity checks currently. We can leverage that fact to simplify the query -- any failures will be on Events since that's how we'll record checks going forward.
* Recent checks will all have Events, as well.


hackartisan:
Try moving fallback logic from wayfinder into the custom query. Use a right outer join to get both the current event and the value of the primary file's fixity status. then select the right row conditionally (the event if it exists, otherwise the other one) and count them.

This will be blocked on #5589

tpendragon (approved):

PR #5596, opened 2022-12-14
Add more documentation about PreservationObject race condition & lock tokens

### LINKED ISSUE #5588: Add more in-code documentation about how checking lock tokens prevents creating extra preservation objects
There's a reference to why we do this here https://github.com/pulibrary/figgy/blob/97aac35677211897d661215cdd70e51616f1ac55/app/jobs/preserve_resource_job.rb#L27-L28

But it would be helpful to document some additional info about how the race condition plays out and how this prevents it.

tpendragon:
Closes #5588


hackartisan (approved):
This is helpful, thank you!

PR #5597, opened 2022-12-14
Use a unique index to prevent two PreservationObjects

### LINKED ISSUE #3592: Ensure only 1 PreservationObject is made per Resource.

tpendragon:
Closes #3592

This PR closes a tiny race condition where someone edits a resource, a PreservationJob kicks off, gets past the check for lock tokens, and then they edit again while that PreservationJob is running and before it saves a PreservationObject.

This unique index makes it impossible to have two PreservationObjects for the same resource. To deploy it we'll have to delete all duplicate PreservationObjects.


eliotjordan (changes requested):
Looks good, but could you rebase it against main? There have been some significant changes merged since this PR.

tpendragon:
@eliotjordan Sure, rebased.

eliotjordan (approved):

PR #5600, opened 2022-12-15
Enable Preserver::Importer to set correct PCDM use on recovered files

### LINKED ISSUE #5546: Tombstones should be created for any type of primary file
Currently only created for original files.

https://github.com/pulibrary/figgy/blob/5a6f2a033024f327603a4efe2794e7bb09f5415f/app/change_set_persisters/change_set_persister/create_tombstone.rb#L11

We should also look at Preserver::Importer.rb line 19-22, where it discards any Use information.

Steps:

- [x] Use `primary_file` instead of `original_file`
- [ ] enable importer to carry original PCDM use to restore files

eliotjordan:
- Enable Preserver::Importer to set correct PCDM use on recovered files
- Refactor Preserver::Importer

Closes #5546

### DIFF
--- a/app/services/preserver/importer.rb
+++ b/app/services/preserver/importer.rb
@@ -38,7 +38,7 @@ class Preserver
     end
 
     def import!
-      fs = build_file_set(metadata_file_identifier)
+      fs = build_file_set
       fs_change_set = ChangeSet.for(fs)
 
       files = import_binary_nodes(binary_file_identifiers)
@@ -73,11 +73,8 @@ class Preserver
         )
       end
 
-      def build_file_set(file_identifier)
-        return resource_class.new if file_identifier.nil?
-        metadata_file = storage_adapter.find_by(id: file_identifier)
-
-        metadata_file_contents = metadata_file.read
+      def build_file_set
+        return resource_class.new if metadata_file_identifier.nil?
         metadata_json = JSON.parse(metadata_file_contents)
         # Delete historic metadata because we are about to add new ones for the
         # restored file(s)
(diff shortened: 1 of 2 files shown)


app/services/preserver/importer.rb
```diff
@@ -11,23 +11,16 @@ def self.default_storage_adapter
...
-      metadata_file_identifier = if file_metadata.nil?
-                                   nil
-                                 else
-                                   file_metadata.file_identifiers.first
-                                 end
-      binary_nodes = resource.binary_nodes
-      binary_file_identifiers = binary_nodes.map(&:file_identifiers)
-      binary_file_identifiers.flatten!
-
+      metadata_file_identifier = file_metadata.try(:file_identifiers).try(:first)
```
tpendragon, line 14: Is this `.try` because it can be nil? If so, can we do `file_metadata&.file_identifiers&.first`?

app/services/preserver/importer.rb
```diff
@@ -38,77 +31,68 @@ def initialize(metadata_file_identifier:, binary_file_identifiers:, change_set_p
...
       fs_change_set.validate(files: files)
 
-      persisted = nil
+      # Delete historic metadata because we are about to add new ones for the restored file(s)
+      fs_change_set.validate(file_metadata: [])
+
+      imported = nil
       change_set_persister.buffer_into_index do |buffered_change_set_persister|
-        persisted = buffered_change_set_persister.save(change_set: fs_change_set, external_resource: true)
+        imported = buffered_change_set_persister.save(change_set: fs_change_set, external_resource: true)
```
tpendragon, line 43: 👍 Good rename!

app/services/preserver/importer.rb
```diff
@@ -38,77 +31,68 @@ def initialize(metadata_file_identifier:, binary_file_identifiers:, change_set_p
...
-      fs = build_file_set(metadata_file_identifier)
-      fs_change_set = ChangeSet.for(fs)
-
       files = import_binary_nodes(binary_file_identifiers)
+      fs_change_set = ChangeSet.for(file_set)
       fs_change_set.validate(files: files)
 
-      persisted = nil
+      # Delete historic metadata because we are about to add new ones for the restored file(s)
+      fs_change_set.validate(file_metadata: [])
```
tpendragon, line 39: Validating and not syncing to keep access to the old file_metadata is clever!

tpendragon (changes requested):
Really tiny change, I really like this though, good work!

app/services/preserver/importer.rb
hackartisan: Where is it used that way? It seems like everything is pulled from the old data before we get here. Is that why you made this change @eliotjordan?

hackartisan (changes requested):
I agree I like this refactor. If something clever is happening though I might want more comments about it 😁

app/services/preserver/importer.rb
eliotjordan: Good catch. I'm not sure why this would every be an issue, but there is a test for it and this line is a compact version of the previous logic.

app/services/preserver/importer.rb
eliotjordan: I didn't put too much thought into this, other than thinking I wanted to merge
`fs_change_set.validate(files: files)` and `fs_change_set.validate(file_metadata: [])` into a single line.

`fs_change_set.file_metadata = []` also works but #sync does not (though I'm not sure why). @hackartisan is right, I was thinking that the old file metadata is pulled out before we get here.

eliotjordan: So... nothing intentionally clever.

tpendragon (approved):

Issue #5542, opened 2022-11-29
Rename Preservation Master to Preservation File


tpendragon:
See https://pulibrary.github.io/dls-handbook/inclusive_language.html for why.

See https://github.com/samvera/valkyrie/pull/854 for upstream change

Changing this throughout all code will be a data migration.

hackartisan:
prod is done.

```
irb(main):007:0> query_service.custom_queries.find_by_property(model: FileSet, property: :file_metadata, value: { use: [Valkyrie::Vocab::PCDMUse.PreservationMasterFile] }, lazy: true).c
ount
=> 0
irb(main):008:0> query_service.custom_queries.find_by_property(model: FileSet, property: :file_metadata, value: { use: [Valkyrie::Vocab::PCDMUse.PreservationFile] }, lazy: true).count
=> 9128
```

PR #5613, opened 2023-01-10
Use new PreservationFile use value

hackartisan:
- Allow either preservation use value (until we complete a migration)
- Use new PreservationFile over deprecated term.
- Use "global" instead of "master"

advances #5542


eliotjordan (dismissed):
💯 Good job. Glad this is finally done.

eliotjordan (approved):

PR #5615, opened 2023-01-12
Add a migration to convert deprecated use value to PreservationFile

hackartisan:
There were 9127 of these in prod.

refs #5542


hackartisan:
Ran this in staging and it was successful.

tpendragon (approved):

PR #5616, opened 2023-01-12
Document Original/Preservation/IntermediateFile.

### LINKED ISSUE #5547: Document use case for Preservation Files vs Original Files vs IntermediateFiles.
I think this is just because we have some projects which have a non-watermarked Preservation file and a watermarked IntermediateFile.

tpendragon:
Closes #5547

### DIFF
--- a/app/derivative_services/audio_derivative_service.rb
+++ b/app/derivative_services/audio_derivative_service.rb
@@ -28,8 +28,10 @@ class AudioDerivativeService
     @resource ||= query_service.find_by(id: id)
   end
 
-  # Only ever checks primary_file, if this were to check preservation_file
-  # it would override derivatives for MediaResources. Take care!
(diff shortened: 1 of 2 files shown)


hackartisan (changes requested):
this is a helpful writeup.

I looked at some derivative services and they do check explicitly for the intermediate file first. I wonder if we need a change to the comments maybe on this method? https://github.com/pulibrary/figgy/blob/8e1604ffc64c0e92d6c6c5c796dd705efed5dc35/app/derivative_services/audio_derivative_service.rb#L31-L35

It might be nice to specify under the 2 use cases given, which one is considered the primary file.

How / where is primary file used besides generating derivatives? Say something about preservation?

tpendragon:
@hackartisan How's that look?

PR #5621, opened 2023-01-13
Rescue document generation exceptions in GeoBlacklightEventGenerator

### LINKED ISSUE #5551: GeoBlacklight Event generator shouldn't build a GeoBlacklight document during DeleteMemberJob
This creates an error when the resource is a scanned map because GBL document attempts to build a IIIF manifest link, and something needed to build the manifest has already been deleted. Either the document shouldn't get built or this method should rescue an exception and return false:

https://github.com/pulibrary/figgy/blob/7b61f1dd8e0e0fff661aa17bc867c3ae3cd6a56d/app/services/event_generator/geoblacklight_event_generator.rb#L84

Figgy/production] NoMethodError: undefined method `pyramidal_derivative' for nil:NilClass

## Backtrace

    line 777 of [PROJECT_ROOT]/app/services/manifest_builder.rb: base_url
    line 756 of [PROJECT_ROOT]/app/services/manifest_builder.rb: manifest_image_path
    line 52 of [PROJECT_ROOT]/app/services/geo_discovery/document_builder/iiif_builder.rb: iiif_path

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/90953229)

eliotjordan:
Closes #5551

### DIFF
--- a/app/services/event_generator/geoblacklight_event_generator.rb
+++ b/app/services/event_generator/geoblacklight_event_generator.rb
@@ -37,7 +37,7 @@ class EventGenerator
     def valid?(record)
       return false if record.is_a?(FileSet)
       return false unless record.try(:geo_resource?)
-      return false if errors?(record)
+      return false if schema_errors?(record)
 
       true
     end
@@ -80,9 +80,14 @@ class EventGenerator
         GeoDiscovery::DocumentBuilder::SlugBuilder.new(record).slug
       end
 
-      def errors?(record)
-        doc = document_generator(record).to_hash
-        return true unless doc[:error].nil?
+      def schema_errors?(record)
+        begin
+          doc = document_generator(record).to_hash
+          return true unless doc[:error].nil?
+        rescue
+          # the resource could not generate a document at all
+          return false
+        end
 
         false
       end
--- a/spec/services/event_generator/geoblacklight_event_generator_spec.rb
+++ b/spec/services/event_generator/geoblacklight_event_generator_spec.rb
@@ -156,5 +156,17 @@ RSpec.describe EventGenerator::GeoblacklightEventGenerator do
         expect(event_generator.valid?(record)).to be false
       end
     end
+
+    context "when a record generates an invalid geoblacklight document" do
+      let(:document_builder) { instance_double(GeoDiscovery::DocumentBuilder) }
+      before do
+        allow(GeoDiscovery::DocumentBuilder).to receive(:new).and_return(document_builder)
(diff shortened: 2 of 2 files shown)


hackartisan (changes requested):
I want to approve this because it clearly closes the ticket in the recommended way. But I don't really understand it and I think it's confusing that the test says "when the document is invalid it's valid".  Maybe the test can spell out the use case for this functionality more specifically? In the ticket it says "This creates an error when the resource is a scanned map because GBL document attempts to build a IIIF manifest link." (that doesn't make sense to me either; why wouldn't a scanned map have a manifest?) It also might be helpful to add a comment to `errors?` since it isn't concerned with some kinds of errors -- the kind you'd have to rescue from..

spec/services/event_generator/geoblacklight_event_generator_spec.rb
```diff
@@ -156,5 +156,17 @@
...
     end
+
+    context "when a record generates an invalid geoblacklight document" do
+      let(:document_builder) { instance_double(GeoDiscovery::DocumentBuilder) }
+      before do
+        allow(GeoDiscovery::DocumentBuilder).to receive(:new).and_return(document_builder)
+        allow(document_builder).to receive(:to_hash).and_raise(StandardError)
+      end
+
+      it "is still valid" do
```
hackartisan, line 167:
```suggestion
      it "is still emits a valid event" do
```

app/services/event_generator/geoblacklight_event_generator.rb
```diff
@@ -81,8 +81,12 @@ def slug(record)
       end
 
       def errors?(record)
-        doc = document_generator(record).to_hash
-        return true unless doc[:error].nil?
+        begin
+          doc = document_generator(record).to_hash
+          return true unless doc[:error].nil?
+        rescue
```
hackartisan, line 87:
```suggestion
        # the resource could not generate a document at all
        rescue
```

app/services/event_generator/geoblacklight_event_generator.rb
```diff
@@ -81,8 +81,12 @@ def slug(record)
       end
 
       def errors?(record)
```
hackartisan, line 83:
```suggestion
      def schema_errors?(record)
```

hackartisan (changes requested):
changes per discussion

eliotjordan:
@hackartisan Ready for re-review.

PR #5643, opened 2023-01-25
Quietly exit CloudFixityJob if resource no longer exists

### LINKED ISSUE #4078: Cloud fixity failures for items no longer in Figgy
Instead of erroring, we should check figgy to see if the item has been deleted and then... just carry on?

HB notice: https://app.honeybadger.io/projects/53391/faults/59168201

This honeybadger notice is gone, but to test this we can run the job on a non-existent ID.

## Success Criteria

Don't error if the object is gone.

eliotjordan:
Closes #4078


tpendragon (approved):

eliotjordan:
Pushed bad commit. :( Closing for now.

Issue #5642, opened 2023-01-24
Resources which had duplicate PreservationObjects deleted are no longer in preservation storage


tpendragon:
We have a list of FileSet IDs which had their duplicates deleted. We can re-preserve all of them.

ID File:

[resources.csv](https://github.com/pulibrary/figgy/files/10494247/resources.csv)

## Potential Strategies:

1. Delete the PreservationObject for all of those, and re-preserve them in a background job.
2. Manually call `Preserver#preserve_binary_content(force: true)` and `Preserver#preserve_metadata` for all of the above FileSets. We have to force it, because those POs already thing that they're preserved and it doesn't validate that against GCS.

eliotjordan:
PreservationJobs completed.

PR #5645, opened 2023-01-25
Add force_preservation parameter to Preserver and PreserveResourceJob

eliotjordan:
Works towards #5642

### DIFF
--- a/app/jobs/preserve_resource_job.rb
+++ b/app/jobs/preserve_resource_job.rb
@@ -6,7 +6,8 @@ class PreserveResourceJob < ApplicationJob
 
   # @param id [String] resource id
   # @param lock_tokens [Array<String>] serialized lock tokens
-  def perform(id:, lock_tokens: nil)
+  # @param force_preservation [Boolean] force binary node preservation
+  def perform(id:, lock_tokens: nil, force_preservation: false)
     resource = query_service.find_by(id: id)
     # If two jobs get queued before this job runs once (forcing future
     # PreserveResourceJobs to run inline), then without this it would create two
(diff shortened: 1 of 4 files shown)


app/jobs/preserve_resource_job.rb
```diff
@@ -6,7 +6,8 @@ class PreserveResourceJob < ApplicationJob
 
   # @param id [String] resource id
   # @param lock_tokens [Array<String>] serialized lock tokens
-  def perform(id:, lock_tokens: nil)
+  # @param force_preservation [Boolean] force binary node preservation
+  def perform(id:, lock_tokens: nil, force_preservation: nil)
```
tpendragon, line 10: Suggestion: Default to false instead of nil.

app/services/preserver.rb
```diff
@@ -2,28 +2,29 @@
 class Preserver
   # Provided as a mechanism for switching preservation strategies depending on
   # the profile. For now, there's only one, so fall down to new.
-  def self.for(change_set:, change_set_persister:, storage_adapter: nil)
+  def self.for(change_set:, change_set_persister:, storage_adapter: nil, force_preservation: nil)
```
tpendragon, line 5: Suggestion: Default to false instead of nil

app/services/preserver.rb
```diff
@@ -2,28 +2,29 @@
...
 
   class NullPreserver
     def self.preserve!; end
   end
 
-  attr_reader :change_set, :storage_adapter, :change_set_persister
+  attr_reader :change_set, :storage_adapter, :change_set_persister, :force_preservation
   delegate :resource, to: :change_set
-  def initialize(change_set:, storage_adapter: nil, change_set_persister:)
+  def initialize(change_set:, storage_adapter: nil, change_set_persister:, force_preservation: nil)
```
tpendragon, line 16: Suggestion: Default to false instead of nil, then you don't have to do the || below

tpendragon (changes requested):
Just a suggestion for defaulting to false for the param. In my experience, things can get messy the more we use nil as a stand-in for false.

app/jobs/preserve_resource_job.rb
eliotjordan: Good idea. 👍

tpendragon (approved):

PR #5646, opened 2023-01-25
Exit CloudFixityJob silently when resource is not found

### LINKED ISSUE #4078: Cloud fixity failures for items no longer in Figgy

eliotjordan:
Closes #4078

### DIFF
--- a/app/jobs/cloud_fixity_job.rb
+++ b/app/jobs/cloud_fixity_job.rb
@@ -9,6 +9,10 @@ class CloudFixityJob < ApplicationJob
     @resource_id = resource_id
     @child_property = child_property
     @child_id = child_id
+
+    # Do not create an event and honeybadger notification if the resource that
+    # was being checked no longer exists. This happens on occasion.
+    return unless resource_exist?
     event_change_set = EventChangeSet.new(Event.new)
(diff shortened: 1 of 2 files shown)


app/jobs/cloud_fixity_job.rb
```diff
@@ -9,6 +9,7 @@ def perform(status:, resource_id:, child_property:, child_id:)
     @resource_id = resource_id
     @child_property = child_property
     @child_id = child_id
+    return unless resource_exist?
```
tpendragon, line 12: Question: What's the reasoning behind an early return versus a rescue for the perform method?

tpendragon (changes requested):

app/jobs/cloud_fixity_job.rb
eliotjordan: The perform method is already pretty long. You'd have to shove the same into it as in the `resource_exist?` method.

eliotjordan:
@tpendragon Added comment.

tpendragon (dismissed):

PR #5651, opened 2023-01-27
Actively preserve un-preserved children.

### LINKED ISSUE #3463: Children aren't being preserved always on being marked complete
See https://figgy.princeton.edu/catalog/781a8874-941f-4c63-8a3b-5d7dbabf92f4 for an example. Probably some sort of race condition.

Screen Shot 2022-11-17 at 12 07 38 PM

Our implementation shows "in progress" even if it's never going to happen.

tpendragon:
When preserving this now checks for children which don't have PreservationObjects and queues them up, rather than relying on it being the first time the parent was preserved.

This should fix any race conditions where that first attempt is missed, and will allow resources to preserve children on subsequent saves.

Closes #3463

### DIFF
--- a/app/queries/find_never_preserved_child_ids.rb
+++ b/app/queries/find_never_preserved_child_ids.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+class FindNeverPreservedChildIds
+  def self.queries
+    [:find_never_preserved_child_ids]
+  end
+
+  attr_reader :query_service
+  delegate :orm_class, to: :resource_factory
+  delegate :adapter, to: :query_service
+  delegate :connection, to: :adapter
+  delegate :resource_factory, to: :query_service
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def find_never_preserved_child_ids(resource:)
+    # Get IDs of resources that have preservation objects.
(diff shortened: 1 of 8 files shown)


app/queries/find_never_preserved_child_ids.rb
```diff
@@ -0,0 +1,32 @@
...
+  delegate :resource_factory, to: :query_service
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def find_never_preserved_child_ids(resource:)
+    # Get IDs of resources that have preservation objects.
+    file_set_ids = adapter.connection[preservation_object_query, id: resource.id.to_s].to_a.map { |x| x[:file_set_id] }
+    # Subtract those from member_ids
+    resource.member_ids.select do |member_id|
```
eliotjordan, line 20: suggestion: might be clearer to use `reject` rather than `select` with negative `include?`

eliotjordan (approved):
One suggestion, otherwise this this looks good.

hackartisan (changes requested):
👍🏻 I agree with @eliotjordan; blocking for his change request.

eliotjordan (approved):

PR #5670, opened 2023-02-09
Remove rights-statement class from all over the place

### LINKED ISSUE #5188: Rights note cannot be edited for new geo resources
When you select a rights statement you should be able to edit to add a note.
- Rights note is always read-only when adding new map, raster, or vector resource
  - Does work for Scanned Resource

hackartisan:
Note to reviewer: We didn't see this class used anywhere else, and it was interfering with our local javascript. Is there somewhere we should check that we may have overlooked? A double-check would be appreciated.

It's used in the javascript that activates the rights statement note,
but it was present in a bunch of fields on the edit form. No evidence of
further use in other javascript or in styling. We also checked upstream.

fixes #5188


tpendragon (approved):
Yup, looks like this was just left from copy and paste. Good work!

PR #5671, opened 2023-02-09
Index and display DeletionMarkers

### LINKED ISSUE #3350: I should be able to list and search for resource tombstones

eliotjordan:
Closes #3350
Screenshot 2023-02-09 at 12 02 14 PM

---

Screenshot 2023-02-09 at 12 02 23 PM

---

Screenshot 2023-02-09 at 12 02 32 PM

### DIFF
--- a/app/adapters/indexing_adapter.rb
+++ b/app/adapters/indexing_adapter.rb
@@ -10,8 +10,7 @@ class IndexingAdapter
   def self.no_index_models
     [
       Event,
-      PreservationObject,
-      DeletionMarker
+      PreservationObject
     ]
   end
   attr_reader :metadata_adapter, :index_adapter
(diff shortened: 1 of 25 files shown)


tpendragon:
@eliotjordan I was thinking, maybe what we can do is add the deleted object to an embedded property on the DeletionMarker when we make it? That way we can adjust the indexing rules later and have the data to pull from.

eliotjordan:
PR Updated:

- Added identifiers to indexed fields
- "Delete Deletion Marker" button is hidden in production
- A nested copy of the deleted object is saved in the marker

tpendragon (approved):

PR #5675, opened 2023-02-13
Numismatic accessions date validation

### LINKED ISSUE #5187: Numismatic Accessions should validate dates
- allows STRING in the DATE field when editing:

taylor-yamashita:
Closes #5187

### DIFF
--- a/app/decorators/numismatics/accession_decorator.rb
+++ b/app/decorators/numismatics/accession_decorator.rb
@@ -75,9 +75,9 @@ module Numismatics
     end
 
     def parsed_date
-      if date.to_s.include?("/")
+      if date.to_s.include?("-")
         # It's an american hand-entered date, parse it as such
-        Time.zone.strptime(date, "%m/%d/%Y")
+        Time.zone.strptime(date, "%Y-%m-%d")
       else
         Time.zone.parse(date.to_s)
       end
--- a/app/decorators/numismatics/accession_decorator.rb
+++ b/app/decorators/numismatics/accession_decorator.rb
@@ -77,7 +77,7 @@ module Numismatics
     def parsed_date
       if date.to_s.include?("-")
         # It's an american hand-entered date, parse it as such
-        Time.zone.strptime(date, "%Y-%m-%d")
+        Time.zone.strptime(date.to_s, "%Y-%m-%d")
       else
         Time.zone.parse(date.to_s)
       end
--- a/app/validators/accession_date_validator.rb
+++ b/app/validators/accession_date_validator.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+class AccessionDateValidator < ::ActiveModel::Validator
+  def validate(record)
+    return if record.date.blank?
+    return if record.date.is_a?(DateTime)
+    record.errors.add(:date, "Dates should be DateTime objects")
+  end
+end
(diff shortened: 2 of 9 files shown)


taylor-yamashita:
The validator I created currently says "Dates should be DateTime objects" when given an invalid date input, but what kind of error message would be most helpful?

app/validators/accession_date_validator.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+class AccessionDateValidator < ::ActiveModel::Validator
+  def validate(record)
+    return if record.date.blank?
+    return if record.date.is_a?(DateTime)
+    record.errors.add(:date, "Dates should be DateTime objects")
```
hackartisan, line 6: Probably something like "Please enter dates as MM/DD/YYYY". once coersion is working it will accept a variety of inputs, but we probably just need to choose one to recommend. "YYYY/MM/DD" would be another option. I don't know whether figgy has an opinion or recommendation on this anywhere else.

taylor-yamashita: Changed the error message to the first suggestion.

hackartisan:
@taylor-yamashita it looks like they've accounted for this issue in reform... I'm not sure exactly how it all hooks up. Maybe it's this coercion require? https://trailblazer.to/2.1/docs/reform.html#reform-options-coercion

taylor-yamashita:
> @taylor-yamashita it looks like they've accounted for this issue in reform... I'm not sure exactly how it all hooks up. Maybe it's this coercion require? https://trailblazer.to/2.1/docs/reform.html#reform-options-coercion

I did try this approach (it uses Types::Form::DateTime instead of Types::Params::DateTime) but wasn't able to resolve the issue.

taylor-yamashita:
Update: Worked with @eliotjordan and decided that coercion wasn't working. Reworked the validator to validate with `DateTime.strptime()` and `rescue Date::Error` instead of coercion. (With the coercion approach, invalid dates would get caught up in a changeset error before any of the validation could happen.)

Note: the existing numismatics accessions entries all have differently formatted dates; this change will only affect newly created entries and edits to old ones.

taylor-yamashita:
This will require a followup date cleanup migration to standardize all existing numismatic accessions date formats.

app/decorators/numismatics/accession_decorator.rb
```diff
@@ -75,9 +75,9 @@ def formatted_date
     end
 
     def parsed_date
-      if date.to_s.include?("/")
+      if date.to_s.include?("-")
         # It's an american hand-entered date, parse it as such
-        Time.zone.strptime(date, "%m/%d/%Y")
+        Time.zone.strptime(date.to_s, "%Y-%m-%d")
```
taylor-yamashita, line 80: Without adding .to_s, it throws "TypeError: no implicit conversion of DateTime into String" and does not actually use the specified format in parsed date. (This previously caused a test to fail by putting the original date with the time attached in formatted output.)

eliotjordan (changes requested):
We might need to do a migration, first, before merging and deploying this PR. In existing accession records with UTC date/time values, the edit form raises:

```
TypeError: no implicit conversion of DateTime into String
```

https://github.com/pulibrary/figgy/blob/bebfd1948407bf567449604ef1c9ae8a1972b3c4/app/validators/accession_date_validator.rb#L5

eliotjordan:
@taylor-yamashita Created ticket for migration. https://github.com/pulibrary/figgy/issues/5699

eliotjordan (approved):

Issue #5631, opened 2023-01-19
Generate report for any and all Graphic Arts items in Figgy


kelea99:
### Summary
Request by new Graphic Arts curator, Molly D.  Interested in any and all Graphic Arts materials in figgy. Assuming this will also be by location prefix but will confirm what all of those are.

This is a cataloging and catalog cleanup project.

### Impact 
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._

### Priority recommendation
Perhaps run at same time as #5630 ?
- [ ] asap
- [x] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "Maintenance/Research" labels_

There's postings for a job that is going out, and Kim needs to hand this report over by Friday.

hackartisan:
@kelea99 We still need more information about the requirements here. It would be helpful to hear the full user story that's motivating this request.

Is there another way to get these records besides location code? Do we think there are records that are part of graphics are but aren't in the graphic arts collection?

And what data is needed about them?

kelea99:
@hackartisan I can do a search in figgy to confirm that all items / most items are in the GA collection in figgy

kelea99:
Requested Collections/Searches to run: 

- https://figgy.princeton.edu/?f%5Bmember_of_collection_titles_ssim%5D%5B%5D=Treasures+of+the+Graphic+Arts+Collection
- GA: https://figgy.princeton.edu/catalog?q=%22GA%22
- GAX/Patron Requests: https://figgy.princeton.edu/?f%5Bmember_of_collection_titles_ssim%5D%5B%5D=Patron+Requests&q=%22GAX%22

kelea99:
_User story: As a Graphic Arts Curator that is new to PUL, I would like a report generated of any/all GA/GAX items in the digital repository. I am in the process of preparing for a large 3-6 year cataloging and digitization project for my materials and I would like to get a better idea:
1. How many items have been ingested in figgy from my collections
2. How many images (total) have been ingested in figgy from my collections
3. A CSV download of GA/GAX/"Treasures of the Graphic Arts Collection" items, containing:  The title, Source Metadata Identifier, Ark, Visibility, state, and call number
I will use this information to determine what collections should be targeted for digitization, metadata cleanup, or metadata creation._

hackartisan:
Note that GA and GAX are partial location code values.

[...]

PR #5684, opened 2023-02-21
Create CSV Report Generator.

tpendragon:
Work towards #5631, might be useful for a more generic report generator.

### DIFF
--- a/app/values/csv_report.rb
+++ b/app/values/csv_report.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+class CSVReport
+  attr_reader :resources, :fields
+  def initialize(resources, fields: [:title])
+    @resources = resources
+    @fields = fields
+  end
+
+  def csv_rows
+    @csv_rows ||= resources.lazy.map do |resource|
+      Row.new(resource)
+    end
+  end
+
+  def write(path:)
+    CSV.open(path, "w") do |csv|
+      csv << headers
+      csv_rows.each do |row|
+        csv << row.to_h.values_at(*fields)
+      end
+    end
+  end
+
+  def headers
+    fields.map(&:to_s).map(&:humanize)
+  end
+
+  class Row
+    attr_reader :resource
+    def initialize(resource)
+      @resource = resource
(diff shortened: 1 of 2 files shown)


app/values/csv_report.rb
```diff
@@ -0,0 +1,65 @@
...
+    attr_reader :resource
+    def initialize(resource)
+      @resource = resource
+    end
+
+    def to_h
+      Hash[
+        # __attributes__ is much faster and doesn't include nil values.
+        resource.__attributes__.merge(attributes_without_reserved(imported_metadata)).map do |key, values|
+          values = Array.wrap(values).map(&:to_s).join(", ")
```
hackartisan, line 45: Is this separating values with a comma? Can we do that with a CSV? Maybe add a field to the test with multiple values.

hackartisan (changes requested):
Yes, this looks really handy! I'd be interested in seeing how a refactor of https://github.com/pulibrary/figgy/blob/fcd0e83eaa77e27c7a78038a4a94ea35de0761a6/app/controllers/reports_controller.rb would look. I'm undecided on blocking for that because I worry it would get lost if we don't do it now and I don't want equivalent code everywhere, but I recognize that this ticket is asap. What do you think @tpendragon?

hackartisan:
And this is really nicely commented thank you for that

tpendragon:
> I'm undecided on blocking for that because I worry it would get lost if we don't do it now and I don't want equivalent code everywhere, but I recognize that this ticket is asap. What do you think @tpendragon?

I looked at that too. I didn't tackle it because the reality is I'm out of time - I don't think I'll be able to do that this week.

hackartisan:
Yeah, that makes sense. It might be an interesting refactor for a student to work on and we have room on that board for sure.

tpendragon:
@hackartisan I added the extent field, which has a comma in it. Work?

hackartisan (approved):

hackartisan:
i created https://github.com/pulibrary/figgy/issues/5687

PR #5693, opened 2023-02-24
Delete PDF derivative files when a PDF is deleted.

### LINKED ISSUE #4161: Should pdf derivative file sets be deleted when a pdf file set is deleted?
The pdf splitting functionality creates intermediate tiff files. These are cleaned up and re-created when derivatives are regenerated, but currently aren't deleted when the pdf file set is deleted.

m-ichelleliu:
Resolves #4161.

### DIFF
--- a/spec/change_set_persisters/change_set_persister/cleanup_pdf_derivatives_spec.rb
+++ b/spec/change_set_persisters/change_set_persister/cleanup_pdf_derivatives_spec.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe ChangeSetPersister::CleanupPDFDerivatives do
+  subject(:hook) { described_class.new(change_set_persister: change_set_persister, change_set: change_set) }
+  let(:file) { fixture_file_upload("files/sample.pdf", "application/pdf") }
+  let(:adapter) { Valkyrie::MetadataAdapter.find(:indexing_persister) }
+  let(:storage_adapter) { Valkyrie.config.storage_adapter }
+  let(:change_set_persister) { ChangeSetPersister.new(metadata_adapter: adapter, storage_adapter: storage_adapter) }
+  let(:change_set) { ChangeSet.for(fileset) }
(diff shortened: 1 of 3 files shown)


spec/change_set_persisters/change_set_persister/cleanup_pdf_derivatives_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  describe "#run" do
+    with_queue_adapter :inline
+    context "with a PDF fileset" do
+      it "deletes the derivatives" do
+        scanned_resource
+        hook.run
+        expect(Wayfinder.for(scanned_resource).members.count).to eq(1)
+      end
+    end
+
```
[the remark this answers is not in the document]
hackartisan: If you want, for this new test and for the one that tests "doesn't run" we could mock the change set persister we pass in and test that `.save` isn't called.

hackartisan (changes requested):
Yay! There are just a few cleanup ideas I have, let's look together.

spec/change_set_persisters/change_set_persister/cleanup_pdf_derivatives_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  describe "#run" do
+    context "with a PDF fileset" do
+      it "deletes the derivatives" do
+        scanned_resource
+        allow(change_set_persister).to receive(:save).and_call_original
+        hook.run
+        expect(change_set_persister).to have_received(:save).at_least(:once)
+        expect(Wayfinder.for(scanned_resource).members.count).to eq(1)
```
eliotjordan, line 26: question: The remaining member is the to-be-deleted deleted FileSet (because this a `before_delete` action)?

eliotjordan (approved):
Excellent work.

spec/change_set_persisters/change_set_persister/cleanup_pdf_derivatives_spec.rb
hackartisan: That's right.

PR #5695, opened 2023-02-28
Remove resource from all collections when it's made a child

### LINKED ISSUE #5678: Remove Collections from any/all child records when child is moved to a parent record in figgy.
### Summary
If collections are not removed from scanned resources that become title-only children of a parent record, those children are pulled in as individual scanned resources to DPUL exhibits. Which is totes bad. 

### Impact 
This is not hindering or stopping work. Rather, it is making DPUL exhibits messy and confusing (like Lapidus).
### Priority recommendation

- [x] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

This is a "make developers lives easier" ticket.  Collections  cannot be removed from child records by figgy users. developers are required for that task.  there is likely a lot of cleanup needed for child records that were added to parents after ingest as scanned resources. 

We're making this ASAP because if this happens post-cleanup we've recently done then it'll undo the time we spent to do that cleanup.

hackartisan:
I wanted to add this logic into the existing `AppendToParent` class, but that's used as a post save hook and this needed to be pre-save. I'm not in love with the way this mirrors the logic in that class, and how the two may need to be changed together. Open to other ideas / approaches.

closes #5678

### DIFF
--- a/app/change_set_persisters/change_set_persister.rb
+++ b/app/change_set_persisters/change_set_persister.rb
@@ -37,9 +37,11 @@ class ChangeSetPersister
         CacheParentId
       ],
       after_save: [
-        AppendToParent,
         UpdateAuthToken,
         GenerateMosaic
+      ],
+      around_save: [
+        AppendToParent
       ],
       after_save_commit: [
         PublishMessage::Factory.new(operation: :update),
@@ -195,6 +197,9 @@ class ChangeSetPersister
     private
 
       def before_save(change_set:)
+        registered_handlers.fetch(:around_save, []).each do |handler|
+          handler.new(change_set_persister: self, change_set: change_set, post_save_resource: nil).run_before_save
+        end
         registered_handlers.fetch(:before_save, []).each do |handler|
           handler.new(change_set_persister: self, change_set: change_set).run
         end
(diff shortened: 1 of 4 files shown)


tpendragon:
Oh interesting, so to put this in one place we'd almost need an around hook or hooks that can have both a pre-save and a post-save method?

hackartisan:
@tpendragon do you think it's appropriate to move it all to pre-save?

Yeah the wrap around idea is interesting..

tpendragon:
I don't think you can append to parent pre-save because you don't have an ID

hackartisan:
aha, for a new resource, that makes sense.

kelea99:
> I don't think you can append to parent pre-save because you don't have an ID

This could perhaps be solved with a workflow change. saving the parent record first before appending the child/children to the parent record? Then an id would be there. 
the soon-to-be child records that are being appended to an existing parent (with ID) are taken care of already.  Does this help at all?

hackartisan:
@kelea99 the scenario where there isn't an id is when the child is new, e.g. the "attach a scanned resource" button on a resource show page, which takes you to a new resource form.

hackartisan:
@tpendragon what do you think of this?

tpendragon:
Looks great to me!

tpendragon (approved):

tpendragon (approved):

PR #5698, opened 2023-03-02
Run regular local fixity checks via cron

### LINKED ISSUE #5554: Run regular local fixity checks via cron
## Success Criteria

Randomly sample 100% of the repository every year for local fixity checks. See if our background workers can keep up with that, and adjust as necessary.

eliotjordan:
Closes #5554


hackartisan (approved):
Looks good!

PR #5717, opened 2023-03-23
Remove find_cloud_fixity_failures query

### LINKED ISSUE #5667: Remove find_cloud_fixity_failures query
It's now redundant with `find_fixity_events` and no longer in use, since #5595.

Move its specs into the `find_fixity_events` spec if they don't duplicate the specs that are in there.

m-ichelleliu:
Resolves #5667


spec/services/migrations/event_current_migrator_spec.rb
hackartisan, line 1: @eliotjordan should we remove this migrator also?

spec/queries/find_fixity_events_spec.rb
```diff
@@ -60,6 +60,25 @@
       end
     end
 
+    context "when a fixity check which has failed later succeeds" do
+      # Old one uses preservation objects and real resources, this one doesn't; not sure if necessary
+      let(:event) { FactoryBot.create_for_repository(:cloud_fixity_event, status:  "FAILURE", resource_id: event_resource_id, current: true) }
+      let(:event2) { FactoryBot.create_for_repository(:cloud_fixity_event, status: "FAILURE", resource_id: event2_resource_id) }
+      let(:event3_resource_id) { Valkyrie::ID.new(SecureRandom.uuid) }
+      let(:event3) { FactoryBot.create_for_repository(:cloud_fixity_event, status: "SUCCESS", resource_id: event3_resource_id, current: true) }
```
hackartisan, line 68: It looks like for this test we want 2 events with the same resource id, where the first failed and the second succeeded. and then we want a 3rd event with a different resource id and a failed status. the query should return only the 3rd event, since the resource connected to the first 2 events has now had a successful fixity check.

hackartisan (changes requested):

spec/queries/find_fixity_events_spec.rb
m-ichelleliu: Oops, didn't notice that the first two events used the same id. Just changed it!

spec/services/migrations/event_current_migrator_spec.rb
eliotjordan: Yeah. I think we're done with it.

hackartisan (changes requested):
The spec looks good! I double-checked that it fails when certain parts of the query are removed. I think from here we can either add a commit that removes `spec/services/migrations/event_current_migrator_spec.rb` and `app/services/migrations/event_current_migrator.rb` or revert deletion of the comment on that spec and ticket the removal of those files.

eliotjordan (approved):

PR #5730, opened 2023-03-30
Replace Browse Everything

tpendragon:
This does a few things:

1. Creates a new FileBrowser vue component with sub-components, and a little controller for returning directory info.
2. Removes BrowseEverything gem and javascript.
3. Re-adds background server upload appending, so really large files or thousands of files can be initiated from the UI without a lot of waiting.

Closes #5318, closes #5733

PR #5732, opened 2023-03-31
Figgy Ruby 3.1 Compatibility

tpendragon:
Depends on #5730


tpendragon:
I tested this in staging without upgrading Ruby and it worked fine, should make it pretty easy to upgrade.

hackartisan (approved):

PR #5739, opened 2023-04-06
Fix EZID type mismatch error.

### LINKED ISSUE #5350: EZID is checking ARKs that point to restricted content on Figgy
Raising exceptions.

[Figgy/production] ActionController::RespondToMismatchError: respond_to was called multiple times and matched with conflicting formats in this action. Please note that you may only call respond_to and match on a single format per action.

## How to Fix

These requests should be returning a 401 Unauthorized, not erroring.

## Backtrace

    line 33 of [PROJECT_ROOT]/app/controllers/application_controller.rb: deny_resource_access

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/87113178)

tpendragon:
When an HTTP ACCEPT header is sent which is a wildcard, Rails picks the first response in a `respond_to` block. Blacklight runs first, which has HTML first, and then it'd hit our catch block and do JSON first, which would then error because Rails tries to respond with both HTML and JSON.

I couldn't get the spec to replicate this behavior (something's up with the testing harness), but I was able to
make sure that this method does the HTML behavior when given a wildcard header.

Closes #5350

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,20 +30,24 @@ class ApplicationController < ActionController::Base
 
   rescue_from Blacklight::AccessControls::AccessDenied, CanCan::AccessDenied, with: :deny_resource_access
   def deny_resource_access(exception)
+    # When HTTP_ACCEPT is set to *, it picks the first format. If Blacklight is
+    # HTTP first we're JSON first, it errors, so make sure these are in the same
+    # order as Blacklight.
+    # See figgy#5350 for what happens if they aren't.
     respond_to do |format|
-      format.json do
-        if exception.action == :manifest && @resource.present?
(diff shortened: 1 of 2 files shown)


hackartisan (commented):
So you added the test, but it never fails? Should we note that on the test? It seems like it's giving us a false sense of security.

tpendragon:
> So you added the test, but it never fails? Should we note that on the test? It seems like it's giving us a false sense of security.

It failed before I switched the HTML/JSON, because it was going through the JSON logic (which renders an unauthorized)

hackartisan:
okay! 👍🏻

hackartisan (approved):

hackartisan:
excited to see this bug squashed!

PR #5740, opened 2023-04-06
PDF Coversheet should inherit some parent metadata for MVWs.

### LINKED ISSUE #5661: Update PDF coversheet for a title-only ingest child record to use the parent metadata + title of child.
### Summary

When I download a PDF of a Volume from a MVW , I expect that coverpage to reflect the same metadata as a parent record PDF download coverpage, plus the title of the child record being downloaded. I would also expect the citation information to be updated to reflect the preferred citation and the title only title from the child (see PDF example below).

[cover_page_generator.rb
](https://github.com/pulibrary/figgy/blob/c5f851aea6e7c65306e17fb40ec9cf74c40a0a7f/app/services/pdf_generator/cover_page_generator.rb#L79)
Example of Ideal Child Record Coverpage (manually created by @escowles ) - see below.
[vol1.pdf](https://github.com/pulibrary/figgy/files/10582716/vol1.pdf)

### Impact 
This is not impeding work, nor is there a hard deadline. It is just not the awesomeness it could be. 
### Priority recommendation

- [ ] asap
- [x] within the next 3 weeks
- [ ] PO will prioritize

Okayed by @tpendragon 
[Added by @tpendragon] Much of our content is MVWs and the only way they can get a PDF is per-volume. The cover page is meant to remind them of what they're looking at, and right now they get no pointer, which could stop researchers from being able to organize their content.

### Expected behaviour
When I go to a volume in a MVW work within a viewer and download a PDF for a specific volume, I expect to have a coversheet at the beginning of that PDF that includes the metadata that is seen in a single scanned resource attached to an identifier/the parent record and  the title of the child after the parent title. The title of the child will also be present in the Citation information. 

### Actual behaviour
When I go to a volume in a MVW work within a viewer and download a PDF for a specific volume, I only get the title of the child record and no metadata from the parent.

### Steps to reproduce behaviour
Go to a MVW
Try it out.

tpendragon:
Closes #5661

### DIFF
--- a/app/services/pdf_generator/cover_page_generator.rb
+++ b/app/services/pdf_generator/cover_page_generator.rb
@@ -47,7 +47,7 @@ class PDFGenerator
         prawn_document.image Rails.root.join("app", "assets", "images", "pul_logo_long.png").to_s, position: :center, width: Canvas::LETTER_WIDTH - 30
         prawn_document.stroke_color "000000"
         prawn_document.move_down(20)
-        header(prawn_document, resource.title, size: 24)
+        header(prawn_document, pdf_metadata[:title], size: 24)
         resource.rights_statement.each do |statement|
           text(prawn_document, rights_statement_label(statement))
           rights_statement_text(statement).split(/\n/).flat_map { |x| x.split("<br/>") }.each do |line|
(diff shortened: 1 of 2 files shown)


app/services/pdf_generator/cover_page_generator.rb
```diff
@@ -103,6 +90,43 @@ def apply(prawn_document)
...
+
+      # Returns a field value by asking both the resource and the
+      # parent, returning the first time it finds one.
+      def imported_field_value(field)
+        resource.try(field) || resource.try(:primary_imported_metadata)&.[](field) || parent_resource&.try(field) || parent_resource&.try(:primary_imported_metadata)&.[](field)
+      end
+
+      # Returns the field values by using send() against both the parent and the
+      # resource and combining them.
+      def parented_field_value(field)
```
eliotjordan, line 112: 😆 "parented"

spec/services/pdf_generator_spec.rb
```diff
@@ -301,6 +301,35 @@
       end
     end
 
+    context "when a resource is a volume in a MVW" do
+      let(:resource) { FactoryBot.create_for_repository(:scanned_resource, files: [file], pdf_type: ["color"], source_metadata_identifier: "991234563506421") }
```
eliotjordan, line 305: question: Do we need this if we're defining resource on line 308?

eliotjordan (changes requested):
Looks good except for rubocop and question about let statement.

tpendragon:
Fixed - thanks

eliotjordan (approved):

PR #5742, opened 2023-04-07
Fix ocr upload_file test

eliotjordan:
~~Controller tests don't like decorated `Rack::Test::UploadedFile` objects.~~

Updated to rename the spec support method causing the exception so we're not overriding.

### DIFF
--- a/spec/characterization_services/geo_characterization_service_spec.rb
+++ b/spec/characterization_services/geo_characterization_service_spec.rb
@@ -60,7 +60,7 @@ RSpec.describe GeoCharacterizationService do
   context "when a file set contains a preservation file and an intermediate file" do
     let(:tika_output) { tika_shapefile_output }
     it "characterizes both files" do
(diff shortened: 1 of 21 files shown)


tpendragon (changes requested):
Question: If Rack::Test::UploadedFile matches the API of what comes through when I use a file upload, why are we overriding what `fixture_file_upload` returns to add to that API? Were we using `fixture_file_upload` inappropriately? Should we have some other method that returns a decorated file?

eliotjordan:
Because there is limited support for Intermediate files in Figgy and they're hard to work with in tests. Good idea to change the method name.

eliotjordan:
@tpendragon Updated.

tpendragon (approved):

tpendragon:
@eliotjordan Still failing CI for some reason.

eliotjordan:
@tpendragon Missed a few includes. Ready now.

tpendragon (approved):

PR #5755, opened 2023-04-18
Allow FileAppender to fail for derivatives.

### LINKED ISSUE #5078: File appender failure notices are too frequent
https://app.honeybadger.io/projects/53391/faults/84134442

It takes a lot of time to triage this notice
* copy the resource id
* navigate to it in figgy
* if it's a resource, click the file manager and browse through children clicking "regenerate" button as needed
* if it's a file set (usual case), click through to parent and its file manager
* if it's a file set with no parent link, go into prod console to see if it has a parent, potentially delete the file set

It may be sufficient to update the notice with more info (like a link to the file appender page of the parent if it's a file set -- note that sometimes it's a resource). Or maybe we need to think about whether we can find some circumstances under which we can retry.

The underlying errors appear to vary. examples:
* NetTimeout
* SSL error
* "Resource temporarily unavailable @ rb_sysopen"
* multipart upload failed: Your socket connection to the server was not read from or written to within the timeout period.

Let's use this space at least for now to record which underlying error we saw and whether we found a resource with derivatives that failed to generate, a resource that had been deleted, or a resource that looked fine (these cases may have been resolved by the ingester already). Maybe then we can identify some cases where it makes sense to do a retry.

## Fix Proposal

This Honeybadger notice and rescue in https://github.com/pulibrary/figgy/blob/792ac531a5c59187578073a6d6c37af6e60d458d/app/services/file_appender.rb#L126-L131 was added in response to flakiness ingesting from Google Drive. We don't ingest from Google Drive anymore, and most of the errors that happen here are fixed by a manual retry from the developers. If we remove that block of code, Figgy will retry by itself, meaning most times we won't have to do anything.

Let's remove it.

tpendragon:
Closes #5078

### DIFF
--- a/app/services/file_appender.rb
+++ b/app/services/file_appender.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 # Creates file sets to be added to resources via a changeset
 class FileAppender
+  class FileUploadFailed < StandardError; end
   attr_reader :files, :parent
   attr_reader :change_set_persister
   delegate :storage_adapter, to: :change_set_persister
@@ -125,6 +126,11 @@ class FileAppender
       node
     rescue StandardError => error
       message = "#{self.class}: Failed to append the new file #{original_filename} for #{node.id} to resource #{parent.id}: #{error}"
+      # For FileSet parents we're likely attaching a derivative, so let the
+      # error fall through so sidekiq restarts.
+      raise FileUploadFailed, message if file_set?(parent)
+      # Prevent error fallthrough for resource file attachments - otherwise if
+      # this keeps retrying we'll end up with a bunch of orphan files on disk.
       Valkyrie.logger.error message
       Honeybadger.notify message
       nil
--- a/app/services/file_appender.rb
+++ b/app/services/file_appender.rb
@@ -127,10 +127,12 @@ class FileAppender
     rescue StandardError => error
       message = "#{self.class}: Failed to append the new file #{original_filename} for #{node.id} to resource #{parent.id}: #{error}"
       # For FileSet parents we're likely attaching a derivative, so let the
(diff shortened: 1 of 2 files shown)


app/services/file_appender.rb
```diff
@@ -125,6 +126,11 @@ def create_node(file)
       node
     rescue StandardError => error
       message = "#{self.class}: Failed to append the new file #{original_filename} for #{node.id} to resource #{parent.id}: #{error}"
+      # For FileSet parents we're likely attaching a derivative, so let the
+      # error fall through so sidekiq restarts.
```
hackartisan, line 130: suggestion: use "raise the error" instead of "let the error fall through"

app/services/file_appender.rb
```diff
@@ -125,6 +126,11 @@ def create_node(file)
...
+      raise FileUploadFailed, message if file_set?(parent)
+      # Prevent error fallthrough for resource file attachments - otherwise if
+      # this keeps retrying we'll end up with a bunch of orphan files on disk.
```
hackartisan, line 133: Suggestion: Add a little more here about the use case that results in the orphan files. Specifically it's because of the files that were ingested in the loop before we hit this one that's failing.

Suggestion: Use "do not raise" instead of "prevent error fallthrough"

hackartisan (changes requested):
I'm excited for this update I think it will really help! Some suggestions on comments thank you

app/services/file_appender.rb
tpendragon: Done

app/services/file_appender.rb
tpendragon: Done - take a look?

eliotjordan (approved):

PR #5756, opened 2023-04-18
Add Complete When Processed to Bulk Ingest

### LINKED ISSUE #5566: Remove ability to bulk ingest directly to "complete" state
This causes a variety of issues with derivative generation and preservation. Common with geo material.

### LINKED ISSUE #5579: Add "Complete (When Processed)" as a state to Bulk Ingest
This state will automatically mark the resource complete when all the derivatives are generated.

## Implementation Details

Ideas:

1. Add a new state (processing_for_complete maybe?)
2. Use Sidekiq batches or something else to know when everything's generated derivatives and automatically mark it as complete when that's done.
3. Remove "Complete" as an option from Bulk Ingest (ticketed separately as #5566)

## Related Tickets

#5566 
#3477

tpendragon:
This adds the "complete_when_processed" state which can't be advanced to, but can be selected in bulk ingest.

It also adds an auto-completer - every 30 minutes every resource which is marked `complete_when_processed` will be marked complete if it has members and all of its members, deep, are processed.

One downside of the auto-completer is that for bulk-ingested MVWs it marks all the volumes as "complete_when_processed", but when the MVW is marked complete it marks it for the children - so we get stale object errors for the volumes: https://app.honeybadger.io/projects/53391/faults/95377090. Maybe that's okay? Maybe we want to ignore stale object error exceptions and not notify?

Closes #5579, closes #5566

Screen Shot 2023-04-18 at 3 32 43 PM

Screen Shot 2023-04-18 at 3 36 01 PM

### DIFF
--- a/architecture-decisions/0010-resource-autocomplete.md
+++ b/architecture-decisions/0010-resource-autocomplete.md
@@ -1,0 +1,48 @@
+# 10. Resource Auto Completion
+
+Date: 2023-04-18
+
+## Status
+
+Accepted
+
+## Context
+
+Our users often don't have time to go back to a resource they've ingested and
+check to see if it's okay after the initial ingest, and so tend to immediately
+mark a resource complete at ingest time. Unfortunately this results in a lot of
+churn in our preservation infrastructure - every file that gets added during
+that ingest forces a re-preservation of the parent's metadata.
+
+To prevent that, and also make sure that resources are being completed that are
+useful to our patrons, we want to add a new state that will automatically mark a
+resource complete when all of its files are attached and the derivatives
+generated.
+
+This auto-completion could be handled in one of two ways - either via a cron
+job that runs on some regular interval or by utilizing [Sidekiq
+Batches](https://github.com/sidekiq/sidekiq/wiki/Batches).
+
+The cron job is not as immediate, but is easy to maintain and reason through.
+The batches allow us to track the lifecycle of the object through jobs and will
+enable early-as-possible completion, but is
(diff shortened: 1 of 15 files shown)


architecture-decisions/0010-resource-autocomplete.md
```diff
@@ -0,0 +1,48 @@
...
+useful to our patrons, we want to add a new state that will automatically mark a
+resource complete when all of its files are attached and the derivatives
+generated.
+
+This auto-completion could be handled in one of two ways - either via a cron
+job that runs on some regular interval or by utilizing [Sidekiq
+Batches](https://github.com/sidekiq/sidekiq/wiki/Batches).
+
+The cron job is not as immediate, but is easy to maintain and reason through.
+The batches allow us to track the lifecycle of the object through jobs and will
```
hackartisan, line 27: suggestion: "would enable" instead of "will enable" -- since we're not going to do it that way

architecture-decisions/0010-resource-autocomplete.md
```diff
@@ -0,0 +1,48 @@
...
+enable early-as-possible completion, but is
```
hackartisan, line 28: suggestion: "but would be" instead of "but is"

hackartisan (changes requested):
I like how straightforward it is to add this to figgy.

One thing that is nice about a job is that it logs when it starts and ends. I wonder if it would be nice to add that in here, so we can go look at the logs and see how long it takes.

hackartisan (approved):

PR #5810, opened 2023-05-15
Update universalviewer from 3 to 4, use package via yarn

### LINKED ISSUE #5778: Update to universalviewer v4
blocks https://github.com/pulibrary/figgy/issues/5376

this is the closest thing there is to an upgrade guide: https://github.com/UniversalViewer/universalviewer/wiki/UV-Examples

hackartisan:
Need to test custom use cases.

closes #5778


hackartisan:
Notes we're taking while we work on this: https://docs.google.com/document/d/1CD2BucVBFki0ujtUGiApOQc340Ts4sYLcNZjoI9kbjw/edit

app/views/layouts/viewer_layout.html.erb
```diff
@@ -2,6 +2,7 @@
 <head>
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
     <%= csrf_meta_tag %>
+    <script src="https://cdn.jsdelivr.net/npm/hls.js@1"></script>
```
hackartisan, line 5: Does it not work without this? If the mime type is the fix maybe bringing in hls via vite is sufficient?

PR #5820, opened 2023-05-16
Combine all metadata query filters into one

### LINKED ISSUE #5788: fixity dashboard does not load
Investigate which query is slowing this page down and come up with a strategy to speed it up. Either improving the query, caching, something else?

hackartisan:
closes #5788

I tried this on staging. It didn't make a difference, but staging wasn't slow to begin with. The explain analyze looks really promising so I think it's worth merging and trying in production.


hackartisan:
the fixity dashboard on staging is messed up I think for other reasons. I'll put this in draft for now.

tpendragon (approved):

PR #5826, opened 2023-05-17
Add and use default values in stub_ezid

### LINKED ISSUE #5666: Document "stub_ezid"
Help the caller know what values to use for `shoulder` and `blade` params.

https://github.com/pulibrary/figgy/blob/9ab6a3450590964f02b18bc5a2a6d146d17aa7c0/spec/support/stub_ezid.rb

hackartisan:
For the most part the test does not care what the shoulder and blade are. Most of the time, we just need to stub it to make webmock happy. Provide default values and use them in most specs. This will reduce confusion of what the values should be, obviating the need for documentation on the method.

Where tests were actually exercising ark behavior, I left custom values in place.

closes #5666


tpendragon (approved):
This is great! Maybe i won't have to copy this every time now

PR #5841, opened 2023-05-23
Ensure deletion markers have titles

### LINKED ISSUE #5803: Deletion markers should have appropriate titles
- [x] Ephemera Vocabularies
- [x] Ephemera Terms
- [x] Ephemera Fields
- [x] Numismatics Accessions
- [x] Numismatics Firm
- [x] Numismatics Person
- [x] Numismatics Place

They show up in the listing as `(Deletion Marker)`

Created during preservation fire drill, #5557

eliotjordan:
Closes #5803

### DIFF
--- a/spec/change_set_persisters/change_set_persister_spec.rb
+++ b/spec/change_set_persisters/change_set_persister_spec.rb
@@ -1426,6 +1426,84 @@ RSpec.describe ChangeSetPersister do
         expect(File.exist?(disk_preservation_path.join(resource.id.to_s, "#{resource.id}.json"))).to eq false
       end
     end
+
+    context "when deleting a numismatic place resource" do
+      it "assigns a title to the deletion marker" do
+        resource = FactoryBot.create_for_repository(:numismatic_place)
+        change_set_persister.delete(change_set: ChangeSet.for(resource))
+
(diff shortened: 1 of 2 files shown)


spec/change_set_persisters/change_set_persister_spec.rb
```diff
@@ -1426,6 +1426,84 @@
...
     end
+
+    context "when deleting a numismatic place resource" do
+      it "assigns a title to the deletion marker" do
+        resource = FactoryBot.create_for_repository(:numismatic_place)
+        change_set_persister.delete(change_set: ChangeSet.for(resource))
+
+        deletion_markers = change_set_persister.query_service.find_all_of_model(model: DeletionMarker)
+        deletion_marker = deletion_markers.find { |m| m.resource_id == resource.id }
+        expect(deletion_marker.resource_title).not_to be_nil
```
hackartisan, line 1437: Is this title dynamically generated? I think it would be nice if we could test for the string we expect, rather than not nil.

hackartisan (changes requested):

spec/change_set_persisters/change_set_persister_spec.rb
eliotjordan: They are generated. Added.

eliotjordan:
@hackartisan ready for re-review.

PR #5842, opened 2023-05-23
Add rights statement to bulk ingest

### LINKED ISSUE #5751: Add Rights Statement to Bulk Ingest
### Summary

The rights statement needs to be changed to Copyright Not Evaluated for finding aids material the studio's ingesting, but there's no way to do it from bulk ingest.

### Impact 

Would make bulk ingest significantly easier, so they don't have to bulk ingest to a new collection then bulk edit, or ingest one at a time. This was requested by the studio in #digital_studio.

### Priority recommendation

- [x] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

eliotjordan:
Closes #5751 

Screenshot 2023-05-23 at 12 26 16 PM

### DIFF
--- a/app/views/bulk_ingest/show.html.erb
+++ b/app/views/bulk_ingest/show.html.erb
@@ -38,6 +38,13 @@
                 <%= select_tag "collections", options_for_select(@collections), multiple: true, class: "form-control" %>
               </div>
               <div class="form-group">
+                <label for="rights-statement">Rights Statement</label>
+                <%= select_tag :rights_statement,
+                  options_from_collection_for_select(ControlledVocabulary.for(:rights_statement).all, "value", "label"),
+                  class: 'form-control rights-statement',
+                  required: true %>
+              </div>
+              <div class="form-group">
                 <label for="holding_location">Holding Location</label>
                 <%= select_tag :holding_location,
                   options_from_collection_for_select(ControlledVocabulary.for(:holding_location).all, "value", "label"),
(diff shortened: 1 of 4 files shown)


hackartisan (changes requested):
Is there a controller test that ensures the resource doesn't have a rights statement value when that parameter isn't sent? I think that would be nice to have.

eliotjordan:
@hackartisan You have to choose a rights statement. It's a required field in the regular new resource form as well as the bulk ingest form. Previously, bulk ingested resources all received a default value.

app/views/bulk_ingest/show.html.erb
```diff
@@ -37,6 +37,13 @@
                 <label for="collections">Collection(s)</label>
                 <%= select_tag "collections", options_for_select(@collections), multiple: true, class: "form-control" %>
               </div>
+              <div class="form-group">
+                <label for="rights-statement">Rights Statement</label>
+                <%= select_tag :rights_statement,
+                  options_from_collection_for_select(ControlledVocabulary.for(:rights_statement).all, "value", "label"),
+                  class: 'form-control rights-statement',
+                  required: false %>
```
hackartisan, line 45: Make this `required: true`?

hackartisan (changes requested):
Aha, I see. Okay can we make the form require it?

app/views/bulk_ingest/show.html.erb
eliotjordan: 👍

eliotjordan:
@hackartisan Ready for re-review.

hackartisan (approved):
Hm. that doesn't seem to make the "required" label show up on the form. but I'm thinking maybe we don't block this PR on that, what do you think @eliotjordan

eliotjordan:
@hackartisan Yeah, I think it's fine to merge.

PR #5844, opened 2023-05-24
Separate repairing and in progress statuses on health report

### LINKED ISSUE #5823: Health report should not display "repairing" message when there are no fixity events
Today the sidekiq queue was full so a lot of fixity checks were deferred longer than usual after ingest.

The resource show page displays a repairing message, when in fact no fixity checks have been run yet. See screenshots.

Screenshot 2023-05-16 at 1 04 39 PM
Screenshot 2023-05-16 at 1 04 50 PM

## Implementation

When repairing, the status should be "Repairing" instead of "in Progress", and "in Progress" is what should display when there are no local fixity events yet (or not enough.)

hackartisan:
closes #5823


hackartisan:
Re: the failing test, the false positives are caused by https://github.com/pulibrary/figgy/blob/a5de0eab0166e864e97c15cc7115af357c081d0a/spec/wayfinders/wayfinder_spec.rb#L146-L155

But I'm not sure about the way it's written. I would think that events on all nodes of a resource should be reported. Hopefully @tpendragon can weigh in on this setup.

eliotjordan (approved):

PR #5853, opened 2023-05-25
Check self for cloud/local fixity tests

### LINKED ISSUE #5807: Re-instate Health Report on a File Set page
The health report was removed as part of https://github.com/pulibrary/figgy/pull/5789 because it was showing that the file was healthy even when it was not, for both local and cloud fixity.

To put this back in we need to update the health report specifically for when the resource is a file set.

## Implementation Notes

We might be able to fix this by ensuring that the health report checks the resource itself as well as its members - right now it just checks its members, and FileSets don't have any.

We're not sure if for cloud if it's checking non-fileset member events. We want it to if not.

tpendragon:
Closes #5807 

This adds the file set health report back and makes sure the health report checks metadata events for nested members as well as itself.

### DIFF
--- a/app/values/health_report/local_fixity_check.rb
+++ b/app/values/health_report/local_fixity_check.rb
@@ -20,31 +20,23 @@ class HealthReport::LocalFixityCheck
 
   def status
     @status ||=
-      if fixity_map[0]&.positive?
+      if wayfinder.deep_failed_local_fixity_count.positive?
         :needs_attention
-      elsif fixity_map[2]&.positive?
+      elsif wayfinder.deep_repairing_local_fixity_count.positive?
         :repairing
-      elsif fixity_map[nil]&.positive?
+      elsif unknown_count.positive?
         :in_progress
       else
         :healthy
       end
   end
 
-  def fixity_map
-    @local_fixity_map ||=
-      begin
-        m = {}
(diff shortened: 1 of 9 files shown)


config/locales/en.yml
```diff
@@ -182,13 +182,23 @@ en:
     local_fixity_check:
       type: 'Local Fixity'
       summary:
+        self:
+          in_progress: "Local fixity check is in progress this resource. If this doesn't resolve itself then contact DLS with a link to this resource via #digital-library."
```
hackartisan, line 186: there's a word missing -- maybe it should be"in progress for this resource"

app/values/health_report/local_fixity_check.rb
```diff
@@ -20,31 +20,23 @@ def type
...
         :repairing
-      elsif fixity_map[nil]&.positive?
+      elsif unknown_count.positive?
         :in_progress
       else
         :healthy
       end
   end
 
-  def fixity_map
```
hackartisan, line 34: nice refactor

hackartisan (changes requested):

config/locales/en.yml
tpendragon: Good catch - fixed.

hackartisan (approved):

Issue #5780, opened 2023-05-01
Reports Needed for 2 Collections (Patron Requests, Patron Requests Portions)


kelea99:
### Summary
I need figgy reports generated for 2 collections (see below):
### [Patron Requests](https://figgy.princeton.edu/collections/b7e8c42b-6fd0-4cfb-bc6d-b5e5a5c79f5c/manifest) (NOT including Microfilm)

Total Image Counts for itemized visibilities (Open, Private, Princeton Only, Reading Room)

05/01/2022-05/01/2023 ingests (NOT including microfilm)
- Total Image Counts for itemized visibilities (Open, Private, Princeton Only, Reading Room)
- Total titles count for visibilities (Open, private, prinecton only, reading room)

### [Patron Request Portions](https://figgy.princeton.edu/collections/a5454850-8b97-4750-8390-2c5884134f1f/manifest)

Total image counts for itemized visibilities (Open, Private, Princeton Only, Reading Room)

05/01/2022-05/01/2023 ingests
- Total Image Counts for itemized visibilities (Open, Private, Princeton Only, Reading Room)
- Total titles count for visibilities (Open, private, prinecton only, reading room)

### Impact 
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._
Not stopping work, but would help greatly to have for a meeting.
### Priority recommendation

- [x] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "Maintenance/Research" labels_
Requested for meeting on 05/02/2023

tpendragon:
Done. The script was this:

```ruby
csp = ChangeSetPersister.default
query_service = csp.query_service
collection_id = "b7e8c42b-6fd0-4cfb-bc6d-b5e5a5c79f5c"
collection = csp.query_service.find_by(id: collection_id)
members = Wayfinder.for(collection).members
# Filter out microfilm
members = members.select do |member|
  !Array.wrap(member.call_number).any? { |x| x.to_s.include?("MICROFILM") }
end
# Filter out anything that is a volume
members = members.select do |member|
  Wayfinder.for(member).parent.blank?
end
# Filter out anything before May 1 2022
members = members.select do |member|
  member.created_at >= Time.new(2022, 0o6, 0o1)
end
members = members.group_by { |x| x.visibility.first }
result = Hash[
[...]

Issue #5677, opened 2023-02-16
Build UI to display the item and image count for a given collection over a given time frame


tpendragon:
This UI will allow someone to independently generate the reports like those requested in https://github.com/pulibrary/figgy/issues/5627.

## Implementation Options

1. Provide a facet for date created in search results and some summary of item and image count in that result set.

* Range slider for Date Created
  * There's a date range picker in lux we could use instead.
* Summary box? A button for "Reports"?
* It might take a long time for these numbers to be calculated, so that might be a complication here.
* @kelea99 also notes that users explicitly want a report they can send around.

2. Create a CSV export button for search results, include a summary row in that CSV that has "Image Count", "Resource Count", and "Megabytes". Email users when that report is ready (or something)

  * This solution also requires the date range facet for date created
  * Maybe include a totals row, and in each record row have "Image Count". Don't include "Resource Count", because we have no way to correlate "physical objects" to "digital representations."

sdellis:
Date Range and Collection ID fields should be required and validated. No "since forever" start dates.

Issue #5627, opened 2023-01-18
Generate Report for "Patron Requests - Portions" and "Patron Requests" for stakeholder


kelea99:
### Summary
From Jon: Would like a report of all items in the Patron Requests - portions and Patron Requests collections in figgy, minus the Microfilms. 

Date parameters: 7-1-2021 through 6-30-2022
Collections: [Patron Requests](https://figgy.princeton.edu/catalog/b7e8c42b-6fd0-4cfb-bc6d-b5e5a5c79f5c) , [Patron Requests - Portions](https://figgy.princeton.edu/catalog/a5454850-8b97-4750-8390-2c5884134f1f)

Do not include: items in the Microfilm collection

Needs: image count , item count

 ### Impact 
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._
"Sometime in the next few days" @jpstroop , is there a hard deadline?
### Priority recommendation

- [x] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "Maintenance/Research" labels_
See above

hackartisan:
This is immediate need; more permanent solution will also be necessary; see #5677

tpendragon:
We now know the resource count is 2043:

```sql
select COUNT(id) from orm_resources WHERE created_at between '2021-07-01' and '2022-06-30' AND metadata @> '{"member_ids": []}' AND (metadata @> '{"member_of_collection_ids": [{"id": "b7e8c42b-6fd0-4cfb-bc6d-b5e5a5c79f5c"}]}' OR metadata @> '{"member_of_collection_ids": [{"id": "a5454850-8b97-4750-8390-2c5884134f1f"}]}') AND NOT metadata @> '{"member_of_collection_ids": [{"id": "d7e0141b-ff30-4a9b-b260-09525662a65c"}]}'

```

tpendragon:
The image count should be 319,450:

```sql
WITH RECURSIVE deep_members AS (
  select member.*
  FROM orm_resources a,
  jsonb_array_elements(a.metadata->'member_ids') AS b(member)
  JOIN orm_resources member ON (b.member->>'id')::UUID = member.id
[...]

PR #5872, opened 2023-06-08
Add a report generator for image counts.

tpendragon:
Relates to #5780, #5677, and #5627

### DIFF
--- a/app/queries/find_by_property.rb
+++ b/app/queries/find_by_property.rb
@@ -14,6 +14,8 @@ class FindByProperty
 
   # Find by an arbitrary property. If property is :metadata, then value should
   # be a hash to query for the value of multiple properties.
+  # @param created_at [DateTime, Range<DateTime>] Either a date or a range of
+  #   dates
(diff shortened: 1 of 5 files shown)


app/queries/find_by_property.rb
```diff
@@ -14,7 +14,7 @@ def initialize(query_service:)
 
   # Find by an arbitrary property. If property is :metadata, then value should
   # be a hash to query for the value of multiple properties.
-  def find_by_property(property:, value:, model: nil, lazy: false)
```
hackartisan, line 17: It would be helpful to add to the comment about the created_at parameter, in particular its expected type and that it should be a range

hackartisan (changes requested):
It looks good, just one request for a comment.

app/queries/find_by_property.rb
tpendragon: Done.

hackartisan (approved):

PR #5879, opened 2023-06-22
Preserve collections

### LINKED ISSUE #5798: Collections do not restore from deletion
Behavior observed: You can find the collection's deletion marker in the deletion marker search UI, and you can click its "restore" button. Then you get a flash message saying it is queued for restoration. However, it does not restore and its deletion marker continues to be listed in that search UI.

Question: Since you can only delete a collection once it's been emptied, is there a use case for restoring a collection? Do we want to make this work or do we want to make it not available as a restore target?

Discovered during fire drill exercise, #5557

tpendragon:
Closes #5798

### DIFF
--- a/app/change_sets/collection_change_set.rb
+++ b/app/change_sets/collection_change_set.rb
@@ -18,4 +18,8 @@ class CollectionChangeSet < ChangeSet
   def primary_terms
     [:title, :slug, :source_metadata_identifier, :description, :owners, :restricted_viewers]
   end
+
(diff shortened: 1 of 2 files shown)


eliotjordan (approved):

eliotjordan (changes requested):
Nice and easy. Just needs a rebase.

tpendragon:
Rebased.

eliotjordan (approved):

PR #5889, opened 2023-06-29
Allow download of geotiff from scanned map with single raster child

### LINKED ISSUE #5710: Allow GeoTiff download option for single sheet GeoTiff
Allow GeoTiff download option in the UV for for single sheet GeoTiffs.

Probably need to make an adjustment here:
https://github.com/pulibrary/figgy/blob/27cb7b25edcb20e7865b62a9414d5933d712ef09/app/services/manifest_builder/canvas_rendering_builder.rb#L26

Example in Maps Portal: https://maps.princeton.edu/catalog/princeton-dccz30q432r
Example in Figgy: https://figgy.princeton.edu/catalog/3af117c8-96d9-40b9-9aeb-783f41c9d48f
File that should be downloadable: https://figgy.princeton.edu/catalog/parent/c79243a8-3ef5-43a4-9403-736bef4c96d0/22d70018-0961-409f-a814-88d3a5c3473d

Note: This already works if there's multiple sheets (see code above), but doesn't work if there's only one sheet. That got implemented here: https://github.com/pulibrary/figgy/pull/5051

eliotjordan:
Closes #5710

### DIFF
--- a/app/services/manifest_builder.rb
+++ b/app/services/manifest_builder.rb
@@ -604,15 +604,33 @@ class ManifestBuilder
       Valkyrie::ResourceDecorator.new(resource).header
     end
 
-    # Returns a geotiff child for MapSets which have attached Raster resources.
+    # Returns a geotiff FileSet for ScannedMaps and MapSets which have attached Raster resources.
+    # Use Case 1:
+    # ScannedMap (MapSet) ->  ScannedMap -> RasterResource -> FileSet (target set to tiles)  + FileSet (no target - uncropped)
+    #
+    # Use Case 2:
+    # ScannedMap -> RasterResource -> FileSet (target set to tiles - uncropped)
     def geotiff_child
       @geotiff_child ||=
         begin
           return false unless parent_node.is_a?(ScannedMapNode)
           wayfinder = Wayfinder.for(resource)
           parent_wayfinder = Wayfinder.for(wayfinder.parent)
-          if parent_wayfinder.mosaic_file_count == 1
-            parent_wayfinder.uncropped_geotiff_filesets.first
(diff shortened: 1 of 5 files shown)


app/services/manifest_builder.rb
```diff
@@ -611,8 +611,15 @@ def geotiff_child
           return false unless parent_node.is_a?(ScannedMapNode)
           wayfinder = Wayfinder.for(resource)
           parent_wayfinder = Wayfinder.for(wayfinder.parent)
-          if parent_wayfinder.mosaic_file_count == 1
-            parent_wayfinder.uncropped_geotiff_filesets.first
+          raster_resource = parent_wayfinder.raster_resources&.first
```
tpendragon, line 614: Can I get some comments about what these individual use cases are? I'm having a tough time following.

tpendragon (changes requested):

app/services/manifest_builder.rb
eliotjordan: This took me a long time to untangle as well. The main issue is that in cases where a single scanned map (not part of a MapSet) has a raster child, there is only one geotiff fileset (versus two for the MapSet mosaic case). 

ScannedMap -> RasterResource -> GeoToff FileSet (target set to tiles - uncropped)

vs.

ScannedMap -> ScannedMap -> RasterResource -> GeoTiff FileSet (target set to tiles) + GeoTiff FileSet (no target - uncropped)

The old logic only returned uncropped GeoTiff files sets with the target set to 'tiled", which doesn't work for the single image non-mosaiced resources.

eliotjordan: Added comments to the method.

tpendragon (approved):

Issue #5903, opened 2023-07-14
Migrate to Postgres 15


tpendragon:
Postgres 10 is super out of date and difficult to even get installed now it's so unsupported. Migrate to figgy-db-prod1.

Notes: https://docs.google.com/document/d/1dBN7AllCKwp-AyYc7KVBeQFo6YtWz85q2ShlQnWdUAw/edit#heading=h.kqmal60l50l

## Steps

Ruby
- [x] Update CI to use Postgres 15 (see https://github.com/pulibrary/figgy/pull/5918)

Staging
- [x] Create a staging database infrastructure (two staging database servers) (Create on production infra) (https://github.com/pulibrary/figgy/issues/5947)
- [x] Allow figgy-db-staging1 to access lib-postgres3 (update pg_hba to add figgy-db-staging1's IP)
- [x] Dump figgy_staging database schema
    * `pg_dumpall --schema-only -d figgy_staging -H lib-postgres3 -U postgres -f figgy_staging_schema.sql`
- [x] Dump figgy_staging database users
    * `pg_dumpall --roles-only -H lib-postgres3 -U postgres -f figgy_staging_roles.sql` (then trim it down to the figgy ones)
- [x] Set up the second staging database server as a "warm standby" replica
    * https://hevodata.com/learn/postgresql-streaming-replication/
- [x] Restore figgy_staging_schema.sql & figgy_staging_roles.sql to the new staging database.
- [x] Set up logical replication between lib-postgres3 and new one.
     * On lib-postgres3, connected to figgy_staging: `CREATE PUBLICATION figgy_staging_publication FOR ALL TABLES;`
     * On staging db, connected to figgy_staging: `CREATE SUBSCRIPTION figgy_staging_subscription
CONNECTION 'host=lib-postgres3.princeton.edu port=5432 dbname=figgy_staging user=postgres password=postgrespassword' PUBLICATION figgy_staging_publication;`
- [x] Copy postgres users from lib-postgres3 to figgy-db-staging1
- [x] Verify row counts for all tables between lib-postgres3 and figgy-db-staging1 (ensures replication has finished)
- [x] Set figgy-staging to read only
- [x] Verify row counts for all tables between lib-postgres3 and figgy-db-staging1 (ensure replication is up to date)
- [x] Update sequences in figgy-db-staging1
    * https://wiki.postgresql.org/wiki/Fixing_Sequences
- [x] Swap figgy-staging to figgy-db-staging1.
[...]

PR #5918, opened 2023-07-27
Test and develop on postgres 15

hackartisan:
To make this work locally I had to upgrade lando and upgrade my local postgres (running via brew -- its pg_dump is used to create the schema file stored with the rails project)

advances #5903

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -2,9 +2,11 @@ FROM cimg/ruby:3.1.0-browsers
 
 RUN git clone https://github.com/pulibrary/figgy.git \
   && cd figgy \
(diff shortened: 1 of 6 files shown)


tpendragon:
Things I had to do to get this installing locally:

`HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade postgresql`
Install Lando 3.18.0 from https://github.com/lando/lando/releases

hackartisan:
I just hit a problem since upgrading to postgres 15 locally where running rails server and hitting the database caused a segmentation fault. Adding `export PGGSSENCMODE="disable"` to my zshrc resolved it. there are issues in the pg repo, e.g. https://github.com/ged/ruby-pg/issues/538

tpendragon (approved):

PR #5929, opened 2023-08-01
Prevent deletion of EphemeraTerm if used in EphemeraFolder

### LINKED ISSUE #5806: Do not allow an ephemera term to be deleted if it is used by a folder
Currently if you delete and then restore a term it is not replaced back on the folder it was previous on.

Created during preservation fire drill, https://github.com/pulibrary/figgy/issues/5557

eliotjordan:
Closes #5806

### DIFF
--- a/app/controllers/ephemera_terms_controller.rb
+++ b/app/controllers/ephemera_terms_controller.rb
@@ -12,6 +12,16 @@ class EphemeraTermsController < ResourcesController
     render "index"
   end
 
+  def destroy
+    folders = query_service.custom_queries.find_id_usage_by_model(model: EphemeraFolder, id: resource.id)
+    if folders.empty?
+      super
+    else
+      flash[:alert] = "This is term is currently in use. To delete, please remove from related folders."
+      redirect_back(fallback_location: root_path)
+    end
+  end
+
   def load_vocabularies
     @vocabularies = query_service.find_all_of_model(model: EphemeraVocabulary).map(&:decorate)
   end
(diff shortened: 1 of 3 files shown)


app/controllers/ephemera_terms_controller.rb
```diff
@@ -12,6 +12,16 @@ def index
     render "index"
   end
 
+  def destroy
+    folders = query_service.custom_queries.find_id_usage_by_model(model: EphemeraFolder, id: resource.id)
+    if folders.empty?
+      super
+    else
+      flash[:alert] = "This is term is currently in use. To delete, please remove from related folders."
+      redirect_back(fallback_location: root_path)
```
hackartisan, line 21: redirect_back seems cool I didn't know about that

spec/change_set_persisters/change_set_persister/cleanup_terms_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ChangeSetPersister::CleanupTerms do
```
hackartisan, line 5: Interesting, we needed to add this test because we can't exercise the code any other way? Is it still useful to have the cleanup terms even if it never executes? Do we think it will execute in some circumstance, like if we force something via console?

hackartisan (commented):
This looks good, blocking for discussion question.

hackartisan (changes requested):
actually blocking this time for real now (discussion question)

spec/change_set_persisters/change_set_persister/cleanup_terms_spec.rb
eliotjordan: That was my thought. It can be deleted in the console and we should still ensure that the ids are cleaned up.

hackartisan (approved):

PR #5930, opened 2023-08-02
Add unique index for metadata events per resource id

hackartisan:
This may help us pinpoint where duplicate metadata events are created and prevent them
refs #5938


tpendragon:
I think this is a good idea, and will probably prevent what's likely a race condition.

Some things to think about:

1. We're logically replication lib-postgres3 to figgy-db-prod1 right now - running this won't create the index on figgy-db-prod1 (schema doesn't replicate)
2. When we make this index, we probably want to make it CONCURRENTLY, so it doesn't lock the db.

hackartisan:
Added the concurrently algorithm to the migration.

eliotjordan (approved):

PR #5956, opened 2023-08-11
Use "Sound" instead of "Audio" for manifests

### LINKED ISSUE #5955: IIIF Type should be "sound" for audio files
https://iiif.io/api/presentation/3.0/#type

Audio isn't a specified value.

hackartisan:
closes #5955


tpendragon:
You tested this in staging? I feel like there was a bug in the version of UV we had at the time that forced this issue

hackartisan:
ended up pulling this onto the UV upgrade branch, so I'll close this one.

Issue #5971, opened 2023-09-08
Move DK Ephemera  Project Boxes 22 and 23 to SAE Ephemera Project


kelea99:
### Summary

DK agencies has finished metadata for boxes 22 and 23 in [DK Ephemera Project](https://figgy.princeton.edu/catalog/f09fc91d-7a9b-47b5-afff-ce7db76b4e92). Please move Boxes 22 and 23 to [SAE Ephemera Project](https://figgy.princeton.edu/catalog/f99af4de-fed4-4baa-82b1-6e857b230306https://figgy.princeton.edu/catalog/f99af4de-fed4-4baa-82b1-6e857b230306).

### Impact 
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._
Before September 20th
### Priority recommendation

- [ ] asap
- [X] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "Maintenance/Research" labels_
Due to meeting with journalists and photographer for article featuring Ellen  and Aditya.

tpendragon:
Note: Add how you did this to this ticket. We'll be doing it again.

kelea99:
@tpendragon  - checking in on this one can we get this done today?

tpendragon:
Yup it's at the tip top of the priority.

tpendragon:
Done. Commands:

```
bundle exec rake figgy:bulk:move_box TARGET_PROJECT_ID=f99af4de-fed4-4baa-82b1-6e857b230306 BOX_ID=e41600e1-b49a-4bb8-b191-85661ee95f06
bundle exec rake figgy:bulk:move_box TARGET_PROJECT_ID=f99af4de-fed4-4baa-82b1-6e857b230306 BOX_ID=2b544ba0-0d23-480d-b982-fd0a18dc9c8b
```

kelea99:
Thank you! slacked stakeholder Ellen A. !

PR #5976, opened 2023-09-19
Add a rake task to move boxes.

tpendragon:
Work towards #5971

### DIFF
--- a/lib/tasks/bulk.rake
+++ b/lib/tasks/bulk.rake
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 namespace :figgy do
   namespace :bulk do
+    desc "Move a box to a new project"
+    task move_box: :environment do
+      target_project_id = ENV["TARGET_PROJECT_ID"]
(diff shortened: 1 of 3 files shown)


lib/tasks/bulk.rake
```diff
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 namespace :figgy do
   namespace :bulk do
+    desc "Move a box to a new project"
+    task move_box: :environment do
+      target_project_id = ENV["TARGET_PROJECT_ID"]
```
sdellis, line 6: I may be misinterpreting this, but I'm curious if this meens we need to add environment variables each time before this is run for the target project and the box? Would it make more sense to specify them as arguments that can be added onto the rake command?

tpendragon: Arguments in rake commands are pretty tough, so we use env variables in the command line as a replacement.

sdellis (approved):
LGTM 👍

PR #5986, opened 2023-09-22
Adds UI for Item/Image Count Report

sdellis:


hackartisan (approved):

app/views/reports/collection_item_and_image_count.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      <th>Reading Room Titles</th>
+      <th>Princeton Only Titles</th>
+      <th>Open Image Count</th>
+      <th>Private Image Count</th>
+      <th>Reading Room Image Count</th>
+      <th>Princeton Only Image Count</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @report.to_h.each do |r| %>
```
hackartisan, line 30: Here's where it calls to_h. I wonder if we just needed a feature spec. That could also be useful for ensuring there's not a flash message when the report is first loaded (as would have been the case at some point during our refactors I think)

hackartisan (changes requested):

hackartisan (approved):

PR #6002, opened 2023-10-02
Allow search results for child records with parameter

### LINKED ISSUE #6001: Display child records in search results when needed
Dan Walker has a very large set of georeferenced tiffs that are organized by ARK. These ARKs refer to child images in Sandborn volumes and can't be searched which makes his work tedious. 

Proposal is to add a param (similar all_models=true) that allows for search results for child records.

eliotjordan:
Closes #6001

- Adds membership facet with two values: "no parent" and "has parent"
- Sets the default facet to "no parent" to replicate the current default search filter
- Removes the filter_parented search filter
- Allows for searching child resources

Screenshot 2023-10-03 at 12 13 55 PM

Screenshot 2023-10-03 at 12 14 08 PM

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -156,6 +156,11 @@ class CatalogController < ApplicationController
       assumed_boundaries: [1100, Time.current.year + 1],
       segments: true
     }
+    config.add_facet_field "membership_facet", label: "Membership", query: {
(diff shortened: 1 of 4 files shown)


tpendragon (changes requested):
Suggestion: Is it possible to add this as a facet? Maybe one of those calculated ones? I think it'd be nice to have this be part of the UI and not a hidden feature.

tpendragon (approved):

PR #6003, opened 2023-10-04
Adds error notifications to File Manager files

### LINKED ISSUE #5824: Implement Preservation Errors on FileSet items in File Manager UI
Implements [#5528](https://github.com/pulibrary/figgy/issues/5528#issuecomment-1398705984)

### Expected Behavior
This mockup provides preservation error messaging given that there may be several types of errors on a given fileset, that preserved filesets should display "preserved", and cards should remain the same size. The errors and warnings stack on top of each other below the thumbnail. 

We will need to display the following possible error messages:

- Repairing Derivatives
- Repairing Local Fixity
- Repairing Cloud Fixity
- Derivatives Failed | Report
- Local Fixity Failed | Report
- Cloud Fixity Failed | Report

filemanager-fixity5

One thing to note is that a user may "Report" a fixity failure to DLS to try to fix. Clicking that link will generate a Honeybadger notification with the `user's id`, the `filemanager page URL`, and the `ID of the problematic fileset`. This will allow us to easily create a GitHub ticket with the click of a button. The "Report" link will not generate events or save state to the server.

Team consensus is that, at a minimum, there will be some system notification to the user that the issue was reported. This may be a flash alert or in-place animation that confirms it was successfully reported (or not). The state of which files have been reported would not be saved on the server. 

Note: If state is not saved, it could be hard for a user to know which issues have been reported or not if there are a lot of errors in the Filemanager interface. We could mitigate this confusion by storing ids that have been reported locally on the client, saving the state at least for the duration of the user's session or until they clear their cookies. However, this could be added later if the use case arises and should not get in the way of closing  this ticket.

sdellis:
closes #5824

### DIFF
--- a/app/assets/stylesheets/components/file_manager.scss
+++ b/app/assets/stylesheets/components/file_manager.scss
@@ -176,6 +176,7 @@
   width: 100%;
   bottom: 0px;
   align-items: center;
+  font-size: .9em;
 }
 
 .fm_error {
--- a/app/assets/stylesheets/components/file_manager.scss
+++ b/app/assets/stylesheets/components/file_manager.scss
@@ -171,16 +171,46 @@
   z-index: 10000 !important
 }
 
-.fm_error {
+.fm_errors {
   position: absolute;
   width: 100%;
-  border: 3px solid #f8a000;
   bottom: 0px;
-  border-bottom-right-radius: 4px;
-  border-bottom-left-radius: 4px;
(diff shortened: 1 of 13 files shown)


sdellis:
Note: This PR may have to wait until a more performant way of querying for fixity errors can be implemented.

tpendragon (approved):

Issue #5995, opened 2023-09-27
FileSet errors aren't saving on the `original_file` on derivative failure.


tpendragon:
We have a bunch of code in our DerivativeServices to save these, but they get rolled back by the transaction rollback system in place for our composite DerivativeServices like:

https://github.com/pulibrary/figgy/blob/2f370aaf8c30f2fa2b42ec9ed81164de79cbbd87/app/derivative_services/default_derivative_service.rb#L68-L74

If you write a test that uses the DefaultDerivativeService instead of the tested derivative services it replicates the error. That transaction needs to stick around, but we need a pattern to allow updating that FileSet outside of the transaction independent of success/failure.

hackartisan:
How do I replicate? There's a suggestion but it's not specific.

tpendragon:
You can replicate by changing this:

https://github.com/pulibrary/figgy/blob/5978-preservation-status/spec/derivative_services/vips_derivative_service_spec.rb#L165

to be `DefaultDerivativeService.new` instead of `derivative_service.new` - the right thing is probably to add a "malformed file" test to DefaultDerivativeService.

PR #6048, opened 2023-10-24
Register update_error_message to happen after_rollback.

tpendragon:
Delaying actions until after the rollback lets us update error messages
after they've been undone by the transaction rolling back.

Closes https://github.com/pulibrary/figgy/issues/5995


app/derivative_services/image_derivative_service.rb
```diff
@@ -154,10 +156,10 @@ def cleanup_derivative_metadata(derivatives:)
...
     def update_error_message(message:)
-      target_file.error_message = [message]
+      resource = change_set_persister.query_service.find_by(id: self.resource.id)
+      target = resource.file_metadata.find { |x| x.id == target_file.id }
+      target.error_message = [message]
       updated_change_set = ChangeSet.for(resource)
-      change_set_persister.buffer_into_index do |buffered_persister|
-        buffered_persister.save(change_set: updated_change_set)
-      end
+      change_set_persister.save(change_set: updated_change_set)
```
hackartisan, line 163: Sometimes you changed this to pull it out of a `buffer_into_index` and sometimes not. Can you say a little something about each case?

I sort of wonder whether we should pull this method up to a superclass but it appears there isn't one!

app/change_set_persisters/change_set_persister.rb
```diff
@@ -139,13 +140,19 @@ def buffer_into_index
...
       metadata_adapter.persister.buffer_into_index do |buffered_adapter|
         with(metadata_adapter: buffered_adapter) do |buffered_changeset_persister|
           yield(buffered_changeset_persister)
           delayed_queue = buffered_changeset_persister.delayed_queue
         end
       end
+      @after_rollback = nil
       delayed_queue.run
+    rescue StandardError => error
+      after_rollback.run
```
hackartisan, line 153: that's cool.

hackartisan (changes requested):
Blocking on a question.

app/derivative_services/image_derivative_service.rb
tpendragon: There was a looping error at one point, but to be honest I think it was just whatever I copied/pasted from. 😂 

I thought about superclass too, but it looked like the "target" was different in different derivative services.

tpendragon: I can put 'em back in `buffer_into_index` so there's less of a diff.

hackartisan (approved):

PR #6053, opened 2023-10-24
Add Derivative Health Report

tpendragon:
TODO: Make it work for FileSets
TODO: Figure out a way to replicate locally. Right now bad files are failing characterization.


app/values/health_report/derivative_check.rb
```diff
@@ -0,0 +1,64 @@
...
+    else
+      :healthy
+    end
+  end
+
+  def summary
+    if resource.respond_to?(:member_ids)
+      I18n.t("health_status.derivative_check.summary.#{status}")
+    else
+      I18n.t("health_status.derivative_check.summary.self.#{status}")
```
hackartisan, line 47: They don't look different? What am I missing?

hackartisan (changes requested):

app/values/health_report/derivative_check.rb
hackartisan: To be more specific, it looks like the translations are identical in both cases. I think we don't need a code branch here, and we could just have one set of translations.

tpendragon: Oh I copied it from the others. I can remove it.

hackartisan (approved):

Issue #5964, opened 2023-08-30
[Figgy/production] Notice: FileAppender: Failed to append the new file converted_from_pdf_page_17.tiff for e376e282-936f-40be-8f93-6558cd401fa1 to resource 265da7e9-e4e4-43f5-8a4a-cd91e53a883d: No such file or directory @ rb_sysopen - /tmp/pdf_derivatives20230830-20415-csde2d/intermediate_file16.tif


tpendragon:
## Backtrace

    line 137 of [PROJECT_ROOT]/app/services/file_appender.rb: rescue in create_node
    line 114 of [PROJECT_ROOT]/app/services/file_appender.rb: create_node
    line 144 of [PROJECT_ROOT]/app/services/file_appender.rb: block in file_nodes

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/95398408)

tpendragon:
Example of a broken record: https://figgy.princeton.edu/catalog/265da7e9-e4e4-43f5-8a4a-cd91e53a883d

tpendragon:
We want this to fail and retry, and maybe clean up after itself if it fails. Right now we're getting broken records that you can only tell are broken if you do manual QA.

tpendragon:
Sometimes this happens during PDF generation and sometimes it happens for PDF generation. The previous example is a PDF generation error, https://figgy.princeton.edu/catalog/c938a0dd-9af8-48f4-8d93-0c6e1d4fa807 is one where there should be a second page but it didn't get added because of a FileAppender error and now it's complete with one less page than it should have.

tpendragon:
The sudden priority portion of this ticket is closed by #6054.

hackartisan:
@tpendragon do you have a sense of what remains here?

tpendragon:
@hackartisan I think something like "FileAppender shouldn't error quietly" - the remaining issue is ones like https://figgy.princeton.edu/catalog/c938a0dd-9af8-48f4-8d93-0c6e1d4fa807 where it errored and now is missing a file and nobody knew.

PR #6054, opened 2023-10-25
Stop using temp files for PDF Derivatives

tpendragon:
They're randomly disappearing, likely because of garbage collection. A cleanup process we run daily can instead make sure that abandoned files are deleted.

Work towards #5964

### DIFF
--- a/lib/tasks/clean.rake
+++ b/lib/tasks/clean.rake
@@ -26,5 +26,17 @@ namespace :figgy do
 
       tus_storage.expire_files(expiration_date)
     end
+
+    desc "Clean old derivative processing files"
+    task derivative_artifacts: :environment do
+      all_files = Dir.glob("#{Dir.tmpdir}/derivative_generation")
+      expiration_date = 1.day.ago
+      all_files.each do |file|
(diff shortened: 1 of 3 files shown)


lib/tasks/clean.rake
```diff
@@ -26,5 +26,17 @@ namespace :figgy do
 
       tus_storage.expire_files(expiration_date)
     end
+
+    desc "Clean old derivative processing files"
+    task derivative_artifacts: :environment do
+      all_files = Dir.glob("#{Dir.tmpdir}/derivative_generation")
```
eliotjordan, line 32: issue: I think this should be `Dir.glob("#{Dir.tmpdir}/derivative_generation/*")` Tried this in irb and the trailing `/*` was necessary for the match.

eliotjordan (changes requested):

lib/tasks/clean.rake
tpendragon: Thanks - fixed.

eliotjordan (approved):

Issue #6031, opened 2023-10-19
Add facet for collection name to deletion marker search


hackartisan:
This means that when a resource is deleted we have to store the name of the collection it was a member of.

We're currently stashed the entire object in the deletion marker. Do a migration to look up member_of_collection_ids on deletion markers created before this issue was closed and add them to the deletion marker.

created out of #5557

eliotjordan:
Migrated DeletionMarkers.

![Screenshot 2023-10-26 at 12 57 35 PM](https://github.com/pulibrary/figgy/assets/784196/9504393b-5660-41eb-9924-88fe5352e9cc)

PR #6055, opened 2023-10-25
Index DeletionMarker deleted resource collection titles

eliotjordan:
Makes progress on #6031

### DIFF
--- a/app/indexers/collection_indexer.rb
+++ b/app/indexers/collection_indexer.rb
@@ -3,11 +3,7 @@ class CollectionIndexer
   delegate :query_service, to: :metadata_adapter
   attr_reader :resource
   def initialize(resource:)
-    @resource = if resource.is_a?(DeletionMarker)
-                  resource.deleted_object
-                else
-                  resource
-                end
+    @resource = resource
   end
 
   # Adds a collection title index entry to a resource's index document for each collection that resource belongs to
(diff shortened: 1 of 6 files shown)


app/indexers/collection_indexer.rb
```diff
@@ -31,8 +35,9 @@ def ephemera_collection_titles
 
   def collections
     return [] unless resource.respond_to?(:member_of_collection_ids) && resource.member_of_collection_ids
-    @collections ||=
-      query_service.find_references_by(resource: resource, property: :member_of_collection_ids).to_a.map(&:decorate)
+    @collections ||= resource.member_of_collection_ids.map do |id|
+      query_service.find_by(id: id)&.decorate
```
tpendragon, line 39: Won't this mean if I have a bad collection ID now it'll error instead of just not index the name?

tpendragon (changes requested):
Question: What happens when we have to reindex and the collection's gone? Or the name of the collection has changed since I deleted it (this one might actually be a good thing I really don't know.)

Did you look into caching the collection titles in the deletion marker on create?

app/indexers/collection_indexer.rb
eliotjordan: Fixed.

eliotjordan:
@tpendragon Stashing a title in a new property and doing a full migration of DeletionMarker resources seems heavy handed to me. 🤷 Easy enough to change.

app/indexers/collection_indexer.rb
```diff
@@ -3,7 +3,11 @@ class CollectionIndexer
   delegate :query_service, to: :metadata_adapter
   attr_reader :resource
   def initialize(resource:)
-    @resource = resource
+    @resource = if resource.is_a?(DeletionMarker)
```
tpendragon, line 6: I wonder if we should delegate `member_of_collection_ids` on DeletionMarker so we don't have to branch on class?

tpendragon (changes requested):

eliotjordan:
@tpendragon I was going to try it by stashing on the resource. Might be helpful to keep that collection title around if we're keeping deletion markers forever.

eliotjordan:
@tpendragon I updated and I like it better. Ready for re-review.

tpendragon (dismissed):

tpendragon:
Looks great! Good work and iteration.

eliotjordan:
@tpendragon Had a small error that I fixed. Needs review one more time.

tpendragon (approved):

PR #6063, opened 2023-10-27
Prevent deletion of playlist with tracks  and recordings with tracks in playlist

### LINKED ISSUE #6028: Prevent deletion of a track that's part of a playlist
And document this in https://github.com/pulibrary/figgy/blob/main/docs/technical/preservation/deletion.md

We can't restore the fact that it was in the playlist; this forces people to make an explicit choice to delete it both from the playlists and the recording.

It's also consistent with the way we've treated other deletions.

created out of #5557

eliotjordan:
Closes #6028

### DIFF
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -42,6 +42,18 @@ class PlaylistsController < ResourcesController
     @logical_order = WithProxyForObject.new(@logical_order, members)
   end
 
+  def around_delete_action
+    # Only allow deleting empty playlists. Avoids difficulty of restoring a
+    # full playlist from preservation and supporting restoring accidentally
+    # deleted playlists.
+    if @change_set.resource.member_ids.count.positive?
+      flash[:alert] = "Unable to delete a playlist with tracks."
+      redirect_to solr_document_path(@change_set.resource.id.to_s)
+    else
+      yield
+    end
+  end
+
   private
 
     def manifest_builder
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -46,7 +46,7 @@ class PlaylistsController < ResourcesController
     # Only allow deleting empty playlists. Avoids difficulty of restoring a
     # full playlist from preservation and supporting restoring accidentally
     # deleted playlists.
-    if @change_set.resource.member_ids.count.positive?
+    if @change_set.resource.is_a?(Playlist) && @change_set.resource.member_ids.count.positive?
       flash[:alert] = "Unable to delete a playlist with tracks."
       redirect_to solr_document_path(@change_set.resource.id.to_s)
     else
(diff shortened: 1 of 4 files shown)


app/controllers/scanned_resources_controller.rb
```diff
@@ -64,6 +64,20 @@ def save_and_ingest
     end
   end
 
+  def around_delete_action
+    return unless @change_set.is_a? RecordingChangeSet
+
+    # Only allow deleting empty recordings whose tracks are not in playlists.
```
tpendragon, line 70: It's okay if they're not empty right? Maybe juist a small comment fix.

```suggestion
    # Only allow deleting recordings whose tracks are not in playlists.
```

tpendragon (changes requested):

app/controllers/scanned_resources_controller.rb
```diff
@@ -64,6 +64,22 @@ def save_and_ingest
...
+  def around_delete_action
+    # Only allow deleting recordings whose tracks are not in playlists.
+    playlists = if @change_set.is_a? RecordingChangeSet
+                  Wayfinder.for(@change_set.resource).playlists
+                else
+                  []
+                end
+    if playlists.count.positive?
+      playlist_ids = playlists.map(&:id).join(", ")
+      flash[:alert] = "Unable to delete a recording with tracks in a playlist. Plase remove this recording's tracks from the following playlists: #{playlist_ids}"
```
hackartisan, line 76: typo: "Plase" should be "Please"

app/controllers/playlists_controller.rb
```diff
@@ -42,6 +42,24 @@ def structure
     @logical_order = WithProxyForObject.new(@logical_order, members)
   end
 
+  def around_delete_action
+    # Only allow deleting empty playlists. Avoids difficulty of restoring a
+    # full playlist from preservation and supporting restoring accidentally
+    # deleted playlists.
+    count = if @change_set.resource.is_a? Playlist
```
hackartisan, line 49: I think maybe you wrote the other one first and then used the same logic, but I read this one first and it was sort of confusing why there's a count variable. I got a bit tripped up by the zero value (didn't see what it had to do with a yield at first) and I think it would be a little easier to understand if line 55 instead were more like `if @change_set.resource.is_a? Playlist && @change_set.resource.member_ids.count.positive?` What do you think?

Also, it's nice to have these comments!

hackartisan (changes requested):
around delete is interesting; I'm curious why to use this instead of delete itself?

eliotjordan:
@hackartisan We already set a precedent using `around_delete_action` for the same purpose in [Collections](https://github.com/pulibrary/figgy/blob/main/app/controllers/collections_controller.rb#L47). I think it's helpful to be consistent.

app/controllers/playlists_controller.rb
eliotjordan: good suggestion. 👍

hackartisan (approved):

PR #6064, opened 2023-10-27
Try generating pmtiles without guessing max zoom after failure

eliotjordan:
Fixes and issue with pmtiles generation where the original shapfile/dataset only has one feature.


hackartisan (commented):
This seems good in terms of the code. I'm not sure what a feature is, though, or what it has to do with max zoom. Can you add a little more info? It might also be nice to have a screenshot of the failure and how it looks with this fix.

eliotjordan:
@hackartisan The program we use to generate the pmtiles file has an option to automatically set a max zoom level on the dataset. Otherwise you have to set it manually. The command fails when there is only one feature (point in most cases) in the original data. E.g. a dataset of the port locations in Singapore, but there is only one. This setting requires more than one point and the command fails. The PR adds a fallback that reruns the command without this option.

hackartisan:
Thanks for the extra context that's helpful!

hackartisan (approved):

PR #6091, opened 2023-11-17
improve mosaics

### LINKED ISSUE #6090: Improve mosaic functionality
- Streamline building mosaic documents
- Remove requirement to callback to Figgy for every tile request
- Implement auto cache invalidation

eliotjordan:
Along with https://github.com/pulibrary/geoservices-aws/pull/2, this pull request:

1. Simplifies creation mosaic json documents
2. Decouples Figgy and the TiTiler lambda service to some degree; mostly when serving tiles, which improves performance.
3. Addresses potential race conditions when generating mosaic json documents by only allowing one MosaicJob for a resource to run at a time. 
4. Makes it possible to automatically invalidate the TiTiler / CloudFront cache when the mosaic data changes. Currently, the ID for a resource which you would use to invalidate is passed to TiTiler as a query param. While it is possible to cache using query params as keys in CloudFront, it is not possible to invalidate using query params. These PRs move the id from a param to the url path. See followup issue: https://github.com/pulibrary/figgy/issues/6104
5. Sets a path for future work on restricted raster content. 

Closes #6090

- Only queue up one MosaicJob per resource at a time
- Run mosaic job on raster or scanned map ancestors
- Streamline building mosaic documents
- Simplify tile metadata controller
- Check if MosaicJob is currently running
- Allow derivatives to trigger MosaicJobs in non-complete state
- Set Honeybadger to ignore MosaicJob::JobRunning errors
- Add fingerprint check to MosaicJob
- Work on cache invalidation
- Build mosaic and cog links with id in url
- Use default tile path in development 
- Stub MosaicJob in spec

### DIFF
--- a/app/derivative_services/raster_resource_derivative_service.rb
+++ b/app/derivative_services/raster_resource_derivative_service.rb
@@ -188,7 +188,8 @@ class RasterResourceDerivativeService
     def generate_mosaic
       ancestor_resource = find_ancestor(resource)
       return unless ancestor_resource.is_a?(RasterResource) || ancestor_resource.is_a?(ScannedMap)
-      MosaicJob.perform_later(ancestor_resource.id.to_s)
+      fingerprint = query_service.custom_queries.mosaic_fingerprint_for(id: ancestor_resource.id)
+      MosaicJob.perform_later(resource_id: ancestor_resource.id.to_s, fingerprint: fingerprint)
     end
 
     # Recursively find a resource's base ancestor
--- a/app/derivative_services/raster_resource_derivative_service.rb
+++ b/app/derivative_services/raster_resource_derivative_service.rb
@@ -185,6 +185,9 @@ class RasterResourceDerivativeService
       end
     end
 
+    # The mosaic needs to be regenerated if any part of a MapSet's hierarchy changes, so
+    # for any map resource look up in its hierarchy for the root node and regenerate the Mosaic
+    # if it should have one.
     def generate_mosaic
       ancestor_resource = find_ancestor(resource)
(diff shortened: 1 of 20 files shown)


app/jobs/mosaic_job.rb
```diff
@@ -1,16 +1,38 @@
...
+    @resource_id = resource_id
+    @fingerprint = fingerprint
+    return unless valid_fingerprint?
+    raise JobRunning if currently_running?
+    TileMetadataService.new(resource: resource, generate: true).path
   end
 
   private
 
+    def valid_fingerprint?
```
tpendragon, line 20: @hackartisan recommends adding some documentation about the race conditions we're avoiding here.

tpendragon (changes requested):

app/jobs/mosaic_job.rb
eliotjordan: Comments added.

app/derivative_services/raster_resource_derivative_service.rb
```diff
@@ -186,11 +186,21 @@ def update_error_message(message:)
     end
 
     def generate_mosaic
```
tpendragon, line 188:
```suggestion
    # The mosaic needs to be regenerated if any part of a MapSet's hierarchy changes, so
    # for any map resource look up in its hierarchy for the root node and regenerate the Mosaic
    # if it should have one.
    def generate_mosaic
```

tpendragon (changes requested):
Just a small comment suggestion.

tpendragon (approved):

Issue #6081, opened 2023-11-08
SAE/Box 8 ephemera folders do not have default first image thumbnails selected


kelea99:
### Summary
https://figgy.princeton.edu/catalog/819f1f65-a0cc-4571-920e-5de07c303cf2

may be a bug? thumbnails default to first image if none selected by humans. 

### Impact 
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._
disrupting usability of collection in dpul. no thumbnails showing in browse or search lists.

### Priority recommendation

- [ ] asap
- [x] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "Maintenance/Research" labels_
affecting usability of collection in dpul

kelea99:
do in next six weeks (after thanksgiving). thanks!

tpendragon:
I took a quick peek at these because I noticed we had an invalid thumbnails fixer rake task - unfortunately that rake task won't work. These have thumbnails set, but they're set to orphaned file sets. I dunno why.

tpendragon:
## Success Criteria

Write a rake task that will fix all of Box 8's thumbnails.

hackartisan:
Okay I think this works, search for `"Box 8"` in quotes and limit it to the South Asian Ephemera project. https://figgy.princeton.edu/?f%5Bephemera_project_ssim%5D%5B%5D=South+Asian+Ephemera&parent_id=2a9d1d43-c9e1-4b09-b685-872af18aaa84&q=%22box+8%22

PR #6098, opened 2023-11-27
Add rake task to clear bad thumbnails in an ephemera box

hackartisan:
- Add ruby debug as an alternative to pry/byebug
   - I kept hitting this undefined constant error, see https://github.com/deivid-rodriguez/byebug/issues/564
   - to use debug insert `binding.break`
- Add rake task and service  to fix orphan thumbnails

refs #6081

### DIFF
--- a/app/services/migrations/reset_box_thumbnails.rb
+++ b/app/services/migrations/reset_box_thumbnails.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module Migrations
+  class InvalidResourceTypeError < StandardError; end
+
+  class ResetBoxThumbnails
+    def self.call(box_id:)
+      new(box_id: box_id).run
+    end
+
+    attr_reader :box_id
+    def initialize(box_id:)
+      @box_id = box_id
+      raise InvalidResourceTypeError unless box.is_a? EphemeraBox
+    end
+
+    # @return the number of thumbnails that were reset
+    def run
+      progress_bar
+      reset = 0
+      folders.each do |folder|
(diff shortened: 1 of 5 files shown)


spec/services/migrations/reset_box_thumbnails_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        reset = described_class.call(box_id: box.id)
+
+        folder1 = query_service.find_by(id: folder1.id)
+        expect(folder1.thumbnail_id).to eq [file_set1.id]
+        expect(reset).to eq 1
+      end
+    end
+
+    context "when the no thumbnail resource doesn't exist and there are no members" do
```
eliotjordan, line 56: issue: small typo. "when the no thumbnail" -> "when the thumbnail"

eliotjordan (changes requested):
Looks good. Just a few minor issue in the spec.

app/services/migrations/reset_box_thumbnails.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  class ResetBoxThumbnails
+    def self.call(box_id:)
+      new(box_id: box_id).run
+    end
+
+    attr_reader :box_id
+    def initialize(box_id:)
+      @box_id = box_id
+      raise InvalidResourceTypeError unless box.is_a? EphemeraBox
```
eliotjordan, line 14: 💯 This is great. Exit early if not the correct resource type.

eliotjordan (approved):

PR #6103, opened 2023-11-30
Update collection_members_with_children task to run on a single collection

hackartisan:
refs pulibrary/dpul#1420


eliotjordan (approved):

sdellis (commented):
@hackartisan is there a reason the collection_id would be set as an ENV variable and not an argument to the rake task?

hackartisan:
@sdellis it's our convention to use env to pass arguments to rake because the argument passing is weird and hard to remember how to do. If you look around at more tasks you will see that all rake arguments are handled this way.

Issue #5747, opened 2023-04-12
IIIF presentation version 3


hackartisan:
For scanned maps we serve a version 3 manifest at a path like v3/manifest. For audio we just serve v3 manifests from the original path.

Should we choose one of these two approaches over the other?
Are we ready to serve all resources in v3? And if so, do we need to continue providing v2 for any resources?

Some discussion and possibly research is needed before we can proceed.

PR #6109, opened 2023-12-06
Make logo an array for iiif3 manifests 

### LINKED ISSUE #5046: `logo` property in IIIF 3 manifests should be an array.
This is stopping our AV manifests from rendering in other AV viewers (specifically https://samvera-labs.github.io/bloom-iiif/)

See https://github.com/IIIF-Commons/vault for a related issue.

Seanwarren77:
Closes #5046

### DIFF
--- a/app/services/manifest_builder/logo_builder_v3.rb
+++ b/app/services/manifest_builder/logo_builder_v3.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+class ManifestBuilder
+  class LogoBuilderV3
+    attr_reader :resource
+
+    ##
+    # @param [Resource] resource the Resource being viewed
+    def initialize(record)
+      @record = record
+      protocol = /localhost/.match?(Figgy.default_url_options[:host]) ? "http" : "https"
+      @host = "#{protocol}://#{Figgy.default_url_options[:host]}"
+    end
+
+    def apply(manifest)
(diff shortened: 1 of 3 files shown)


sdellis (commented):
Is there any reason why we're keeping `logo_builder.rb` and `manifest_service_locator.rb` around? Or if they should be kept around, how does Figgy know which version to use? I don't fully know how this works, so perhaps it's fine and I just need some context. 

Also looks like there's a coverage error: https://output.circle-artifacts.com/output/job/1a0d13ec-efa1-43b2-905f-3882a8d115ff/artifacts/0/coverage/index.html#30298fd38501e40873d22f00988a6d2fbe310db4

hackartisan:
> Is there any reason why we're keeping `logo_builder.rb` and `manifest_service_locator.rb` around? Or if they should be kept around, how does Figgy know which version to use? I don't fully know how this works, so perhaps it's fine and I just need some context.

@sdellis We still serve iiif2 manifests for most things, so we need to keep supporting those. There's a ticket to move everything over: https://github.com/pulibrary/figgy/issues/5747

> Also looks like there's a coverage error: https://output.circle-artifacts.com/output/job/1a0d13ec-efa1-43b2-905f-3882a8d115ff/artifacts/0/coverage/index.html#30298fd38501e40873d22f00988a6d2fbe310db4

Thank you, we'll look at that.

sdellis (approved):
LGTM 👍

PR #6110, opened 2023-12-06
Add a IIIF download option for cropped GeoTiffs

### LINKED ISSUE #5857: Allow download of cropped GeoTiff in UniversalViewer
The maps group would like to see a download of the original GeoTiff as well as the cropped version here: https://maps.princeton.edu/catalog/princeton-dc1257b365d 

The original tiff appears, but the cropped one isn't downloadable.

## Implementation Notes

We'd have to add this as another "Rendering" in the IIIF Manifest.

eliotjordan:
Closes #5857

### DIFF
--- a/app/services/manifest_builder/canvas_rendering_builder.rb
+++ b/app/services/manifest_builder/canvas_rendering_builder.rb
@@ -23,14 +23,7 @@ class ManifestBuilder
       return unless downloadable?
       manifest["rendering"] ||= []
       manifest["rendering"] << download_hash
-      # When given a MapSet with both ScannedMap tiffs and attached Raster
-      # Resources we attach a link to the Raster's primary file so users can
-      # download the GeoTiff from the viewer embedded in the catalog.
-      uncropped_download = geotiff_download_hash(label: "Download GeoTiff")
-      manifest["rendering"] << uncropped_download if uncropped_download
-      # Add a download link for cropped geotiffs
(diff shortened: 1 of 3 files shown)


yarn.lock
tpendragon, line 1: Why'd yarn.lock change?

app/services/manifest_builder.rb
```diff
@@ -635,6 +635,22 @@ def geotiff_child
         end
     end
 
+    def cropped_geotiff_child
```
tpendragon, line 638: Question: Would it make sense to move a bunch of this navigation logic to the wayfinder? So we can do like Wayfinder.for(resource).try(:cropped_geotiff_child)`?

hackartisan (commented):
Is the same change also needed in `manifest_builder_v3/canvas_rendering_builder.rb`? I'm not sure I have a great sense of what's serving v2 vs v3 manifests.

yarn.lock
eliotjordan: Accident. Was probably trying to get node-mapnik installed.

app/services/manifest_builder.rb
eliotjordan: Good idea. Done.

tpendragon (changes requested):
Code looks good, that yarn.lock is just still there.

tpendragon (approved):

PR #6113, opened 2023-12-06
Add CSV path param to  PreservationStatusReporter

### LINKED ISSUE #6108: Update PreservationStatusReporter to take an input

eliotjordan:
Closes #6108


tpendragon (changes requested):
Instead of a CSV, can it take a file that's newline delimited? Mostly because then we can just feed the previous failure list to it.

eliotjordan:
@tpendragon The output of the failure file will work here. It is a de-facto new line separated CSV.

PR #6115, opened 2023-12-07
Use a subclass for checking an existing list of ids for preservation

### LINKED ISSUE #6108: Update PreservationStatusReporter to take an input

hackartisan:
closes #6108 again


hackartisan:
close in favor of #6110 6117

PR #6117, opened 2023-12-11
Refactor preservation audit over full repository vs list of ids

### LINKED ISSUE #6108: Update PreservationStatusReporter to take an input

hackartisan:
And have the id check write to a different file
closes #6108


hackartisan (commented):
- [ ] Add a rake task that first delete (or moves?) the bad_resources.txt and since.txt so that a brand-new (not-resumed) full audit can run.
- [ ] Add checking first to see if there's a bad_resources_recheck.txt and if so use that as the input before rotating it.
- [ ] rotating a file should use its own created datetime, not the current timestamp, to reflect when that file was generated (not when I rotated it)

tpendragon (approved):

PR #6121, opened 2023-12-15
Back out requirement for specific bundler version

hackartisan:
We no longer need this since later versions of bundler first install
whatever bundler version the Gemfile.lock was generated with, then run
that to bundle. This will give us a little more flexibility on server
bundler versions.

refs pulibrary/princeton_ansible#3164


hackartisan:
deployed on staging

eliotjordan (approved):

Issue #6067, opened 2023-10-30
Re-enable cloud fixity checks


hackartisan:
Follow-on to https://github.com/pulibrary/figgy/pull/6066

Once everything is re-preserved according to #6024 and #5556, turn cloud fixity checks back on. Keep an eye on gcs egress and billing to ensure we're seeing expected levels. Make sure to remove the `# :nocov:` comments in `app/services/cloud_fixity`

Helpful links from when @tpendragon was troubleshooting:
- [RepairCloudFixityJob in logs](https://app.datadoghq.com/logs?query=%22Performed%20RepairCloudFixityJob%22%20&analyticsOptions=%5B%22line%22%2C%22dog_classic%22%2Cnull%2Cnull%5D&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=sliding&stream_sort=desc&viz=timeseries&from_ts=1697919144846&to_ts=1698523944846&live=true)
- [CloudFixityJob in logs](https://app.datadoghq.com/logs?query=%22CloudFixityJob%22%20&analyticsOptions=%5B%22line%22%2C%22dog_classic%22%2Cnull%2Cnull%5D&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=sliding&stream_sort=desc&viz=timeseries&from_ts=1697919144846&to_ts=1698523944846&live=true)
- [billing report](https://console.cloud.google.com/billing/0144DF-195EB5-AFAA02/reports;grouping=GROUP_BY_SKU;projects=pulibrary-figgy-storage-1;products=services%2F95FF-2EF5-5EA1?authuser=1&project=pulibrary-figgy-storage-1)
- [gcs monitoring console](https://console.cloud.google.com/storage/monitoring?authuser=1&project=pulibrary-figgy-storage-1) - especially ReadObject calls and data egress.

eliotjordan:
@tpendragon This ticket complete. I didn’t mean to close on Friday, but I ended up finishing it.

eliotjordan:
Yeah, I noticed that the CheckFixityJobs were acting weird again. Tons of RepairFixityJobs
we’re being queued when I looked at it last.

hackartisan:
Thoughts / ideas generated by team:
- Do a really thorough code read
- Is there a way to replicate in development so that we don't spend hundreds of dollars every time we try to troubleshoot?
- Get some example resources that are getting queued for repair that shouldn't, then look closely at those resources.
[...]

PR #6149, opened 2024-01-08
Prevent infinite repair loop when FileSet is not preserved.

tpendragon:
Work towards #6067 

I'm not totally sold on this - I might work on adding some way to ensure that repairing doesn't happen multiple times? I'm not sure how though. Maybe store a count somewhere?


tpendragon:
@hackartisan It was testing a file set without a parent before, which was an orphan and so wouldn't preserve, so I had to give it a parent

eliotjordan (approved):

PR #6156, opened 2024-01-12
Save characterization failure messages 

### LINKED ISSUE #6059: Characterization failures should record an error_message on FileSets.
When things fail characterization they never leave processing and likely will never succeed, and the derivative health report should show that it's failed so they'll know they should reingest that item.

Right now some characterizers error and some ignore them (imagemagick.) Characterizers only have persisters, not ChangeSetPersisters, and the code in there is odd (it has instance methods that change their return value based on an instance variable that changes in a loop.)

eliotjordan:
Closes #6059


app/characterization_services/mediainfo_characterization_service.rb
```diff
@@ -69,61 +75,6 @@ def valid?
 
   private
 
-    # Null Object modeling metadata values from MediaInfo.from
-    # @see MediaInfo::Tracks
-    class NullTracks
```
eliotjordan, line 74: Is there a reason to keep NullTracks? It just silently passes the file on to derivative generation after MediaInfo fails.

tpendragon: If tests don't fail let's keep it out.

tpendragon (approved):
Great work! I think this'll improve the workflow quite a lot.

PR #6165, opened 2024-01-18
Don't use buffered persister in CharacterizationJob so error messages are saved

### LINKED ISSUE #6059: Characterization failures should record an error_message on FileSets.

eliotjordan:
- Characterization failure messages were not being saved on the FileSet when run in production. The issue was the buffered persister which rolled back once the job errored.

- Also removes troublesome and unnecessary ExternalMetadataService test. 

Closes #6059 

Working in staging:
Screenshot 2024-01-18 at 1 38 21 PM

### DIFF
--- a/spec/derivative_services/external_metadata_derivative_service_spec.rb
+++ b/spec/derivative_services/external_metadata_derivative_service_spec.rb
@@ -46,16 +46,9 @@ RSpec.describe ExternalMetadataDerivativeService do
     end
   end
 
-  before do
-    allow(EventGenerator::GeoblacklightEventGenerator).to receive(:new).and_return(event_generator)
-    allow(event_generator).to receive(:record_updated)
-  end
-
-  it "extracts metadata from the file into the parent resource and triggers an update event", rabbit_stubbed: true do
+  it "extracts metadata from the file into the parent resource" do
     parent = query_service.find_by(id: parent_resource.id)
     expect(parent.title).to eq ["China census data by county, 2000-2010"]
     expect(parent.visibility).to eq ["open"]
-    # Twice for VectorResource
-    expect(event_generator).to have_received(:record_updated).exactly(2).times
   end
 end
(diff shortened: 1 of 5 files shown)


spec/derivative_services/external_metadata_derivative_service_spec.rb
```diff
@@ -46,16 +46,9 @@
     end
   end
 
-  before do
-    allow(EventGenerator::GeoblacklightEventGenerator).to receive(:new).and_return(event_generator)
-    allow(event_generator).to receive(:record_updated)
-  end
-
-  it "extracts metadata from the file into the parent resource and triggers an update event", rabbit_stubbed: true do
+  it "extracts metadata from the file into the parent resource", rabbit_stubbed: true do
```
hackartisan, line 49: Do we still need rabbit_stubbed?

hackartisan (changes requested):

spec/derivative_services/external_metadata_derivative_service_spec.rb
eliotjordan: Nope. Good catch. Pushed up the change.

hackartisan (approved):

hackartisan:
I don't totally understand why we wanted the buffered persister to begin with. @tpendragon do you have insight on this?

tpendragon:
> I don't totally understand why we wanted the buffered persister to begin with. @tpendragon do you have insight on this?

It's probably copied from the derivative service, which had a buffer so it'd have a transaction and if the second derivative service failed it'd roll back the first one. In the case of characterization I don't think we ever use two at once?

PR #6193, opened 2024-02-05
Attaching captions ADR

### LINKED ISSUE #6185: Determine and Document How Transcripts Attach to Resources
## Use Case

As a Figgy staff member I should be able to ingest a video associated with a resource (ScannedResource or EphemeraFolder), mark it complete, add subtitles to it later, and it should render in DPUL, the Catalog, and Finding Aids so that researchers can view that video. This means I should be able to add subtitles later which will then make it viewable to the public. (It has to preserve, even if it doesn't have subtitles yet, and I have to be able to add subtitles later.)

## Acceptance Criteria

An ADR which states how a transcript is attached to a Resource such that a video can have multiple transcripts, and a resource can have multiple videos that each have their own transcript. This will be accomplished via a Design Ensemble.

## Useful Info

There's a http://pcdm.org/use#Transcript

## Options

1. FileMetadata with a Use of "Transcript" which is attached to the FileSet that has the video file.
    * We'd have to develop an upload/overwrite UI and figure out preservation of that transcript.
2. FileSet which has a property that points to the FileSet it's a transcript for.
    * We'd have to hide the FileSet from the Structure/Order/File Manager and figure out the UI to add it.
3. A "transcript" property on the FileSet which stores the text of the full transcript in the database.
    * These might be quite big, it might slow down query of those FileSets.
    * We'd have to figure out an Upload/Overwrite UI, but preservation would just happen.

tpendragon:
Closes #6185

### DIFF
--- a/architecture-decisions/0013-captions-attach.md
+++ b/architecture-decisions/0013-captions-attach.md
@@ -8,19 +8,9 @@ Accepted
 
 ## Context
 
-We have the following use cases for video:
+Our use cases for Video can be found in our [use case documentation](/docs/ux/use_cases.md#videos).
 
-As Princeton University I should only allow the public to view videos that have subtitles which can render in the viewer both to legally protect ourselves and more importantly to meet our goals of accessibility. We want (.VTT) files for transcription subtitles so that we can standardize and migrate if need be.
-
-As a Figgy staff member I should be able to bulk ingest several videos, each with their own resource along with their captions so that I can have a vendor mass digitize materials and then ingest them later.
-
-As a Figgy staff member I should be able to ingest several videos to one resource, each with their own captions, and the viewer should render them with a table of contents so that I can display multiple videos in a single "box" of content in the archives.
-
-As Princeton University I want to preserve the captions along with the video so that if Figgy's data is ever lost I can restore them.
-
-As a researcher I want to have multiple captions each in their own language and users should be able to switch between them so that I can provide translations in addition to transcription.
-
-As such, videos in Figgy need a 1-N relationship with closed caption files. We discussed a few options:
+They point to videos in Figgy needing a 1-N relationship with closed caption files. We discussed a few options:
 
 1. FileSets have an extra FileMetadata users can upload which contains the captions, the label of the caption, and the language the caption is for.
 2. FileSets have a descriptive property that contains many nested `Caption` models.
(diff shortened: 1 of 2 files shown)


architecture-decisions/0013-captions-attach.md
```diff
@@ -0,0 +1,38 @@
...
+
+As Princeton University I should only allow the public to view videos that have subtitles which can render in the viewer both to legally protect ourselves and more importantly to meet our goals of accessibility. We want (.VTT) files for transcription subtitles so that we can standardize and migrate if need be.
+
+As a Figgy staff member I should be able to bulk ingest several videos, each with their own resource along with their captions so that I can have a vendor mass digitize materials and then ingest them later.
+
+As a Figgy staff member I should be able to ingest several videos to one resource, each with their own captions, and the viewer should render them with a table of contents so that I can display multiple videos in a single "box" of content in the archives.
+
+As Princeton University I want to preserve the captions along with the video so that if Figgy's data is ever lost I can restore them.
+
+As a researcher I want to have multiple captions each in their own language and users should be able to switch between them so that I can provide translations in addition to transcription.
```
hackartisan, line 21: Do we need something around a "default" caption? Or will captions by off by default?

architecture-decisions/0013-captions-attach.md
```diff
@@ -0,0 +1,38 @@
...
+We compared these options against ease of implementation, ease of reasoning, flexibility for anticipated use cases, and potential performance. We decided ease of reasoning was the most important of those.
+
+## Decisions
+
+FileSets will have an extra FileMetadata users can upload which contains the captions, the label of the caption, and the language the caption is for. This model matches our existing mental model of the way a FileSet contains an original file and all of its supporting files.
+
+## Consequences
+
+1. We have to develop functionality to upload and delete user-defined FileMetadata.
+1. Users will have to upload two files (the original file and the caption) if they delete a FileSet to re-upload it.
```
hackartisan, line 38:
```suggestion
1. If a user deletes a FileSet they will have to upload two files (the original file and the caption) to re-create it.
```

hackartisan (changes requested):
It's nice to get all these use cases here; I wonder how this documentation interacts with our attempt to collect use cases in https://github.com/pulibrary/figgy/blob/main/docs/ux/use_cases.md

This accurately reflects our decisions and is clear. I added a question and a suggestion in line.

tpendragon:
> It's nice to get all these use cases here; I wonder how this documentation interacts with our attempt to collect use cases in https://github.com/pulibrary/figgy/blob/main/docs/ux/use_cases.md

Is this a suggestion to also put them in `use_cases.md` or did you find them not useful here?

hackartisan (approved):

PR #6213, opened 2024-02-27
Add a factory for Videos w/ Transcripts

tpendragon:
This sets up future AV PRs by adding a factory for us to use in them.

### DIFF
--- a/app/services/file_appender.rb
+++ b/app/services/file_appender.rb
@@ -92,7 +92,7 @@ class FileAppender
       }.merge(
         file.try(:container_attributes) || {}
       )
-      file_set = FileSet.new
+      file_set = FileSet.new(attributes)
       change_set = ChangeSet.for(file_set)
       change_set.validate(attributes)
       change_set_persister.save(change_set: change_set)
--- a/app/services/file_appender.rb
+++ b/app/services/file_appender.rb
@@ -92,8 +92,9 @@ class FileAppender
       }.merge(
         file.try(:container_attributes) || {}
       )
-      file_set = FileSet.new(attributes)
+      file_set = FileSet.new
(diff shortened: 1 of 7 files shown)


app/services/file_appender.rb
```diff
@@ -94,6 +94,7 @@ def create_file_set(file_node, file)
       )
       file_set = FileSet.new(attributes)
       change_set = ChangeSet.for(file_set)
+      change_set.validate(attributes)
```
hackartisan, line 97: Does passing the attributes to new and then also to validate kind of set them twice?

Also I can't remember, did we add something in the change set persister that checks `valid?` before doing a save so that we can do things this way?

hackartisan (changes requested):
This defines captions as a transcript on an av resource. I know there are distinctions between captions and transcripts; I lose track of them. They're captions in video and transcripts in audio?

Thank you for thinking to add the factory. just some questions here.

tpendragon:
> This defines captions as a transcript on an av resource. I know there are distinctions between captions and transcripts; I lose track of them. They're captions in video and transcripts in audio?

I don't think we have any transcripts - transcripts aren't synced to the audio, captions (and subtitles) are.

I called them transcripts because the PCDM Use is Transcript (https://pcdm.org/2021/04/09/use#Transcript), but it did make me feel wrong.

app/services/file_appender.rb
tpendragon: > Does passing the attributes to new and then also to validate kind of set them twice?

It does set them twice - I originally only set it in `validate()`, but not every property that we set here is a ChangeSet property.

We don't check `valid?`, but we also didn't check that before either - this is mostly about making sure that the `files` virtual attribute gets through.

Issue #6181, opened 2024-01-25
WebVTT captions should be added to the HLS manifest so they play back in the Universal Viewer.


tpendragon:
Use Case: As Princeton University I should only allow the public to view videos that have subtitles which can render in the viewer both to legally protect ourselves and more importantly to meet our goals of accessibility. We want (.VTT) files for transcription subtitles so that we can standardize and migrate if need be.

## Acceptance Criteria

The viewer should allow users to select from multiple captions that are attached to the resource.

If it turns out we add this and UV doesn't render the captions we have some research to do.

## Sample VTT File

```file.vtt
WEBVTT

00:00.000 --> 00:03.000
Test Caption
```

tpendragon:
I did a little experimenting here:

It looks like in order to do this we need to generate a primary m3u8 playlist that points to a subtitle m3u8 and a video m3u8. I found https://hlsjs.video-dev.org/demo/ useful to test with.

I haven't got it working yet, but it's definitely more complicated than I expected.

tpendragon:
Good news: https://github.com/pulibrary/figgy/compare/fileset_with_caption...attach_vtt_hls has a proof of concept with captions that works in hlsjs.video-dev.org.

Bad news: It doesn't work in UV.

tpendragon:
I can confirm it renders in Clover.

PR #6216, opened 2024-02-28
Use a primary HLS manifest

tpendragon:
This adds a primary manifest in front of the AV manifests, as a place we
can put transcripts.

This also refactors where generating HLS manifests happens, so we can
add to that logic without changing DownloadsController all the time.

Work towards https://github.com/pulibrary/figgy/issues/6181

### DIFF
--- a/app/values/hls_manifest.rb
+++ b/app/values/hls_manifest.rb
@@ -1,7 +1,13 @@
 # frozen_string_literal: true
+# Class responsible for converting an HLS manifest (.m3u8) file from ffmpeg to
+# include an auth token in its file references, if appropriate.
 class HlsManifest
+  # Factory method for returning the different kinds of manifests. If asked for
+  # a "stream" then it generates a manifest dynamically, otherwise it simply
+  # returns the manifest referenced by file_metadata, mutated if needed.
+  # @return [#to_s] Object whose #to_s method returns an HLS Manifest.
   def self.for(file_set:, file_metadata:, as: nil, auth_token: nil)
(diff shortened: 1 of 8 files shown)


app/values/hls_manifest.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+class HlsManifest
+  def self.for(file_set:, file_metadata:, as: nil, auth_token: nil)
+    return unless file_metadata.mime_type.first.to_s == "application/x-mpegURL"
```
hackartisan, line 4: `return unless file_metadata.hls_manifest?`?

app/values/hls_manifest.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+class HlsManifest
```
hackartisan, line 2: Can we have class-level documentation here to clarify how and why it's different from the HlsManifest::Primary?

app/values/hls_manifest.rb
hackartisan: Or should we change the name of that method to `hls_file?`?

hackartisan (changes requested):
This looks like a nice improvement.

app/values/hls_manifest.rb
tpendragon: Thank you!

app/values/hls_manifest.rb
tpendragon: Added some docs.

hackartisan: thank you

PR #6217, opened 2024-02-28
Attach caption HLS manifest to primary manifest. 

tpendragon:
This generates an HLS manifest for caption files (VTT) and associates them with the primary manifest.

Closes https://github.com/pulibrary/figgy/issues/6181

### DIFF
--- a/app/nested_resources/file_metadata.rb
+++ b/app/nested_resources/file_metadata.rb
@@ -28,6 +28,9 @@ class FileMetadata < Valkyrie::Resource
   # PDF Metadata
   attribute :page_count, Valkyrie::Types::Integer
 
+  # Caption Metadata
+  attribute :caption_language, Valkyrie::Types::String.default("eng")
+
   # preservation attributes
   # ID of the object this node is a preservation copy of. A PreservationObject's
   # binary_node (which is a FileMetadata object) uses this value to point to a
--- a/app/nested_resources/file_metadata.rb
+++ b/app/nested_resources/file_metadata.rb
@@ -29,7 +29,7 @@ class FileMetadata < Valkyrie::Resource
   attribute :page_count, Valkyrie::Types::Integer
 
   # Caption Metadata
-  attribute :caption_language, Valkyrie::Types::String.default("eng")
+  attribute :caption_language, Valkyrie::Types::String.optional
 
   # preservation attributes
   # ID of the object this node is a preservation copy of. A PreservationObject's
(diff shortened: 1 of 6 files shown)


app/values/hls_manifest.rb
```diff
@@ -7,11 +7,12 @@ class HlsManifest
...
   def self.for(file_set:, file_metadata:, as: nil, auth_token: nil)
-    return unless file_metadata.hls_manifest?
-    if as == "stream"
+    return unless file_metadata.hls_manifest? || file_metadata.caption?
+    return new(file_set: file_set, file_metadata: file_metadata, auth_token: auth_token) unless as == "stream"
+    if file_metadata.hls_manifest?
       Primary.new(file_set: file_set, file_metadata: file_metadata, auth_token: auth_token)
     else
-      new(file_set: file_set, file_metadata: file_metadata, auth_token: auth_token)
+      Caption.new(file_set: file_set, file_metadata: file_metadata, auth_token: auth_token)
```
hackartisan, line 15: This section might be more clear to me if it were like:
```
return Caption.new... if file_metadata.caption?
return unless file_metadata.hls_manifest?
if as == "stream"
  Primary...
else
  new...
end
```

app/nested_resources/file_metadata.rb
```diff
@@ -28,6 +28,9 @@ class FileMetadata < Valkyrie::Resource
   # PDF Metadata
   attribute :page_count, Valkyrie::Types::Integer
 
+  # Caption Metadata
+  attribute :caption_language, Valkyrie::Types::String.default("eng")
```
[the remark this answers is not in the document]
tpendragon: Oh, yeah, I guess so. 🤔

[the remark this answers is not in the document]
hackartisan: I don't think I like this. is there a downside to having it be nil? We'd just have to set its default every time we set it in code?

app/values/hls_manifest.rb
```diff
@@ -7,11 +7,17 @@ class HlsManifest
...
     return unless file_metadata.hls_manifest?
+    # IIIF Manifests link to a "primary" HLS manifest, which contains links to the caption
+    # manifest and the video manifest. Build it dynamically because it's easier.
     if as == "stream"
       Primary.new(file_set: file_set, file_metadata: file_metadata, auth_token: auth_token)
     else
-      new(file_set: file_set, file_metadata: file_metadata, auth_token: auth_token)
+      # If not asked for the primary manifest, just render the video HLS manifest
+      # ffmpeg created (with auth_tokens if appropriate)
+      new(file_set: file_set, file_metadata: file_metadata, auth_token: auth_token) unless as == "stream"
```
hackartisan, line 20: redundant `unless` here

hackartisan (changes requested):
one small typo from the code reorganization

app/values/hls_manifest.rb
tpendragon: Oof, thank you. Postfix conditionals are sneaky. 🥷

hackartisan (approved):
🎉

Issue #6179, opened 2024-01-25
Attach WebVTT captions to a video FileSet


tpendragon:
Use Case: As Princeton University I should only allow the public to view videos that have subtitles which can render in the viewer both to legally protect ourselves and more importantly to meet our goals of accessibility. We want (.VTT) files for transcription subtitles so that we can standardize and migrate if need be.

## Acceptance Criteria

I should be able to do whatever we decided in #6184 which will result in a data structure we decided in #6185.

1. From the File Manager I should be able to click "Attach Caption" and it will take me to a new page to attach a transcript. That page will have a Label, Language, and File input, which when submitted will create a new WebVTT

## Question

Can we split this ticket up realistically?

## Sample VTT File

```file.vtt
WEBVTT

00:00.000 --> 00:03.000
Test Caption
```
(The below should say "Attach Missing Captions")
![FileManager](https://github.com/pulibrary/figgy/assets/609368/56644a38-46eb-424b-a89a-ed5ad117aa9f)

![FileSetShow](https://github.com/pulibrary/figgy/assets/609368/396b0a11-cf0f-4aba-8ef6-87e1278d9e13)
## Attach Caption Edit Page:
Notes: 
- Language information pop up will have info about having multiple files in the same or different languages, and that the accessibility requirement is only met if there is at least one caption in the original language of the audio. 
- If multiple captions in the same language are attached, the most recent will be used. 
- The filename will be renamed according to some TBD convention (e.i., `en_english_original-language.vtt`)

![Attach-Caption](https://github.com/pulibrary/figgy/assets/609368/001ffbd8-b36b-441a-a1e1-1ff3c02607f9)

tpendragon:
To do
- [ ] Add original language check box
- [ ] Add delete of captions
- [ ] Change caption info box to black, add a tag for language and if it's in the original language
- [ ] Add link to File Manager to upload a caption

PR #6221, opened 2024-02-29
Attach Caption UI

tpendragon:
Work towards #6179 

Screen Shot 2024-02-29 at 1 23 30 PM

### DIFF
--- a/app/controllers/file_metadata_controller.rb
+++ b/app/controllers/file_metadata_controller.rb
@@ -6,12 +6,15 @@ class FileMetadataController < ApplicationController
     authorize! :update, @file_set
     @change_set = ChangeSet.for(FileMetadata.new, change_set_param: change_set_param)
     @change_set.validate(resource_params)
-    # TODO: Add validation check here, re-render form.
-    ingestable_file = @change_set.to_ingestable_file
-    file_set_change_set = ChangeSet.for(file_set)
-    file_set_change_set.files = [ingestable_file]
-    change_set_persister.save(change_set: file_set_change_set)
-    redirect_to solr_document_path(file_set.id.to_s)
+    if @change_set.valid?
+      ingestable_file = @change_set.to_ingestable_file
+      file_set_change_set = ChangeSet.for(file_set)
+      file_set_change_set.files = [ingestable_file]
+      change_set_persister.save(change_set: file_set_change_set)
+      redirect_to solr_document_path(file_set.id.to_s)
+    else
+      render :new
+    end
   end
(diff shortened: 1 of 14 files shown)


app/controllers/file_metadata_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+class FileMetadataController < ApplicationController
+  delegate :query_service, to: :change_set_persister
+
+  def new
+    authorize! :update, file_set_change_set.resource
+    @change_set = ChangeSet.for(FileMetadata.new, change_set_param: change_set_param)
+    render "new"
```
hackartisan, line 8: Does this have to be explicit because of the routing base? Normally it would be inferred I think.

app/controllers/file_metadata_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def create
+    authorize! :update, file_set_change_set.resource
+    @change_set = ChangeSet.for(FileMetadata.new, change_set_param: change_set_param)
+    @change_set.validate(resource_params)
+    if @change_set.valid?
+      ingestable_file = @change_set.to_ingestable_file
```
hackartisan, line 16: I like this

spec/controllers/file_metadata_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+            change_set: "Caption",
+            file: fixture_file_upload("files/caption.vtt", "text/vtt"),
+            caption_language: "eng"
+          }
+        }
+
+        expect(response).to redirect_to solr_document_path(file_set.id.to_s)
+        file_set = ChangeSetPersister.default.query_service.find_by(id: file_set.id)
+        expect(file_set.captions.length).to eq 1
+        expect(file_set.captions.first.file_identifiers).to be_present
```
hackartisan, line 74: maybe test that it saves the change set type?

hackartisan (changes requested):
this looks great just one suggestion

app/controllers/file_metadata_controller.rb
tpendragon: Thank you! I'd completely forgotten 😂

spec/controllers/file_metadata_controller_spec.rb
tpendragon: Oh wow, good catch!

hackartisan (approved):

PR #6224, opened 2024-02-29
Add original_language_caption property to Caption form.

tpendragon:
Work towards #6179

Screen Shot 2024-02-29 at 3 14 10 PM


hackartisan (changes requested):
This looks good but I don't think it rebased right

hackartisan (approved):

PR #6236, opened 2024-03-06
Add a rake task for migrating source metadata identifiers from CSV.

tpendragon:
Work towards #6232


tpendragon:
> Question: Does this also refresh the remote metadata?

Yes, that's why I had to stub the MMS id.

hackartisan (approved):
Thank you!

PR #6238, opened 2024-03-07
Configure redis URL for health check.

tpendragon:


config/initializers/redis_config.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+module RedisConfig
```
hackartisan, line 2: Use in sidekiq initializer?

hackartisan (changes requested):

config/initializers/redis_config.rb
tpendragon: Good catch, done.

eliotjordan (approved):

hackartisan (approved):

PR #6241, opened 2024-03-08
Add media_type to the Embed GraphQL type.

### LINKED ISSUE #6233: Resource GraphQL response should say if it's a video
We'll need to know if a resource is going to be playing video so we can use the new video viewer required by #6222.

The uv_manager already does a GraphQL query using the `resource()` query - I think we could add it as a property of the `embed`, then only use that property in Figgy. Something like `embed.media_type` where the values can be `Image, Video, Audio`? Just a thought.

tpendragon:
Lets us query for videos.

Closes #6233

### DIFF
--- a/app/services/embed.rb
+++ b/app/services/embed.rb
@@ -21,12 +21,14 @@ class Embed
       {
         type: build_type,
         content: build_content,
+        media_type: build_media_type,
         status: "authorized"
       }
     else
       {
         type: nil,
         content: nil,
+        media_type: nil,
         status: unauthorized_status
       }
     end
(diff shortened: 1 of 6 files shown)


app/services/embed.rb
```diff
@@ -63,6 +65,16 @@ def build_type
       end
     end
 
+    def build_media_type
+      if !viewer_enabled?
+        "Download"
+      elsif file_set&.video?
+        "Video"
+      else
+        "Image"
```
hackartisan, line 74: I think this will result in an audio resource having `media_type` "Image". It won't be a problem technically but it is a bit incongruous, what do you think?

hackartisan (changes requested):

app/services/embed.rb
tpendragon: Yeah I just didn't add it because I didn't have a use case for it, but I can.

tpendragon:
@hackartisan Added Audio

hackartisan (approved):

PR #6244, opened 2024-03-11
Extract vtt file language during bulk ingest

### LINKED ISSUE #6180: Bulk Ingest should accept WebVTT captions via a naming standard.
## Use Case

As a Figgy staff member I should be able to bulk ingest several videos, each with their own resource along with their captions so that I can have a vendor mass digitize materials and then ingest them later.

## Acceptance Criteria

When bulk ingesting a folder if there is a "filename.mp4" and a "filename.vtt" in the same folder, it should ingest however we decided in #6185. It should support having multiple transcriptions for a single mp4. Let's try "filename-English (Original Language).vtt" for a transcript thats label is "English (Original Language)"

hackartisan:
- Save the vtt's caption language
- Save the original language flag for the vtt
- Allow multiple vtt files of different languages, validate language value
- Document video / caption bulk ingest

~Note that this PR changes the behavior of the controlled vocabulary for language. Previously it wasn't really functioning as a controlled vocabulary (any value was accepted) and `include?` always returned true. If we have functionality depending on that behavior it will break.~ Update: I changed this so it only validates the iso 639 for caption files.

closes #6180


app/services/bulk_ingest_service/bulk_file_path_converter.rb
```diff
@@ -30,24 +30,53 @@ def ingestable_files
...
+        caption_language: infer_language(vtt_path),
+        original_language_caption: infer_original_language(vtt_path)
+      )
+      change_set.to_ingestable_file
+    end
+
+    # mimic the UploadedFile expected by CaptionChangeSet
+    def uploaded_file_struct(vtt_path)
+      Struct.new(
+        :path, :content_type, :original_filename
```
tpendragon, line 65: Clever! Can you use `IngestableFile.new(file_path: vtt_path, mime_type: "text/vtt", original_filename: vtt_path.basename.to_s)`? It's a little odd, but one of the use cases for IngestableFile was to handle cases where we don't have one of those Rails file uploads.

tpendragon (changes requested):
Just a small suggestion - this looks great!

app/services/bulk_ingest_service/bulk_file_path_converter.rb
hackartisan: Sure thing. It's just `IngestableFile` all the way down.

tpendragon (approved):

PR #6246, opened 2024-03-11
Render V3 seeAlso property in av manifests

### LINKED ISSUE #6245: seeAlso is not an Array for ScannedResources with video
Work towards #6243

eliotjordan:
Closes #6245 

- Render V3 seeAlso property in av manifests
- Change method name to reflect both audio and video


hackartisan (commented):
2 v3 codepaths strikes again!

hackartisan (approved):

PR #6247, opened 2024-03-11
Make HLS manifest caption default, name values accurate

### LINKED ISSUE #6231: Make sure HLS manifest caption properties are correct now that we have multiple languages
Make sure we set DEFAULT to the caption which is the same as the source language, and that the label is reasonable.

hackartisan:
closes #6231

### DIFF
--- a/app/decorators/caption_decorator.rb
+++ b/app/decorators/caption_decorator.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+class CaptionDecorator < Valkyrie::ResourceDecorator
+  def caption_label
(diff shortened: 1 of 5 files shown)


tpendragon (changes requested):
I have no idea why CI would be failing that way, but there it is. Code looks great.

hackartisan:
> I have no idea why CI would be failing that way, but there it is. Code looks great.

It must be because I updated the factory and now it has 2 caption files. I'll update the test.

tpendragon (approved):

PR #6254, opened 2024-03-12
Add Clover viewer for video support

### LINKED ISSUE #6222: Video viewer should render captions
We're adding captions to the HLS manifest now, but Universal Viewer doesn't render them.

We need to identify a new viewer and get it loading for Video.

Options:

Clover: https://samvera-labs.github.io/clover-iiif/
Ramp: https://ramp.avalonmediasystem.org/

We're gonna do Clover.

eliotjordan:
Closes #6222 

- Add Clover viewer for video support
- Lint uv manager file and test
- Add Clover canvas size workaround

Screenshot 2024-03-12 at 2 42 46 PM


app/javascript/viewer/uv_manager.js
```diff
@@ -111,6 +126,15 @@ export default class UVManager {
     this.cdlTimer.initializeTimer()
   }
 
+  createClover () {
+    this.uvElement.show()
+    const uvElement = document.getElementById('uv')
+    const root = ReactDOM.createRoot(uvElement)
+    // Setting canvasHeight is a workaround until this issue is resolved:
+    // https://github.com/samvera-labs/clover-iiif/issues/179
+    root.render(React.createElement(Viewer, { iiifContent: this.manifest, options: { informationPanel: { open: false, renderToggle: false }, background: 'white', withCredentials: true, showTitle: false, showIIIFBadge: false, canvasHeight: uvElement.clientHeight } }))
```
tpendragon, line 135: This workaround is nice, but it breaks in the case that we have two videos - the video selector is pushed below the fold. :(

tpendragon (changes requested):
I'm so excited to have this in and tested, but I don't think the canvas fix is gonna work. We might have to fiddle with something else, or find a way to see if there's multiple videos and adjust the height based on that?

tpendragon:
I threw together https://github.com/samvera-labs/clover-iiif/pull/185, I'm hoping we'll get it in and that'll make this easier for us.

app/javascript/viewer/uv_manager.js
eliotjordan: Work around improved.

tpendragon (approved):

PR #6274, opened 2024-03-18
Add muxdelay 0 to ffmpeg HLS generation.

tpendragon:
I very carefully crafted some subtitles for a demo video, uploaded them, and noticed subtitles were out of sync! According to https://stackoverflow.com/questions/61835223/ffmpeg-burnt-in-subtitles-out-of-sync-when-converting-to-hls ffmpeg adds a delay to AV when creating HLS - and sure enough, when I add this argument everything's in sync.


hackartisan (approved):
Looks good and thank you for the description of what happened and the pointer to docs!

PR #6277, opened 2024-03-19
Generate Checksum for captions.

### LINKED ISSUE #6276: Captions not saving a checksum to its FileMetadata node
This is preventing the resource from preserving.

tpendragon:
This should handle future preservation targets as well.

I moved Characterize to after_save_commit so we'd have access to `post_save_resource` in its handlers - I couldn't find a reason for it to not be there.

Closes #6276


sdellis (approved):
Looks good to me! 👍

PR #6293, opened 2024-03-23
Remove parent from path

### LINKED ISSUE #2566: Get rid of /parent in the URL
We can look up parents and display breadcrumbs anyway, so it serves no purpose.

tpendragon:
Closes #2566


hackartisan (approved):
This looks great. We could potentially rewrite the old paths in routes.rb in case people have things bookmarked? But it seems unlikely that people use figgy that way.

eliotjordan (approved):

PR #6299, opened 2024-03-27
Cleanup mosaic.json files on resource delete

### LINKED ISSUE #6285: Clean up mosaic.json files on resource delete
Add ChangeSetPersister logic to remove mosaic.json files on AWS when a mosaic resource is deleted.

eliotjordan:
Closes #6285

### DIFF
--- a/app/change_set_persisters/change_set_persister.rb
+++ b/app/change_set_persisters/change_set_persister.rb
@@ -63,6 +63,7 @@ class ChangeSetPersister
       before_delete: [
         CreateDeletionMarker,
         CleanupPDFDerivatives,
+        CleanupMosaic,
         CleanupFiles,
         CleanupStructure,
         DeleteReferenced::Factory.new(property: :member_of_vocabulary_id),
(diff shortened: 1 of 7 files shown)


app/jobs/delete_mosaic_job.rb
```diff
@@ -0,0 +1,31 @@
...
+    @resource_id = resource_id
+    storage_adapter.delete(id: mosaic_id)
+  rescue TileMetadataService::Error
+    false
+  end
+
+  private
+
+    def resource
+      query_service.find_by(id: Valkyrie::ID.new(resource_id))
```
tpendragon, line 17: Won't this fail if it's backgrounded? The resource will have been deleted.

tpendragon (changes requested):
I think there's a race condition that won't show up in tests with the inline adapter. To get it to happen I think you'd have to queue jobs and then run them after or something.

app/jobs/delete_mosaic_job.rb
eliotjordan: Ah. Right. It will. I've refactored the job to take the storage adapter id as a parameter instead.

PR #6311, opened 2024-04-01
Wrap pdf generation in a job, call it synchronously

hackartisan:
There are also a couple of refactors in here; it may be helpful to view the commits separately.

- wrap pdf generation in a job, call it synchronously
- Remove change_set param from pdf and structure paths
- Generate pdf from the resource id
- Skip failing tests
- Move controller pdf method to a concern


tpendragon (approved):
This is great. The concern will make the next step much easier too.

Issue #6281, opened 2024-03-19
Create a loading page for PDFs


tpendragon:
Created out of #4876.

Spike: https://github.com/pulibrary/figgy/compare/main...pdf-actioncable-spike

Steps written out discovered from the spike:

* Call the PDF job asynchronously. (Created in #6278)
* Create a `base/pdf.html.erb` template that renders a loading page.
* Make `ScannedResourcesController#pdf` render `base/pdf.html.erb` if the PDF isn't generated already. (POC: https://github.com/pulibrary/figgy/commit/b5145e62be542e288178574a21e065b3274da256#diff-8e0ad8bd2a2f574caf3ce37dd016ada26ec0185ac3bfccf2c308d9aa0a344c1eR52-R58) (If we put this behind a feature flipper, we might be able to merge this if we wanted.)
* Write an ActionCable consumer in JS. Subscribe when some HTML on `base/pdf.html.erb` is present. (POC: https://github.com/pulibrary/figgy/commit/b5145e62be542e288178574a21e065b3274da256#diff-6b4397d53010efb20a9c5c7985af1e0987f53d155049fa9784829d7b5cabdf1f)
* When the consumer receives a message that says `{ pct_complete: 100, url: "/downloads/<scanned_resource_id>/file/<pdf_file>" }`, redirect to the URL.
* Write an ActionCable channel in Ruby which subscribes the above consumer to the "pdf_download_<scanned_resource_id>" topic. (POC: https://github.com/pulibrary/figgy/compare/pdf-actioncable-spike#diff-b5e90984d5a6b7a0d0b2734902dcb20a979056dc9e4d432d9c84717e9fdd44d0R2-R6)
* When the consumer subscribes, send a message with the initial status - 100 if the scanned resource already has a PDF attached, 0 if not.
* When the PDF is done generating in the job, send a message that the `pct_complete` is 100. Maybe put this in its own messaging class (POC: https://github.com/pulibrary/figgy/compare/pdf-actioncable-spike#diff-145e2f4bc32330c795f7290067dc6e9b1dcbd178b74b24d876b3262074d64679R26)
* Leverage our shared cache to prevent multiple jobs spawning for multiple clicks. Do something like ```Rails.cache.fetch("pdf_download_job_id_#{resource.id}", expires_in: 30.minutes) { PDFJob.perform_later(bla).id }```
* Configure the job to be in a "realtime" queue
* Make `base/pdf.html.erb` look like the wireframe in #6280 

[...]

PR #6317, opened 2024-04-02
Add layout for PDF loading page.

tpendragon:
Work towards #6281

Screen Shot 2024-04-02 at 11 07 49 AM


hackartisan (commented):
What font size is this at? I feel like we have some space to make it sort of on the big side. It's hard to tell from the screenshot if it's already big though.

hackartisan (approved):
I like the larger font size

hackartisan (approved):

Issue #6329, opened 2024-04-04
East Asian mp3s not syncing to DPUL


hackartisan:
A set of MP3s were added to Figgy on Tuesday (2 days ago), under the "East Asian Library Tables of Contents" collection. See items 2 through 25 under: https://figgy.princeton.edu/?f%5Bmember_of_collection_titles_ssim%5D%5B%5D=East+Asian+Library+Tables+of+Contents&per_page=50&q=&sort=created_at_dtsi+desc

(Alternately, limit those results to audio file type)

These should be available under https://dpul.princeton.edu/ealtoc but they are not.  Other items added to this collection 
have synced correctly.

Additional info: The access for the MP3s is set to Princeton-only.

reported by @tventimi

hackartisan:
Here is the relevant error in honeybadger: https://app.honeybadger.io/projects/51733/faults/106005729

hackartisan:
There are 9 items still that aren't syncing:
https://figgy.princeton.edu/catalog/f9ff4f4b-f253-4609-b9ed-817cbb02d9f1
https://figgy.princeton.edu/catalog/26c4ec6f-2db0-42cd-98fe-c8dec1a3d73c
https://figgy.princeton.edu/catalog/3f2dff27-a690-453e-b1d6-858978c0a5fb
https://figgy.princeton.edu/catalog/d1d6749f-f39b-49f1-907a-4ad2cd6a0ce6
https://figgy.princeton.edu/catalog/5009cb79-3b8b-470f-a7f4-ccfbee5715dd
https://figgy.princeton.edu/catalog/19523861-7fb8-459a-9b72-7ea6e64d5566
https://figgy.princeton.edu/catalog/a06cb0ee-aeb4-4fc6-9213-3aadd86497e4
https://figgy.princeton.edu/catalog/171d9bdc-4797-4db4-af59-bbfd82df988b
https://figgy.princeton.edu/catalog/126ebf02-ccd5-4443-ae34-936050d9c91b

hackartisan:
potentially related: https://app.honeybadger.io/projects/51733/faults/97930568

A head request to figgy for the manifest is timing out?

think about whether we should monitor this.

hackartisan:
re: head request to manifest timing out. @tpendragon noted that even cached manifests can take a long time to load due to permission-checking.

tpendragon:
Missing DPUL document IDS:

```
["808aa225-1f7a-4da9-a31c-b89483cc5c48",
[...]

PR #6335, opened 2024-04-12
Optimize and test child collection members migration task

hackartisan:
refs #6329


tpendragon (approved):

tpendragon:
Checking if the member of a collection has a parent is a nice optimization.

hackartisan:
@tpendragon totally! I copied it from something I found that you did 😂 https://github.com/pulibrary/dpul/issues/1433#issuecomment-1439277200

PR #6345, opened 2024-04-17
Upgrade Vue and Lux

### LINKED ISSUE #6249: Helenica Posters ephemera items not ingesting
### Summary

Please update Beth Haas once this issue is resolved.

Potentially a result of the jammy machine setup, ingest is not working after running the `lae:ingest_disk_files` task. At first it appeared that the new box simply needed to be added to known_hosts, but images haven't ingested on subsequent runs of the task.

The task is run on lib-proc4. Invocation is:
```
""ssh deploy@figgy-worker-prod2 "cd /opt/figgy/current; RAILS_ENV=production bundle exec rake lae:ingest_disk_files "'/mnt/diglibdata/pudl/lae_to_ingest/32101046816920'""
```

An example item that should be ingested after this run is barcode 32101118176260

### Impact 
This is an evergreen project and the ingest doesn't have a particular deadline, but this script is used for other ephemera loads and will become a work blocker moving forward.

### Priority recommendation

- [ ] asap
- [x] within the next 3 weeks
- [ ] PO will prioritize

This load script is used for ephemera projects and blocks work if it's broken.

### Expected behaviour

After running the script, images are ingested to the ephemera resources

### Actual behaviour

No result.

eliotjordan:
Closes #6249


app/javascript/entrypoints/application.js
```diff
@@ -1,58 +1,58 @@
...
+import setupAuthLinkClipboard from '@/figgy/auth_link_clipboard'
+import AjaxSelect from '@/components/ajax-select.vue'
+import { setupAjaxSelect, setupCocoonLinks } from '@/helpers/setup_ajax_select'
+import FileUploader from '@/components/file-uploader.vue'
+import Initializer from '@/figgy/figgy_boot'
+import LocalUploader from '@/components/local_uploader.vue'
 
-Vue.use(system)
-Vue.component('v-details', VueDetails)
+const app = createApp(
```
eliotjordan, line 20: Not sure if this is correct yet.

sdellis:
I had a hard time configuring vite [resolve.alias](https://vitejs.dev/config/shared-options.html#resolve-alias
). I took out the aliasing for imports and got it to work. If we want to re-implement import aliasing, we can do that in a future ticket, but it's no longer blocking the upgrade.

sdellis:
The [VueDetails](https://github.com/kim366/vue-details) dependency used in the [DirectoryPicker](https://github.com/pulibrary/figgy/blob/main/app/javascript/components/file_browser/directory_picker.vue#L8) does not work in Vue3 and has not been migrated. This is a blocker to the Vue2 -> Vue3 upgrade.

There are three ways to resolve this. 

1. [Implement using `refs`](https://stackoverflow.com/questions/77240675/vue-3-toggleevent-on-details-tag)
2. Implement locally (or fork) and [modify the render](https://github.com/kim366/vue-details/blob/master/src/index.js#L9-L15) code [so that render works in Vue3](https://stackoverflow.com/questions/65529646/why-the-render-h-function-cannot-use-component-name-directly-to-render-in-vue3-x)
3. Implement show/hide functionality without using the <details> tag and its native element state, [as in the Tree component](https://github.com/pulibrary/figgy/blob/main/app/javascript/components/Tree.vue#L117-L120).

sdellis:
Todo: The `itemLabel` field calls the `updateSingle` function `on:input`, but the item in items remains unchanged when [dispatched to the action](https://github.com/pulibrary/figgy/blob/main/app/javascript/components/OrderManagerFilesetForm.vue#L180). It should update the label/caption on every input event.

app/javascript/test/setup.js
```diff
@@ -1,9 +1,6 @@
-import Vue from 'vue'
 import _ from 'lodash'
-import system from 'lux-design-system'
+// import system from 'lux-design-system'
```
hackartisan, line 2: delete?

hackartisan (changes requested):
I found a number of places where there's commented-out code that I wonder if we could just delete.

Other than that, everything looks good. There's a test failure about snapshots that should be resolved. I'm deploying to staging so we can try it out.

hackartisan:
I tried this in staging. I messed with: 
* order manager
* structure manager
* datatables
* removing / adding a child from a MVW

all worked and looked good, in one case better than prod, which was overflowing in the order manager at a moderately zoomed-in screen size.

Issue #6353, opened 2024-04-18
Update omniauth and omniauth-cas gems


eliotjordan:
There are new versions of omniauth and omni-auth that will (hopefully) resolve a long-standing dependabot alert.

Issue #6352, opened 2024-04-18
Update iiif-manifest gem


eliotjordan:
Our IIIF manifest generation code does not work with the latest releases of the [iiif_manifest gem](https://github.com/samvera/iiif_manifest).

PR #6354, opened 2024-04-18
Update all gems

eliotjordan:
Updates gems and should close a number of dependabot alerts.

Gems that could not easily be updated:

1. iiif_manifest: https://github.com/pulibrary/figgy/issues/6352
2. omniauth and omniauth-cas: https://github.com/pulibrary/figgy/issues/6353


hackartisan (approved):
This looks good and I ran a bulk ingest on figgy staging as an additional check 👍🏻

Issue #6343, opened 2024-04-16
Generate a report of all resources that have an MMS-ID and an ARK that points to finding aids


tpendragon:
## Implementation Tips

You don't have to follow the redirect for the ARK, just see what the location header says when you get the ARK.

There's a Report Generator that can create CSVs from a query, might be useful here.

Part of #6262

## Success Criteria

A CSV is attached to this ticket, each row has the ID of the resource, the title, the MMS-ID, the ARK, and the finding aids URL the ARK points at.

## First Step

Write a query that gets all resources that have an MMS-ID.

Figure out how to stub webmock so it replicates a redirect.

hackartisan:
details of alma ids: https://pul-confluence.atlassian.net/wiki/spaces/ALMA/pages/1770185/Alma+System+Numbers

Also, here's a way to get the target via the ezid client: 
```
> ezid = Ezid::Identifier.find("ark:/88435/f1881r915")
I, [2024-04-18T09:19:48.047067 #36039]  INFO -- : EZID GetIdentifierMetadata -- success: ark:/88435/f1881r915
=> #<Ezid::Identifier id=ark:/88435/f1881r915>
> ezid.target
=> "http://findingaids.princeton.edu/collections/MC019/c01058"
```

hackartisan:
possible regexes to use for a query:
* `LIKE '99%6421'`: this should work well if we are sure we only have bibliographic alma ids (I'm pretty sure that's the case)
* `SIMILAR TO '[0-9]+6421'`: this will ensure numeric-only values, ending in 6421. this would allow alma ids that aren't bibliographic ids, but ensure we don't get component ids that happen to end in 6421. assuming we don't have all-numeric component ids.

The risk of false positives / negatives either way is probably negligibly small, especially for this use case.

hackartisan:
Ran the report on prod, it was pretty slow and eventually crashed with

`rake aborted!                                                                                                                                                                            
Faraday::ConnectionFailed: Failed to open TCP connection to n2t.net:443 (getaddrinfo: Temporary failure in name resolution)`

but it did get 520 objects in the report (attached) before it crashed. @tpendragon do we need to be sure we got every object? If so I will add some more error handling and a progress bar.

[ark_mismatch_report.csv](https://github.com/pulibrary/figgy/files/15180044/ark_mismatch_report.csv)

tpendragon:
That seems like enough to get a read on the situation.

PR #6356, opened 2024-04-18
Add report for resources with an alma source metadata id but a finding aids ark target

hackartisan:
advances #6343

### DIFF
--- a/app/queries/all_mms_resources_query.rb
+++ b/app/queries/all_mms_resources_query.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+class AllMmsResourcesQuery
+  def self.queries
+    [:all_mms_resources]
+  end
+
+  attr_reader :query_service
+  delegate :resource_factory, :run_query, to: :query_service
+  delegate :orm_class, to: :resource_factory
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def all_mms_resources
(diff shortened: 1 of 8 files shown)


hackartisan:
On staging it seems to check the resources like 2 per min, which is gonna take 32 days on prod. Not hitting errors anymore, though -- Maybe it'll run faster on prod? Worth a shot.

hackartisan:
@escowles It seems to take a really long time to check these ark resolutions. Trey said we can maybe get a dump of ezids, and that you maybe know how to get that?

hackartisan:
Okay thanks @escowles! I'll just have to look into why resolving them is so slow via the ezid client, and/or do it some other way.

escowles:
FWIW, the script I used the last time I did this just did a HEAD request to resolve them:

```
#!/bin/sh

# lookup redirect URL for all ARKs in file "arks.txt"

function f
{
  read ARK
  while [ "$ARK" ]; do
    LOC=`curl -s -I https://n2t.net/$ARK | grep Location`
    echo $ARK $LOC
    read ARK
  done
}

cat arks.txt | f
```

hackartisan:
Nice! Ha, my first plan was to do HEAD requests, before I got distracted by the Ezid client gem.

tpendragon (approved):

tpendragon:
This is great, good work!

PR #6358, opened 2024-04-19
Upgrade jquery-ui-rails

eliotjordan:
Updates jquery-ui-rails to address 4 dependabot alterts.

Meta question for later: Are we still using all of the[ jquery-ui components](https://github.com/pulibrary/figgy/blob/main/app/assets/javascripts/application.js#L27-L31)? For example, I think we're using the Lux datepicker now.


hackartisan:
How do / can we know that everything using jquery-ui stills works after this merge?

hackartisan:
It looks like it's mostly used in the (old) structure manager.

hackartisan:
I'm not sure it's working right? I just tried the structure manager on figgy-staging for this item: https://figgy-staging.princeton.edu/concern/scanned_resources/123bbbfb-4e7d-4bc6-9752-4b48925a1cdf/struct_manager

Screenshot 2024-04-23 at 4 45 56 PM

tpendragon (changes requested):
Old structure manager definitely breaks, and the nestedSortable plugin isn't updated for new jquery-ui. We'll have to delay this update until we remove the old structure editor.

PR #6371, opened 2024-04-25
Let errors pass through if file appender fails on create_node

hackartisan:
This will keep pdf derivative generation from only generating a portion
of the files, and keep resources from being created if not all files
were successfully attached.

It's a bigger mess to have things half-ingested and have to clean up
after.

refs #5964


tpendragon (dismissed):

tpendragon (approved):
I'm gonna approve, but I have a bunch of code in here so I'm not gonna merge.

PR #6379, opened 2024-04-30
Stop using <details> for file browser, use <li>

### LINKED ISSUE #6372: Remove VueDetails dependency
The [VueDetails](https://github.com/kim366/vue-details) dependency used in the [DirectoryPicker](https://github.com/pulibrary/figgy/blob/main/app/javascript/components/file_browser/directory_picker.vue#L8) does not work in Vue3 and has not been migrated. This is a blocker to the Vue2 -> Vue3 upgrade.

There are three ways to resolve this.

1. [Implement using refs](https://stackoverflow.com/questions/77240675/vue-3-toggleevent-on-details-tag)
2. Implement locally (or fork) and [modify the render](https://github.com/kim366/vue-details/blob/master/src/index.js#L9-L15) code [so that render works in Vue3](https://stackoverflow.com/questions/65529646/why-the-render-h-function-cannot-use-component-name-directly-to-render-in-vue3-x)
3. Implement show/hide functionality without using the `<details>` tag and its native element state, [as in the Tree component](https://github.com/pulibrary/figgy/blob/main/app/javascript/components/Tree.vue#L117-L120).

tpendragon:
Closes #6372 

By managing `expanded` ourselves and rendering the tree based on that we can get rid of this dependency.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1181,6 +1181,7 @@ GEM
     zeitwerk (2.6.13)
 
 PLATFORMS
+  arm64-darwin-21
   arm64-darwin-22
   x86_64-darwin-20
   x86_64-linux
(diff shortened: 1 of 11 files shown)


sdellis (approved):
Looks great! Thank you!

PR #6388, opened 2024-05-09
Track Downloads through Universal Viewer

### LINKED ISSUE #6376: Track Downloads through Plausible
These should be the downloads that go through the "Downloads" controller.

## Acceptance Criteria

- [ ] When I go to https://plausible.io/dpul.princeton.edu (or some filter of it) I see a downloads stat that includes all the downloads that went through Figgy's download controller that originated from DPUL-C.
- [ ] There is some way to show all the downloads that went through Figgy's downloads controller regardless of their origin. In this view you can filter by origin site.

## Notes

We want Figgy to send the data to DPUL's Plausible for this.

We might be able to get this via https://plausible.io/docs/file-downloads-tracking

We might want to track this through the buttons, or we might get this one for free through #6378.

https://plausible.io/docs/subdomain-hostname-filter#how-to-track-the-user-journey-across-domain-and-subdomains has some useful docs.

We either need a way to measure where the downloads are coming from via some tag manipulation, or have a custom metric we send to the API with some parameters and filter from those.

Answering the questions in these notes will take some research, which is part of this ticket.

## First Step

Look at the downloads tracking documentation and see if it'll help.

### LINKED ISSUE #6377: Track IIIF Image Downloads from Viewer
We might be able to get this with a snippet in the "Viewer" pane with download tracking enabled? If not, maybe a custom event? Fired from the IIIF middleware maybe?

## Acceptance Criteria

- [ ] There's a metric for downloads from the UV.

## Notes

We might get outbound link tracking from #6378, it might be a good idea to do this one after that one.

tpendragon:
The download pane in Universal Viewer:

1. Doesn't fire any javascript events (so we can't bind to an event and send a custom event)
2. Doesn't count as an outbound link (https://plausible.io/docs/outbound-link-click-tracking doesn't work)
3. Doesn't count as a "File Download" (https://plausible.io/docs/file-downloads-tracking doesn't work)
4. DOES use `window.open` to open a link. (https://github.com/UniversalViewer/universalviewer/blob/main/src/content-handlers/iiif/extensions/uv-openseadragon-extension/DownloadDialogue.tsx#L431)

Thus we can fire a custom event by, if Universal Viewer is loaded, overriding `window.open` in the viewer. This also adds an "?embed_host" property that can be added to the iframe, and sends the environment it's sending as. This is done via the documentation at https://plausible.io/docs/custom-props/for-pageviews#2-add-your-custom-properties-to-your-plausible-snippet.

Finally, this uses `manual` for Plausible, which means it doesn't automatically track page views - I don't think we care how many people go to `/viewer`, just how many people interact with that viewer.

The documentation for sending a custom event with properties in Plausible is here: https://plausible.io/docs/custom-props/for-custom-events#2-using-the-manual-method

Closes #6376 
Closes #6377 

This will miss any downloads done from within Figgy. We could do that, but I'm not sure it's necessary - that's not a public interface our patrons will be accessing.

### DIFF
--- a/app/javascript/entrypoints/viewer.js
+++ b/app/javascript/entrypoints/viewer.js
@@ -1,6 +1,7 @@
 import UVManager from '@viewer/uv_manager'
 import 'leaflet/dist/leaflet.css'
 const UVManagerInstance = new UVManager()
+window.uv_instance = UVManagerInstance
 let timer = window.setInterval(() => {
   if (window.UV !== undefined) {
     UVManagerInstance.initialize()
--- a/app/javascript/test/viewer/uv_manager.spec.js
+++ b/app/javascript/test/viewer/uv_manager.spec.js
@@ -108,6 +108,10 @@ describe('UVManager', () => {
     )
   }
 
+  function mockPlausible () {
+    window.plausible = vi.fn()
+  }
+
   beforeEach(() => {
     vi.clearAllMocks()
   })
@@ -118,6 +122,7 @@ describe('UVManager', () => {
       mockJquery()
       mockUvProvider()
       mockManifests(200)
+      mockPlausible()
       stubQuery({
         type: 'html',
         content: "<iframe src='https://figgy.princeton.edu/viewer#?manifest=https://figgy.princeton.edu/concern/scanned_resources/78e15d09-3a79-4057-b358-4fde3d884bbb/manifest'></iframe>",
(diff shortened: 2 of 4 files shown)


hackartisan:
Are there any plausible docs you used to work this out? It would be nice to have the links to them in the description of the PR.

tpendragon:
@hackartisan Good idea! Added some links.

tpendragon:
Drafted so I can add some tests of this plausible functionality.

eliotjordan (approved):

Issue #6391, opened 2024-05-10
[Figgy/production] FrozenError: can't modify frozen #<Class:#<RDF::URI:0x00007fbbc11fb3d0>>: #<RDF::Vocabulary::Term:0xf56a04 ID:http://pcdm.org/use#PreservedMetadata>


tpendragon:
## Backtrace

    line 176 of [PROJECT_ROOT]/app/services/preserver.rb: metadata_node
    line 126 of [PROJECT_ROOT]/app/services/preserver.rb: preserve_metadata
    line 28 of [PROJECT_ROOT]/app/services/preserver.rb: preserve!

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/107123516)

PR #6392, opened 2024-05-10
Use constants for PcdmUse

tpendragon:
Prevents the freeze issue and should be faster, too.

Closes https://github.com/pulibrary/figgy/issues/6391


eliotjordan (approved):
The code looks good. Ultimately, I think we should migrate to `http://pcdm.org/use#PreservationFile`.

tpendragon:
I think Anna did a while back.

hackartisan:
I like this, it's nice to have the list as a reference, too.

PR #6399, opened 2024-05-15
View unhealthy resources in health check modal

### LINKED ISSUE #6341: Allow users to diagnose and fix files in child resources from the parent resource.
# Wireframes

## Wireframe State 1: Collapsed list of problematic resources

01_figgy-issue-6297-wireframe-COLLAPSED

## Wireframe State 2: Display spinning Loader as list of problematic resources is generating

02_figgy-issue-6297-wireframe-LOADER

## Wireframe State 3: Expanded list of problematic resources

03_figgy-issue-6297-wireframe-EXPANDED

# Wireframe Updates

This report should:

1. Instead of "Show 10 problematic resources" it should say "Show Affected Resources"
1. Only show up for resources that have child resources (deep) (non-FileSets)
    * E.g if I have an Ephemera Project with a bad folder in a box, the Ephemera project's health report should have a link to the Folder, not the box, since the box's file manager will be empty.
2. Each link should send you to the file manager for that child resource
3. Each link should say how many are bad - e.g `XXI Bk, No. 117 (2416) - 7 Files Affected` 
4. The report should start to load when the health modal is opened.
5. Flip the arrow directions. When collapsed, the arrow should be pointing down, when expanded the arrow should point up.

# First Step

Write a query to return all child resources that are returning bad health statuses.

eliotjordan:
Closes #6341 

Screenshot 2024-05-16 at 11 24 59 AM
Screenshot 2024-05-16 at 11 25 10 AM


app/queries/find_deep_errored_file_sets.rb
```diff
@@ -12,7 +12,7 @@ def initialize(query_service:)
   end
 
   def find_deep_errored_file_sets(resource:)
-    query_service.connection[relationship_query, id: resource.id.to_s].first[:count]
```
tpendragon, line 15: Suggestion: Maybe we should have a `count: true` option? I think this changes the query to always realize every errored file set as a real resource, which is more ruby-expensive than just returning the count. I realize you also need the resources for this case, though, so maybe a param to switch between those states.

tpendragon (changes requested):
Just one thing I think might be a good idea, open to be convinced otherwise though.

app/queries/find_deep_errored_file_sets.rb
eliotjordan: That's a good point. Pushed up a commit to address this.

tpendragon (approved):

PR #6431, opened 2024-06-12
Revert "Use Sound, not Audio"

hackartisan:
This reverts commit 31cd47194b562cca636b435d5295671d7d09b7da.
player loads and everything but sound doesn't play


hackartisan:
Actually it works; see https://figgy-staging.princeton.edu/catalog/f3b0e79e-6b3e-4955-a321-d377660b37e1 -- the derivative files in general seem to be 406ing, even on main, but regenerating derivatives fixes it.

PR #6438, opened 2024-06-24
Adds UV click event listener and sends custom event to Plausible

sdellis:
This may be a naive approach, but it listens for a click event on the #uv element, logs it (for testing), and dispatches a `uv-click` event to Plausible. I am unsure of how to test this in Plausible to make sure it was received.

### DIFF
--- a/app/javascript/viewer/uv_manager.js
+++ b/app/javascript/viewer/uv_manager.js
@@ -125,6 +125,7 @@ export default class UVManager {
     }, this.urlDataProvider)
     this.cdlTimer = new CDLTimer(this.figgyId)
     this.cdlTimer.initializeTimer()
+    this.bindUvClick()
   }
 
   // Universal Viewer uses window.open for download links, and we want to track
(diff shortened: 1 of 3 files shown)


app/javascript/viewer/uv_manager.js
```diff
@@ -137,6 +138,26 @@ export default class UVManager {
...
+    const uvClick = new Event("uv-click", { bubbles: true });
+
+    uvElement.addEventListener(
+      "click",
+      (event) => {
+        /* … */
+        uvElement.dispatchEvent(uvClick)
+        const cachedOpen = window.open
+        window.open = function (url, target, features) {
+          window.plausible('uv-click', { props: { url } })
```
tpendragon, line 152: I don't think you need to override `window.open`, just put this below 149 and delete all this open stuff.

For the download controller that was more about attaching to the way UV does downloads.

app/javascript/viewer/uv_manager.js
```diff
@@ -137,6 +138,26 @@ export default class UVManager {
     }
   }
 
+  bindUvClick () {
+    const uvElement = document.getElementById('uv')
+    const uvClick = new Event("uv-click", { bubbles: true });
+
+    uvElement.addEventListener(
```
tpendragon, line 145: What are the chances we can get a test? I think I wrote one for plausible stuff in the viewer test before that you could use as a model.

tpendragon (changes requested):

app/javascript/viewer/uv_manager.js
sdellis: Done!

tpendragon (approved):

sdellis:
Closes #6438

PR #6445, opened 2024-06-26
Adds custom property to track record page views in figgy

sdellis:
This works! However, this should probably be implemented in DPUL record pages, not Figgy's. @tpendragon can you confirm that's what we want?


tpendragon:
@tpendragon It should be in DPUL, yeah.

tpendragon:
@sdellis It should be in DPUL, yeah.

Issue #6436, opened 2024-06-20
Spike on making PDF derivative generation more resilient to disk failure or slowness


tpendragon:
We've noticed disk problems on our system mounts, and we're seeing a lot of problems with generating PDF derivatives, as in #6366.

This is a test to see if we can improve things without completely rehauling how we do PDF derivatives.

## Acceptance Criteria

- [ ] If something breaks in saving a file while generating PDF derivatives it doesn't leave a bunch of empty files attached to the resource. It instead tries again.

## First Step

Find a way to reproduce this scenario? Or fake it in a test.

## Implementation Tips

Maybe for each page see if it's zero bytes, and if so throw our own error so the job will retry.

## Sudden Priority Justification

PDFs are breakin' all over the place.

tpendragon:
#6446 seemed to help a lot. One thing is it looks like sometimes VIPS doesn't make the file where we told it to - or maybe at all. See https://app.honeybadger.io/projects/53391/faults/108852782. Retrying on ENOENT might be a good idea.

tpendragon:
I regenerated ~ 12 broken PDFs on https://figgy.princeton.edu/catalog/3931aa55-afd6-4c29-88bc-afaac7251f38 and it all came through green.

PR #6446, opened 2024-06-26
PDF Resilience Attempts

tpendragon:
For PDF derivatives this does three things:

1. Ensure the whole resource rolls back appropriately if it fails to generate a derivative.
2. Retry if VIPS creates a zero byte file.
3. Uses the default DPI settings. 300 was resulting in 300 MB files PER PAGE for a PDF that was only 20 MB total.

Now it should only be able to make a zero byte file if the storage uploader somehow creates one when copying the file over.

Work towards #6436


sdellis (approved):
This looks good to me, @tpendragon ! 👍

PR #6447, opened 2024-06-27
I6355 structure manager labels

### LINKED ISSUE #6355: Editing section labels in the Structure Manager should not require enter/button-press
It's confusing to users that they need to hit enter or click the check icon to get the labels to "stick" and save. We should remove this unnecessary step. 

## Acceptance Criteria

- [ ] User can enter a label in the structure editor, not hit the checkmark, hit save, and it saves.
- [ ] There's no checkmark

## First Step

Update the feature test so it doesn't hit the check mark.

AminEdZare:
This fixes #6355. One suggestion is to autofocus label edits when folder is created. This is beyond the scope of the ticket but could also be added here.

### DIFF
--- a/app/javascript/components/Tree.vue
+++ b/app/javascript/components/Tree.vue
@@ -53,18 +53,10 @@
                 v-model="structureData.label"
                 type="text"
                 class="folder-label-input"
-                @keyup.enter="saveLabel(id)"
+                @keyup="saveLabel(id)"
+                @keydown.enter="hideLabelInput()"
+                @blur="hideLabelInput()"
               >
-            </div>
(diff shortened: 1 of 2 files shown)


tpendragon (changes requested):
This is nice - it works when I hit enter, but if I click "Edit", change it, and click "Save Structure" it says it saves, but if I refresh it's not saved.

tpendragon (approved):
I was using the wrong branch.

tpendragon (changes requested):
Can we make sure the test fails if the Tree.vue wasn't changed?

tpendragon (approved):

PR #6453, opened 2024-07-05
DPUL Success Dashboard - Phase 1 (data only, no graphs)

sdellis:


app/services/dpul_success_dashboard_report_generator.rb
```diff
@@ -0,0 +1,168 @@
+# frozen_string_literal: true
+require "json"
+
+# frozen_string_literal: true
```
tpendragon, line 4: We should get rid of this extra line.

app/controllers/reports_controller.rb
```diff
@@ -100,6 +116,11 @@ def validate_collection_count_params
       { collection_ids: collection_ids, date_range: valid_dates }
     end
 
+    def validate_dpul_date_range
+      return nil if params[:date_range].blank? || !valid_dates
+      { date_range: valid_dates }
```
tpendragon, line 121: When I run this in staging I get this error: https://app.honeybadger.io/projects/53391/faults/109834978.

When I run the code in the staging console, Plausible gives me this back:

```json
"{\"error\":\"Invalid format for `date` parameter. When using a custom period, please include two ISO-8601 formatted dates joined by a comma. See https://plausible.io/docs/stats-api#time-periods\"}"
```

I think the problem is that Plausible expects dates with no time component, we need to make sure we initialize this report with Dates, not DateTimes.

app/services/dpul_success_dashboard_report_generator.rb
```diff
@@ -0,0 +1,168 @@
...
+      date_str = date_key.to_s
+      stats_hash[date_str]&.merge!(metrics.transform_keys(&:to_s))
+    end
+
+    # Convert stats hash back to array
+    stats_with_metrics = stats_hash.values
+    stats_with_metrics << @totals_hash.transform_keys(&:to_s)
+  end
+
+  def traffic_request
```
tpendragon, line 48: Non-blocking, but I wonder if we can dedup some of this boilerplate with something like

```ruby
def plausible_request
    Faraday.new(url: "https://plausible.io") do |conn|
      conn.request :authorization, "Bearer", Figgy.config["plausible_api_key"]
      conn.adapter Faraday.default_adapter
      conn.headers["Content-Type"] = "application/json"
      conn.params["site_id"] = "dpul.princeton.edu"
      yield conn
    end
end
```

tpendragon (changes requested):
Testing it will probably be tough - I think you'll have to change all your stubs to not accept time components. For example 
```
https://plausible.io/api/v1/stats/breakdown?date=2024-07-01T00:00:00%2B00:00,2024-07-03T00:00:00%2B00:00&interval=date&metrics=visitors,bounce_rate&period=custom&property=visit:source&site_id=dpul.princeton.edu
```

should be

```
https://plausible.io/api/v1/stats/breakdown?date=2024-07-01,2024-07-03&interval=date&metrics=visitors,bounce_rate&period=custom&property=visit:source&site_id=dpul.princeton.edu
```

as the first one raises an error in actual Plausible.

app/services/dpul_success_dashboard_report_generator.rb
sdellis: Done!

tpendragon (approved):

PR #6459, opened 2024-07-10
ScannedMaps should not be start_canvases

### LINKED ISSUE #6456: Scanned Map is not rendering a IIIF Manifest
## Success Criteria

- [ ] https://figgy.princeton.edu/catalog/1419afdd-0766-44f1-8bba-92c251a6885a renders a viewer.

## First Step

Go through the backtrace and find out why it's calling `ocr_content`.

## Backtrace

    line 414 of [PROJECT_ROOT]/app/services/manifest_builder.rb: ocr_content
    line 14 of [PROJECT_ROOT]/app/services/manifest_builder/canvas_rendering_builder.rb: apply
    line 8 of [PROJECT_ROOT]/app/services/manifest_builder/canvas_builder.rb: apply_record_properties

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/109045278)

The resource is broken for viewing in a way that our staff can't fix.

tpendragon:
If a scanned map gets set as a start canvas, don't try to render it as one.

Closes #6456

### DIFF
--- a/spec/services/manifest_builder/by_resource_type/manifest_builder_scanned_maps_spec.rb
+++ b/spec/services/manifest_builder/by_resource_type/manifest_builder_scanned_maps_spec.rb
@@ -38,7 +38,7 @@ RSpec.describe ManifestBuilder do
   context "when given a nested scanned map set" do
     subject(:manifest_builder) { described_class.new(query_service.find_by(id: scanned_map.id)) }
     let(:scanned_map) do
-      FactoryBot.create_for_repository(:scanned_map, description: "Test Description", member_ids: child.id)
+      FactoryBot.create_for_repository(:scanned_map, description: "Test Description", member_ids: child.id, start_canvas: child.id)
     end
     let(:child) { FactoryBot.create_for_repository(:scanned_map, files: [file]) }
     it "builds a IIIF document" do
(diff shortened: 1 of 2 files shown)


spec/services/manifest_builder/by_resource_type/manifest_builder_scanned_maps_spec.rb
```diff
@@ -38,7 +38,7 @@
   context "when given a nested scanned map set" do
     subject(:manifest_builder) { described_class.new(query_service.find_by(id: scanned_map.id)) }
     let(:scanned_map) do
-      FactoryBot.create_for_repository(:scanned_map, description: "Test Description", member_ids: child.id)
+      FactoryBot.create_for_repository(:scanned_map, description: "Test Description", member_ids: child.id, start_canvas: child.id)
```
hackartisan, line 41: It's not clear now why the test setup has the start canvas. Maybe update the context to say that it's mistakenly set as a start canvas? I don't know, it might be better to have a separate test so that it's clear this is one is sort of a happy path test, have a new one for this edge case.

hackartisan (changes requested):
The title is ScannedMaps should not be start_canvas but the code reads like, only a file set should be a start_canvas. seems fine, just attempting to add clarity. please see my comment below.

spec/services/manifest_builder/by_resource_type/manifest_builder_scanned_maps_spec.rb
tpendragon: Good read, made a new one.

hackartisan (approved):

PR #6490, opened 2024-09-10
This is how I got figgy running.

hackartisan:
You can't use rosetta directly anymore. It appears to be built-in
to the mac OS, which transparently detects when it needs to use rosetta.


eliotjordan (approved):

bin/setup_credentials
eliotjordan, line 1: What happened / is happening with this file?

hackartisan: mode changed to executable

PR #6511, opened 2024-09-24
Update readme for clarity and to reduce duplicate documentation

hackartisan:
Several things were repeated, there was old content, and there was too
much detail in some places. Duplicate setup_keys script was removed. It
was probably moved to be called setup_credentials at some point to keep
it from being confusing given the setup_keys rake task.


tpendragon (approved):
This is great - thank you!

PR #6531, opened 2024-10-29
Updating link for Special Collections citation guidelines

escowles:
We noticed this change because a redirect was dropped in the website migration. So we should update to the current URL for the page and not use the old URL which depends on the redirect.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -112,7 +112,7 @@ en:
           label: "IIIF Manifest URL"
         rights_statement:
           boilerplate: |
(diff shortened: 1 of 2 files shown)


tpendragon (approved):

escowles:
@tpendragon I don't think that failure was related to the change I made here — but I did see a couple of things that I thought would improve the test (e.g., selecting a value based on its name instead of just assuming which one would be first in the data file). In a now-removed commit, I dumped the whole text and noticed that the value being tested was present but truncated, so I'm not sure what would cause that.

tpendragon (approved):

PR #6534, opened 2024-11-04
Root health.json errors rarely, filtered checks error if they fail.

### LINKED ISSUE #6533: `health.json` should be able to be used for nginx checks without arguments, but datadog health checks should continue working
## Reasoning

This is a research ticket and implementation ticket. We should be able to put `https://figgy.princeton.edu/health.json` as a check for the load balancer (so it will bring down unhealthy nodes automatically), but still have our non-critical services able to be monitored by datadog. Preferably without having to check the contents of the response.

## Background Info

Non-critical services in the health monitor: https://github.com/lbeder/health-monitor-rails?tab=readme-ov-file#provider-configuration

Datadog checks: https://github.com/pulibrary/princeton_ansible/blob/main/group_vars/figgy/production_webserver.yml#L3-L68

## Success Criteria

- [ ] https://figgy.princeton.edu/health.json returns a 200 even when sidekiq has too many retries.
- [ ] The sidekiq job monitor and datadog health dashboard still recognizes when sidekiq has too many retries (https://app.datadoghq.com/dashboard/96b-77b-bri/dls-health-dashboard?fromUser=false&refresh_mode=sliding&from_ts=1729870060467&to_ts=1730474860467&live=true)

## First Step

Try out setting critical: false and see what the response is for `health.json` and specifically checking that service.

tpendragon:
This might be interesting for upstream, but I can't really imagine how they'd want it. I'll make an issue for discussion in their repo.

Closes #6533.

### DIFF
--- a/spec/requests/health_monitor_spec.rb
+++ b/spec/requests/health_monitor_spec.rb
@@ -11,9 +11,6 @@ RSpec.describe "Health Monitor", type: :request do
   describe "GET /health" do
     it "has a success response" do
       stub_aspace_login
-      allow(Net::SMTP).to receive(:new).and_return(instance_double(Net::SMTP, "open_timeout=": nil, start: true))
-      # stub the number of processes since sidekiq doesn't run in test
-      allow(Sidekiq::Stats).to receive(:new).and_return(instance_double(Sidekiq::Stats, processes_size: 1))
 
       get "/health.json"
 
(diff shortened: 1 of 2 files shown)


tpendragon:
Created https://github.com/lbeder/health-monitor-rails/issues/126 for discussion upstream.

spec/requests/health_monitor_spec.rb
```diff
@@ -11,9 +11,6 @@
   describe "GET /health" do
     it "has a success response" do
```
hackartisan, line 12: Can you update the test to make it clear we expect success even when a non-critical service is failing?

hackartisan (changes requested):

hackartisan (approved):

PR #6546, opened 2024-11-15
SMTP check should happen every 5 minutes.

### LINKED ISSUE #6545: SMTP health monitor should cache a success for 5 minutes
## Reasoning

We don't want to bombard the mail server for every health check, and the mail server check is more about knowing the first time if we've configured the box correctly.

## Acceptance Criteria

- [ ] If the mail server check succeeds it shouldn't ping the mail server again for 5 minutes.

## First step

Write a test. Find a way to enable the rails cache.

tpendragon:
Closes #6545

### DIFF
--- a/app/monitors/smtp_status.rb
+++ b/app/monitors/smtp_status.rb
@@ -1,10 +1,19 @@
 # frozen_string_literal: true
 class SmtpStatus < HealthMonitor::Providers::Base
+  class << self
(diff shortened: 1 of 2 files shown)


maxkadel (dismissed):
Looks good! Thanks @tpendragon!

maxkadel (approved):

PR #6551, opened 2024-11-20
Add PDF type to bulk edit.

### LINKED ISSUE #6550: Add PDF type to bulk edit
## Reasoning

EAL folks often adjust the PDF type for whole collections, especially for DPUL exhibits. If it's in bulk edit we don't have to be the ones to do it.

## Acceptance Criteria

- [ ] PDF Type can be selected from bulk edit.

## First Step

Write a test.

EAL folks need PDF type adjusted for a collection. It's about the same amount of effort to add this as a feature as it is to write a script to do that.

tpendragon:
Closes #6550


hackartisan (changes requested):
Could we have a test that if it's not selected it doesn't change an object?

tpendragon:
> Could we have a test that if it's not selected it doesn't change an object?

Good idea, done.

hackartisan (approved):
Thank you!

Issue #6605, opened 2025-01-17
IIIF Presentation V3 manifests are returning errors on validation


eliotjordan:
Using the [presentation api validator](https://presentation-validator.iiif.io/) with a [staging manifest](https://figgy-staging.princeton.edu/concern/scanned_resources/2325ed52-bb0b-42a9-9fa5-42c7beb106a4/manifest) returns 8 different errors:

- [ ] Error 1 of 8. Message: Additional properties are not allowed ('logo', 'posterCanvas' were unexpected)

```
Detail:
Path: /[Additional properties are not allowed ('logo', 'posterCanvas' were unexpected)]
Description:
{"@context":"[ ... ]","type":"Manifest","id":"https://figgy-staging.princeton.edu/concern/scanned_resources/2325ed52-bb0b-42a9-9fa5-42c7beb106a4/manifest","label":"{ ... }","metadata":"[ ... ]","rendering":"[ ... ]","items":"[ ... ]","behavior":"[ ... ]","posterCanvas":"{ ... }","structures":"[ ... ]","seeAlso":"[ ... ]","rights":"http://rightsstatements.org/vocab/CNE/1.0/","logo":"[ ... ]","thumbnail":"[ ... ]"}
```

- [ ] Error 2 of 8. Message: 'Download as PDF' is not of type 'object'
```
Detail: Language string, must have a language and value must be an array.
Path: /rendering[0]/label/['Download as PDF' is not of type 'object']
Description:
"Download as PDF"
```

- [ ] Error 3 of 8. Message: 'type' is a required property
```
Detail: Classes MUST have an id and type property and MAY have a label.
Path: /rendering[0]/['type' is a required property]
Description:
{"label":"Download as PDF","format":"application/pdf","id":"https://figgy-staging.princeton.edu/catalog/2325ed52-bb0b-42a9-9fa5-42c7beb106a4/pdf"}
```

- [ ] Error 4 of 8. Message: 'id' is a required property
```
Detail: Classes MUST have an id and type property and MAY have a label.
Path: /items[0]/items[0]/items[0]/['id' is a required property]
Description:
{"type":"Annotation","motivation":"painting","body":"{ ... }","target":"https://figgy-staging.princeton.edu/concern/scanned_resources/2325ed52-bb0b-42a9-9fa5-42c7beb106a4/manifest/canvas/42791640-9681-41f0-961a-b391d3781f0f"}
```

- [ ] Error 5 of 8. Message: 'items' is a required property
```
[...]

PR #6625, opened 2025-02-05
Fix IIIF V3 manifests so they validate

eliotjordan:
Closes [#6605](https://github.com/pulibrary/figgy/issues/6605)

- Audio, Video, Playlist, ScannedMap, and MapSet resources all generate 100% valid V3 manifests.
- ScannedMaps and MapSets now return V3 by default
- Added tests into V3 specs to test generated manifests against the V3 JSON schema.
- Disable downloads on a per-canvas level and remove outdated UV-specific disableUI service. Newer versions of UV and other viewers like clover don't use this property.


app/services/manifest_builder_v3/canvas_rendering_builder.rb
```diff
@@ -45,7 +45,11 @@ def resource
       # Determines if the resource be downloaded by the user
       # @return [Boolean]
       def downloadable?
-        @record.respond_to?(:resource) || @record.is_a?(FileSet)
+        if @record.respond_to?(:resource) || @record.is_a?(FileSet)
+          @record.resource.decorate.downloadable?
```
tpendragon, line 49: Can you explain the change in logic here? I think this would be quite expensive, re-fetching the parent for every FileSet.

spec/services/manifest_builder/by_resource_type/manifest_builder_ephemera_spec.rb
```diff
@@ -44,9 +44,9 @@
   context "when an ephemera folder has a transliterated title" do
     subject(:manifest_builder) { described_class.new(query_service.find_by(id: folder.id)) }
     let(:folder) { FactoryBot.create_for_repository(:ephemera_folder, title: ["title"], transliterated_title: ["transliterated"]) }
-    it "includes that in the manifest label" do
+    it "the root label displays the original title only because it is single valued" do
       output = manifest_builder.build
-      expect(output["label"]).to eq ["title", "transliterated"]
+      expect(output["label"]).to eq "title"
```
tpendragon, line 49: What's the impact on viewer display for this change? Anything?

spec/services/manifest_builder/by_resource_type/manifest_builder_scanned_resource_spec.rb
```diff
@@ -158,9 +158,9 @@ def logical_structure(file_set_id)
     let(:scanned_resource) do
       FactoryBot.create_for_repository(:scanned_resource, title: ["title1", "title2"])
     end
-    it "uses an array" do
+    it "displays a root label with a single string as label is a single value in Prezi V2" do
       output = manifest_builder.build
-      expect(output["label"]).to eq ["title1", "title2"]
+      expect(output["label"]).to eq "title1"
```
tpendragon, line 163: Same question as above - what's the impact of this change? Did UV used to concat?

spec/services/manifest_builder_v3/by_resource_type/manifest_builder_v3_audio_resource_spec.rb
```diff
@@ -49,14 +51,36 @@
...
+
+    context "with an accompanying image file" do
+      let(:image_file) { fixture_file_upload("files/example.tif") }
+      let(:change_set) { ScannedResourceChangeSet.new(scanned_resource, files: [file, image_file], downloadable: "none") }
+
+      it "builds a manifest with an accompanyingCanvas element", run_real_characterization: true do
+        change_set_persister.save(change_set: change_set)
+        output = manifest_builder.build
+
+        expect(output["items"][0]["accompanyingCanvas"]["width"]).to eq 200
```
tpendragon, line 79: Does this accompanyingCanvas render in UV?

spec/services/manifest_builder_v3/by_resource_type/manifest_builder_v3_audio_resource_spec.rb
```diff
@@ -208,7 +209,7 @@
       user = User.first
       stub_findingaid(pulfa_id: "C0652")
       stub_findingaid(pulfa_id: "C0652_c0377")
-      IngestArchivalMediaBagJob.perform_now(collection_component: "C0652", bag_path: bag_path, user: user, member_of_collection_ids: [collection.id.to_s])
+      IngestArchivalMediaBagJob.perform_now(collection_component: "C0652", bag_path: bag_path, user: user, member_of_collection_ids: [collection.id.to_s], visibility: "open")
```
tpendragon, line 212: Why was the visibility change necessary?

spec/services/manifest_builder_v3/by_resource_type/manifest_builder_v3_playlist_spec.rb
```diff
@@ -79,9 +84,6 @@
       it "generates the Canvases for the FileSets" do
         expect(output).not_to be_empty
 
-        expect(output).to include("rendering")
-        expect(output["rendering"]).to be_empty
```
tpendragon, line 83: How do we provide access to a canvas' full tiff download? Is the rendering somewhere else? Or do we still have canvas renderings and this just makes sure we don't have blank values?

tpendragon (changes requested):
A few questions. Have you tested the display in Figgy with these changes? If so, any notes to provide there, or should I kick this around on staging?

spec/services/manifest_builder/by_resource_type/manifest_builder_ephemera_spec.rb
eliotjordan: No impact. The viewer only displays one title.

spec/services/manifest_builder/by_resource_type/manifest_builder_scanned_resource_spec.rb
eliotjordan: UV does not concat as far as I can tell.

app/services/manifest_builder_v3/record_property_builder.rb
```diff
@@ -7,17 +7,7 @@ class RecordPropertyBuilder < IIIFManifest::V3::ManifestBuilder::RecordPropertyB
     def apply(manifest)
       manifest = super
       manifest.viewing_direction = viewing_direction if viewing_direction.present? && manifest.respond_to?(:viewing_direction=)
-      manifest.behavior = ["auto-advance"]
-      if record.resource.downloadable == ["none"]
-        manifest["service"] ||= []
-        manifest["service"] << {
-          "@context" => "http://universalviewer.io/context.json",
-          "profile" => "http://universalviewer.io/ui-extensions-profile",
-          "disableUI" => ["mediaDownload"]
```
tpendragon, line 16: So if we remove this, does the download pane still go away for things where download is disabled?

app/services/manifest_builder_v3/canvas_rendering_builder.rb
eliotjordan: We can't (once we upgrade UV or in the places where we use clover) use the custom UV disableUI service. Most of the download rendering are on the individual canvases so we need to check if we should build them or not. I'll push up a commit that's more efficient.

app/services/manifest_builder_v3/record_property_builder.rb
eliotjordan: It does not go away, but says there are no downloads. If you want to make this happen going forward (once we upgrade UV), I think we'll have to somehow pass the [uv_config_downloads_disabled](https://github.com/pulibrary/figgy/blob/main/public/uv/uv_config_downloads_disabled.json) config that we have. Though I don't see anyplace where we are using it currently. I

spec/services/manifest_builder_v3/by_resource_type/manifest_builder_v3_audio_resource_spec.rb
eliotjordan: Not necessary. Fixed.

spec/services/manifest_builder_v3/by_resource_type/manifest_builder_v3_audio_resource_spec.rb
eliotjordan: Hmm... It doesn't appear to in this version of UV. I guess we could add the posterCanvas back until we switch over to Clover to audio and video.

app/services/manifest_builder_v3/record_property_builder.rb
tpendragon: That's surprising, I wonder if it's the config disabling the downloads and we're fine here? I would have expected it to still show that like, "download a jpg" selection that uses the IIIF image API

app/services/manifest_builder_v3/canvas_rendering_builder.rb
```diff
@@ -42,11 +42,11 @@ def resource
         @record.respond_to?(:resource) ? @record.resource.to_model : @record
       end
 
-      # Determines if the resource be downloaded by the user
+      # Determines if the resource can be downloaded by the user
       # @return [Boolean]
       def downloadable?
-        if @record.respond_to?(:resource) || @record.is_a?(FileSet)
-          @record.resource.decorate.downloadable?
+        if @record.respond_to?(:resource)
```
tpendragon, line 48: I'm not sure this logic is equivalent - before FileSets always returned true.

I honestly do not know what this downloadable check is.

spec/services/manifest_builder_v3/by_resource_type/manifest_builder_v3_audio_resource_spec.rb
eliotjordan: Added back posterCanvas.

app/services/manifest_builder_v3/canvas_rendering_builder.rb
eliotjordan: I think the logic was flawed before. It checks if the parent resource has downloadable set to "public" and whether it is in a public state. Only then does it render the download links.

eliotjordan:
I refactored the download logic and added a [test that the parent is only retrieved once](https://github.com/pulibrary/figgy/pull/6625/files#diff-71a71f165382cc6651ee6287e55f5d67612a7c5dbb4317bc237b63dc6233c13aR94) to call the `downloadable?` method. I discovered some [sneaky recursion](https://github.com/pulibrary/figgy/blob/6210853e9b5d9391de1b26b72bfa7e326dac054c/app/services/manifest_builder_v3/root_node.rb#L113) that had to be worked around. The parent resource was being retrieved at least once for every FileSet. This should speed up V3 manifest generation all around.

tpendragon (dismissed):
Nice! I'm so happy that test uncovered some more optimizations. Good work!

tpendragon:
@eliotjordan Looks like we need some coverage.

tpendragon (approved):

PR #6651, opened 2025-03-17
Add support for MP3s

### LINKED ISSUE #6645: Add MP3 File Format at an acceptable/readable format to Figgy.
### Summary (Product Owner)
Requested twice this week by users in SC and EAL for their sound recordings (one from George Blood). Folks are wanting to use an MP3 as the access file in figgy.  @tventimi most recently for 2 spoken-word mp3 collections. It would be helpful to have the option to add this format to figgy as well. 

### User Story (Product Owner, when applicable)
_As a [persona], I want to [describe intent], so that [explain the big problem you are trying to solve and its benefit]_

As a figgy ingester working with sound files from vendors, I would like to have the option to ingest an MP3 file as the primary access file to figgy so that i do not have to convert files already in hand from MP3 to WAV.

### Impact (Product Owner)
There is no hard deadline, but in speaking with DLS this sounds like a relatively light life, code-wise.
### Priority recommendation (Product Owner)

- [ ] asap
- [x] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification (Product Owner)
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "dls-work-cycle" labels_
users would be able to ingest materials that they already have in hand, versus converting to WAV if users need immediate access.

### Implementation Notes

We'll have to create a small MP3 or copy them from one of the AV bags to use for testing.

We think this can get fixed by changing https://github.com/pulibrary/figgy/blob/0a7bdf808046bb944a8a6d18a96fff9ca5516471/app/characterization_services/mediainfo_characterization_service.rb#L77 to handle MP3s.

### Acceptance Criteria

- [ ] When I ingest an MP3, it plays in the player.

### First Step

Write a test.

tpendragon:
Closes #6645


sdellis (approved):
LGTM 👍

PR #6663, opened 2025-03-26
Add support for more file types for Ephemera bulk ingest.

### LINKED ISSUE #6662: Bulk Ingest functionality for Figgy Ephemera needs to include MP4, JPG, PDF.
### Summary (Product Owner)
Reported by @roelmoon . Bulk ingest for Figgy Ephemera does not work for materials that are not TIFF. Attempted: MP4 and JPG. both did not work.

### User Story (Product Owner, when applicable)
_As a [persona], I want to [describe intent], so that [explain the big problem you are trying to solve and its benefit]_
As a digital studio staff member, I need to be able to bulk ingest file formats that include MP4, JPG, TIFF, in order to fulfill bulk ingest request of ephemera materials from Isilon.

### Impact (Product Owner)
Work is stopped for our collaboration until we are able to get the videos in to the ephemera project. The directories number more than 100, so it would be too tedious to do one at a time (which works fine).
### Priority recommendation (Product Owner)

- [x] asap
- [ ] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification (Product Owner)
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "dls-work-cycle" labels_
We are in the final throws of the MOU and are moving forward with negotiations with captioning. We will need to provide folks access to the videos so that they may be captioned before being completed and made available in DPUL. 
### Implementation Notes
_To be filled out by DLS._

### Acceptance Criteria
_To be filled out by DLS.Some requirements to consider, including:_

- [ ] The UI implemented for this issue meets accessibility requirements
- [ ] New functionality is documented

### First Step
_To be filled out by DLS._

tpendragon:
Closes #6662


sdellis (commented):
@tpendragon I'm seeing a mp3 fixture file but don't see mp3 listed in the acceptable formats. Should mp3s be allowed or not?

tpendragon:
> @tpendragon I'm seeing a mp3 fixture file but don't see mp3 listed in the acceptable formats. Should mp3s be allowed or not?

Good catch, thank you! Added.

kelea99:
Thank you for catching the MP3, everyone!

tpendragon:
I could add them, but aren't you planning to add captions post ingest? Or will you just delete and ingest again?

kelea99:
In this use case I would bulk add captions post-ingest. If we get both at the same time (which may happen), i guess my question is: do we:

1.  build out now for bulk ingest of 2 different file types to the same record
2. build out bulk ingest to ingest one file type at a time, but add support for WebVTT bulk ingest
3. build out bulk ingest only for the image files and individually upload the WebVTT, since we need to set options for original 1:1 versus english translation vs... 
the webVTT seems like it would be more persnickety, since you would have to note if it was the original, what languages it was in. and then if there were other webVTT files in that same directory for different translations. knowing nothing, this seems like a bigger ticket to think through. but you know better than I, @tpendragon

sdellis (approved):
LGTM! 👍

PR #6672, opened 2025-04-03
Try Devbox for Figgy Setup

tpendragon:
This is the first time I've gotten Figgy's full suite to run without a bunch of fiddling in my environment in a long time.

Could someone try this out and see if it also works for them?

To set up you do this:

```
curl -fsSL https://get.jetify.com/devbox | bash
devbox run setup
devbox shell
./bin/parallel_rspec_coverage
```

There's no lando or anything, this installs `podman` and runs the containers as services when you setup.


tpendragon:
We'll look at this kinda stuff when we have more energy for looking at dev environments.

PR #6675, opened 2025-04-08
Try Devenv for Figgy.

tpendragon:
Alternative to #6672


tpendragon:
We'll look at this kinda stuff when we have more energy for looking at dev environments.

PR #6686, opened 2025-04-23
Don't propagate visibility for folders.

### LINKED ISSUE #6685: Marking a box "all_in_production" is synchronizing visibility to all folders
### Summary or User Story

This might be erasing visibility data that has been carefully set.

### Acceptance Criteria

- [ ] There's a test that verifies this doesn't happen, or it's fixed.

### First step

Write a test.

### Sudden Priority Justification

If the workflow is opening up material that shouldn't be open, we need to get it fixed immediately.

tpendragon:
Closes #6685


hackartisan (changes requested):
This looks good and I'm ready to approve, but I wonder if we want to consider documenting the way propagation does / should work for various kinds of resources. So I'm going to block so we can consider that question.

hackartisan (approved):

PR #6695, opened 2025-05-08
Validate controller resource class

### LINKED ISSUE #6678: Members method is being called on EphemeraTerms and raising exception
## Backtrace

    line 171 of [PROJECT_ROOT]/app/services/manifest_builder.rb: members
    line 180 of [PROJECT_ROOT]/app/services/manifest_builder.rb: leaf_nodes
    line 600 of [PROJECT_ROOT]/app/services/manifest_builder.rb: av?

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53391/faults/119135864)

```
[Figgy/production] NoMethodError: undefined method `members' for #<EphemeraTermWayfinder:0x00007f39b4e2a140 @resource=#<EphemeraTerm id=#<Valkyrie::ID:0x00007f39cdb91870 @id="3cb7627b-defc-401b-9959-42ebc4488f74"> internal_resource="EphemeraTerm" created_at="2018-02-22T19:13:49.949Z" updated_at="2018-03-09T20:19:33.414Z" new_record=false optimistic_lock_token=[] read_groups=[] read_users=[] edit_users=[] edit_groups=[] label=["Japan"] uri=[] code=["ja"] tgm_label=[] lcsh_label=[] member_of_vocabulary_id=...

```

There is a mismatch between the EphemeraTerm resource class and the controller resource class. I'm not sure what or who is generating this url or why, but the ephemera term id is being used with an ephemera_folders controller. EphemeraTerms do not have a `manifest` route.

https://figgy.princeton.edu/concern/ephemera_folders/3cb7627b-defc-401b-9959-42ebc4488f74/manifest
where "3cb7627b-defc-401b-9959-42ebc4488f74" is an EphemeraTerm.

eliotjordan:
Closes #6678

### DIFF
--- a/spec/controllers/ephemera_folders_controller_spec.rb
+++ b/spec/controllers/ephemera_folders_controller_spec.rb
@@ -169,7 +169,7 @@ RSpec.describe EphemeraFoldersController, type: :controller do
 
         get :manifest, params: { id: resource.id.to_s, format: :json }
 
-        expect(response).to be_forbidden
+        expect(response).to be_unauthorized
       end
       it "displays complete items" do
         resource = FactoryBot.create_for_repository(:complete_ephemera_folder, files: [file])
--- a/app/controllers/numismatics/monograms_controller.rb
+++ b/app/controllers/numismatics/monograms_controller.rb
@@ -28,8 +28,6 @@ module Numismatics
           render json: ManifestBuilder.new(@resource).build
         end
       end
-    rescue Valkyrie::Persistence::ObjectNotFoundError
-      render json: { message: "No manifest found for #{params[:id]}" }
     end
 
     private
--- a/app/controllers/playlists_controller.rb
+++ b/app/controllers/playlists_controller.rb
@@ -29,8 +29,6 @@ class PlaylistsController < ResourcesController
         render json: manifest_builder.build
       end
     end
-  rescue Valkyrie::Persistence::ObjectNotFoundError
-    render json: { message: "No manifest found for #{params[:id]}" }
   end
 
   # View the structural metadata for a given repository resource
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -6,6 +6,7 @@ class ResourcesController < ApplicationController
   include TokenAuth
   include Blacklight::SearchContext
   before_action :load_collections, only: [:new, :edit, :update, :create]
+  before_action :validate_resource_class, only: [:manifest, :file_manager, :order_manager, :structure, :pdf]
   delegate :metadata_adapter, to: :change_set_persister
   delegate :persister, :query_service, to: :metadata_adapter
   skip_before_action :verify_authenticity_token, only: [:refresh_remote_metadata]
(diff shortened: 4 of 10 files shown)


spec/controllers/ephemera_folders_controller_spec.rb
```diff
@@ -169,7 +169,7 @@
 
         get :manifest, params: { id: resource.id.to_s, format: :json }
 
-        expect(response).to be_forbidden
+        expect(response.status).to eq 401
```
hackartisan, line 172: I wonder if the context on this test should say, "when not signed in" -- the test on an anonymous user? in which case 401 seems right

hackartisan (changes requested):
this looks good, I left one small comment below

spec/controllers/ephemera_folders_controller_spec.rb
eliotjordan: The context does say "when not signed in as an admin". I didn't change anything in this test other than to explicitly test the response status. I'm hesitant to change the context wording - I did not write it and I think it makes sense in contrast to it's counterpoint on line 138. Is the issue here "admin"?

hackartisan: I'm asking because this appears to change the response code; forbidden is 403, right? [401 is unauthorized](http://www.railsstatuscodes.com/unauthorized.html). If the test is for "not signed in at all," which I think it maybe is, then 401 seems right. But if it's like, "when signed in as someone, but not an admin," then I think keeping it as 403 is probably right.

eliotjordan: I'm not sure why this PR uncovered this, but this is the currnet logic of the application controller/

Unauthorized if can be discovered:
https://github.com/pulibrary/figgy/blob/c39e1190e42b5bf498afcc6c1aa603d7f76b7ad7/app/controllers/application_controller.rb#L58

It is readable and therefore discoverable:
https://github.com/pulibrary/figgy/blob/c39e1190e42b5bf498afcc6c1aa603d7f76b7ad7/app/models/ability.rb#L205-L219

Folders have "needs_qa" as public readable state:
https://github.com/pulibrary/figgy/blob/c39e1190e42b5bf498afcc6c1aa603d7f76b7ad7/app/models/folder_workflow.rb#L31

```
All states must be included here because any state is viewable if its container allows it
```

I think the manifest denial method invoke can? :manifest instead of can? :discover.

```
  def manifest_denial
    if current_ability.can?(:manifest, @resource)
      head :unauthorized
```

hackartisan: So it was always unauthorized, but the `be_forbidden` test was passing? 😕 I mean, I'll approve, I never thought we'd spend this much time on the question, weird though!

eliotjordan: Ugh. It's failing again. Whack-a-mole with these tests.

eliotjordan: Oh yeah. Some of our tests are flat-out broken, but they passed anyway.

hackartisan (approved):

PR #6696, opened 2025-05-09
Implements Jobs and Rake Tasks for Ingesting DSpace Collections

jrgriffiniii:
Please note that this is being opened to welcome feedback, but I please apologize for the insufficient test coverage and duplication of WebMock stubbing between test suites (I am working to resolve these problems now)


lib/tasks/dspace.rake
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
```
jrgriffiniii, line 1: I apologize for including this if it is extraneous; please let me know if it is preferred that I remove this from the PR.

tpendragon: No I think Figgy's rubocop forces it right now.

tpendragon (approved):
I think this code is quite nice and well isolated, and if it's working for ingest then great!

My main feedback is that if you were interested in refactoring this I'd separate out the dspace parsing into an enumerator, such that you had a service that you could just loop through and get the metadata you need back, rather than putting all that in the ingest code. I think that'd make your ingest code come together a little nicer, might simplify some processing back and forth, and would result in a nice seam to apply some of that pre-calculated DSpace -> MMS-ID cache we talked about a little before.

PR #6744, opened 2025-07-11
Return IIIF refs even when no thumbnail is set

### LINKED ISSUE #6740: MapSets aren't syncing to GeoBlacklight after bulk ingest
After creating a MapSet with bulk ingest tool, a default thumbnail is not set. This prevents the geoblacklight IIIF link from being generated.

eliotjordan:
Closes #6740


hackartisan (commented):
This seems good, I guess I wonder if some day we will wish that there were a default thumbnail set on the resource itself, rather than leaving that blank and filling it in at index time (I think that's what this document is used for)?

tpendragon (approved):

Issue #6381, opened 2024-05-06
Improve PDF Derivative Generation


tpendragon:
Creating files that are members on the resource has made it difficult to clean them up, impossible to have two PDFs in a resource, and has caused a number of race conditions sometimes resulting in duplicate files.

We should have some way to create and attach derivative pages to a PDF file set, and render those in the manifest - much like we do for HLS manifest partials.

## Success Criteria

Creating a PDF ingest results in a viewer experience for the PDF in UV, but does not display the individual pages in the order manager or structure editor.

tpendragon:
https://github.com/pulibrary/figgy/tree/new_pdf_derivatives has code in it that gets us most of the way there by generating pages directly to pyramidal tiffs and uploading them to our derivative storage, then storing references to those pages in the FileSet as DerivativePartial FileMetadata nodes, then adjusting the manifest builder to build a canvas for those DerivativePartials when it builds canvases for a PDF.

I think that setup has a lot of potential and will allow for multiple PDFs on a single resource as well as much faster PDF derivative generation, but there's a bunch of edge cases that would crop up. Some of them are:

1. Folks want to generate OCR for pages pulled from PDFs. How do we do that here?
2. How should the structure work? Do we automatically create a structure node for those PDFs, or let them organize PDFs the same as they do any other node? If the PDF's structurable (like set PDF 1 to be under a node "Chapter 1"), do we need to make the IIIF ranges smart enough to condense all the pages into the range?
3. What needs to change about manifest IDs?
4. How do "downloads" in the manifest work?

PR #6746, opened 2025-07-25
WIP: Create PDF derivatives as "derivative partials" on a PDF FileSet.

tpendragon:
This would:

1. Allow for multiple PDFs in a resource.
2. Prevent folks from adding structure (which is super fragile and would be lost on recreating derivatives) to PDFs.
3. This seems to speed up PDF derivative generation quite a lot - I think because it goes directly from the PDF to pyramidal tiffs with no need for characterization or more derivative jobs.

Current challenges:

1. How to keep this backwards compatible
2. How to still allow for per-page OCR.
4. How should the structure manager work here? Put all canvases from that file set in the node you put it in?

### DIFF
--- a/app/derivative_services/pdf_derivative_service.rb
+++ b/app/derivative_services/pdf_derivative_service.rb
@@ -65,10 +65,10 @@ class PDFDerivativeService
   end
 
   # Delete the filesets that were generated from the pdf.
-  # TODO: Also cleanup the old versions if they're around.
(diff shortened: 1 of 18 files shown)


tpendragon:
We can't really take on the edge cases this would introduce without having a Figgy cycle. Added to this ticket: https://github.com/pulibrary/figgy/issues/6381

Issue #6741, opened 2025-07-09
Migrate DSpace Monographic Reports/Serials to Figgy


tpendragon:
### Summary or User Story

DSpace is being decommissioned. Monographic reports and serials are currently added to DSpace and will need a new home. They're going to get added to Figgy, so we need to migrate all the old ones. Most of these will have MMS-IDs, the ones that don't will be migrated as title only scanned resources and mapped later.

### Acceptance Criteria

- [ ] https://dataspace.princeton.edu/handle/88435/dsp016q182k16g migrated
- [ ] https://dataspace.princeton.edu/handle/88435/dsp01bg257f09p migrated
- [ ] https://dataspace.princeton.edu/handle/88435/dsp01kh04dp74g migrated
- [ ] https://dataspace.princeton.edu/handle/88435/dsp01r781wg06f migrated

### First step

Write a script to build a bulk ingest folder structure for these DSpace Collections

kelea99:
@tpendragon @hectorcorrea @jrgriffiniii 

# Figgy Collections for Migration

- [ ] [Monographic reports and papers (Publicly Accessible)](https://figgy.princeton.edu/catalog/88244e26-b6c4-44d1-9ffb-66a9acf69a61)
- [ ] [Monographic reports and papers (Access Limited to Princeton)](https://figgy.princeton.edu/catalog/014a0b42-3756-43f8-961d-72693f3c2a64)
- [ ] [Serials and series reports (Publicly Accessible)](https://figgy.princeton.edu/catalog/92e43f70-1afd-474f-a508-aa7c57935de4)
- [ ] [Serials and series reports (Access Limited to Princeton)](https://figgy.princeton.edu/catalog/700c90ac-17bd-422c-b5c2-dd60b6c2571d)

PR #6751, opened 2025-07-29
Improve figgy_metadata.json handling in BulkIngest.

tpendragon:
This allows volumes to have `figgy_metadata.json` files and makes it
valid to add `source_metadata_identifier` as a property that gets
imported, for cases when the folder can't be named after the volume
(like when the volume must be ordered via name sort, but has an MMS-ID)

It also ensures the title can be set independently from the folder name,
for cases when the title might be too long to be a folder name - as is
the case in several DSpace monographs.

Work towards #6741

### DIFF
--- a/spec/services/bulk_ingest_service_spec.rb
+++ b/spec/services/bulk_ingest_service_spec.rb
@@ -259,7 +259,8 @@ RSpec.describe BulkIngestService do
         expect(resource.viewing_hint).to eq ["paged"] # brought in from figgy_metadata.json
         expect(resource.member_ids.length).to eq 2 # color.tif, gray.tif
         expect(resource.depositor).to eq ["tpend"]
-        expect(resource.title).not_to eq ["My Title"] # brought in from figgy_metadata.json
+        expect(resource.title.first.to_s).to eq "Bible, Latin." # Imported from source_metadata_identifier.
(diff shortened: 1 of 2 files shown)


spec/services/bulk_ingest_service_spec.rb
```diff
@@ -257,10 +259,57 @@
         expect(resource.viewing_hint).to eq ["paged"] # brought in from figgy_metadata.json
         expect(resource.member_ids.length).to eq 2 # color.tif, gray.tif
         expect(resource.depositor).to eq ["tpend"]
+        expect(resource.title).not_to eq ["My Title"] # brought in from figgy_metadata.json
```
hackartisan, line 262: This might be more clear if you test for the value that will overwrite it (looks like `Bible, Latin.`) and comment that it's overwriting the value that was in figgy_metadata.json

hackartisan: Oh, copy line 307 here!

hackartisan (changes requested):
This looks good, nice feature, one comment and should we also update https://github.com/pulibrary/figgy/blob/main/docs/ux/use_cases.md#figgy_metadatajson ?

hackartisan (approved):

PR #6759, opened 2025-08-05
Downloads Controller should handle fallbacks.

### LINKED ISSUE #6758: Fallback isn't working in the downloads controller for stream derivatives
### Summary or User Story

When I moved everything over to Tigerdata the video streams broke - it looks like the fallback adapter doesn't hit the fallback if it finds a stream ID, probably due to the ID searching functionality in adapters. So it hits something with the stream prefix and then finds the non-fallback adapter.

### Acceptance Criteria

- [ ] Fallback works in the downloads controller for stream derivatives.

### First step

Write a test.

tpendragon:
Closes #6758

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -73,7 +73,7 @@ class DownloadsController < ApplicationController
 
   def binary_file
     return unless file_desc
-    @binary_file ||= storage_adapter.find_by(id: file_desc.file_identifiers.first)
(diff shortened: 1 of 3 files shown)


hackartisan:
Just copying this here from slack:

> Two bugs - the downloads controller was explicitly calling the disk storage adapter so it was never even asking the stream fallback adapter, and then when it was set to search for the right adapter for an ID it'd find the first one registered - which was the primary adapter, not the one that can fallback.

PR #6781, opened 2025-08-25
Cleanup geo derivative working files

### LINKED ISSUE #6780: Vector derivative generation does not clean up after itself
### Summary or User Story

Vector (possibly raster?) derivative generation does not clean up after itself. This leaves potentially large intermediate files around that fill up /tmp.

### Acceptance Criteria

This section can be completed by the DLS team. Some requirements to consider
including:

- [ ] When a vector derivate job is completed, it does not leave files in the temp directory.
- [ ] When a vector derivate job errors, it does not leave files in the temp directory.

### First step

Figure which files are created (I think an intermediate json in GDAL and then pmtiles' output).
Figure how to let the job know which files to clean up. Maybe create a special dir for them?

eliotjordan:
Closes #6780

Ensures that vector and raster derivative generation does not leave files on the worker.

- Creates a temp directory for all output and intermediate steps.
- Ensures that temp files and folders are removed when processing is done or an error is encountered.

### DIFF
--- a/app/derivative_services/raster_resource_derivative_service.rb
+++ b/app/derivative_services/raster_resource_derivative_service.rb
@@ -63,6 +63,11 @@ class RasterResourceDerivativeService
       update_error_message(message: error.message)
     end
     raise error
+
+  ensure
+    FileUtils.rmtree(temporary_working_directory) if Dir.exist?(temporary_working_directory)
+    File.unlink(temporary_display_output.path) if File.exist?(temporary_display_output.path)
+    File.unlink(temporary_thumbnail_output.path) if File.exist?(temporary_thumbnail_output.path)
   end
 
   def cloud_storage_adapter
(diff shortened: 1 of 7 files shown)


app/derivative_services/raster_resource_derivative_service.rb
```diff
@@ -30,7 +30,7 @@ def change_set
 
   def build_cloud_file
     IngestableFile.new(file_path: temporary_display_output.path, mime_type: "image/tiff; gdal-format=GTiff", original_filename: "display_raster.tif", use: use_cloud_derivative,
-                       copy_before_ingest: false)
+                       copy_before_ingest: true)
```
tpendragon, line 33: Why'd this have to change?

tpendragon (changes requested):
Looks good, just one question - can you provide some info on why `copy_before_ingest` had to change? I don't really understand.

app/derivative_services/raster_resource_derivative_service.rb
tpendragon: Looks like when this is `true` it won't create an interstitial file, which is okay for derivatives because we just want to move it directly to the correct place, and don't care if the old file disappears (unlike for the repository.)

tpendragon (approved):

Issue #6790, opened 2025-08-28
Problem with small images and serverless IIIF


eliotjordan:
- https://figgy.princeton.edu/catalog/c8451243-561c-4afc-bde2-9da5a7913d01
- https://digital-collections.princeton.edu/dev/oban/jobs/105533
- https://iiif-cloud.princeton.edu/iiif/2/9c%2F2c%2F32%2F9c2c3233fbd54a18b0d35663136b24d0%2Fintermediate_file/full/299,/0/default.jpg

eliotjordan:
The file that's raising this particular error:
https://figgy.princeton.edu/catalog/c8451243-561c-4afc-bde2-9da5a7913d01

Output of tiffinfo on the pyramidal tiff derivative:
```
=== TIFF directory 0 ===
TIFF Directory at offset 0x29d14 (171284)
  Image Width: 299 Image Length: 383
  Tile Width: 1024 Tile Length: 1024
  Resolution: 96, 96 pixels/inch
  Bits/Sample: 8
  Sample Format: unsigned integer
  Compression Scheme: JPEG
  Photometric Interpretation: RGB color
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 3
  Planar Configuration: single image plane
  JPEG Tables: (289 bytes)

**The image width and height is less than our tile width.**

As opposed to this file:
https://figgy.princeton.edu/catalog/de80148e-95f1-411e-8eb2-f120f68fd493

Where we get this output from tiffinfo:

```
TIFF Directory at offset 0x2510b6 (2429110)
  Image Width: 1232 Image Length: 1530
  Tile Width: 1024 Tile Length: 1024
  Resolution: 37.7953, 37.7953 pixels/cm
  Bits/Sample: 8
  Sample Format: unsigned integer
  Compression Scheme: JPEG
  Photometric Interpretation: RGB color
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 3
  Planar Configuration: single image plane
  JPEG Tables: (289 bytes)

=== TIFF directory 1 ===
TIFF Directory at offset 0x2f3c30 (3095600)
  Subfile Type: reduced-resolution image (1 = 0x1)
  Image Width: 616 Image Length: 765
  Tile Width: 1024 Tile Length: 1024
  Resolution: 37.7953, 37.7953 pixels/cm
  Bits/Sample: 8
  Compression Scheme: JPEG
  Photometric Interpretation: RGB color
[...]

PR #6791, opened 2025-08-29
Increase derivative image size if original image is smaller than pyramidal tile size

eliotjordan:
Clsoes #6790 

- **Reduce derivative image size more accurately**
- **Increase derivative image size if original image is smaller than pyramidal tile size**
- **Refactor image resizing**

### DIFF
--- a/app/derivative_services/vips_derivative_service.rb
+++ b/app/derivative_services/vips_derivative_service.rb
@@ -3,6 +3,7 @@ class VipsDerivativeService
   # Pixel width or height at which point it cuts the size in half for
   # performance.
   REDUCTION_THRESHOLD = 15_000
+  TILE_SIZE = 1024
   class Factory
     attr_reader :change_set_persister
     delegate :metadata_adapter, to: :change_set_persister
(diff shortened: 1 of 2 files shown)


eliotjordan:
Tried it with a small image on staging:

Processing on main branch. IIIF not working:
https://figgy-staging.princeton.edu/catalog/9e053b9e-87dd-4311-99af-a3089c8ae823

Processing on updated branch. IIIF working:
https://figgy-staging.princeton.edu/catalog/43a6e580-63f1-45ba-a790-f5ddaa990658

app/derivative_services/vips_derivative_service.rb
```diff
@@ -98,14 +99,25 @@ def vips_image
...
-          image.resize(0.5)
-        else
-          image
-        end
+        resize(image)
       end
   end
 
+  def resize(image)
+    if image.height >= REDUCTION_THRESHOLD || image.width >= REDUCTION_THRESHOLD
```
tpendragon, line 107: That's not what this code was about before - it wasn't about making all the images smaller than the given size, it was about cutting them in half if they were above it, because a bunch of testing with scrolls and stuff showed that it got performance above a certain mark while keeping the quality good enough to see the grain.

tpendragon (changes requested):
This looks good for small images, but I think we should keep the same logic for big images.

app/derivative_services/vips_derivative_service.rb
eliotjordan: Reverted

Issue #6920, opened 2025-12-02
Ensure SeleneResources can be restored from Preservation


tpendragon:
### Summary or User Story

When a SeleneResource is deleted either because its parent was deleted or it was deleted by itself, we should make sure it has a DeletionMarker and can be restored.

### Acceptance Criteria

- [ ] There's a test that shows a SeleneResource is restored when a ScannedResource w/ FileSet w/ SeleneResource is deleted and restored.
- [x] There's a test that shows a SeleneResource is restorable when it's deleted off a FileSet.

### First step

Write the tests.

PR #6934, opened 2025-12-05
Ensure SeleneResources can be restored from Preservation

eliotjordan:
advances #6920


hackartisan (approved):
This looks good, I think it fulfills one of the 2 acceptance criteria for the issue so I'm going to change the ref so it won't close the issue, and then merge

PR #6954, opened 2025-12-12
Extract data from TFW file on creation of SeleneResource

### LINKED ISSUE #6947: Extract data from TFW file on creation of SeleneResource
### Summary or User Story

When creating the SeleneResource it should find a <ingest_directory>/Selene_Output/depthmap_m1.tfw file in the ingest directory then parse the first line and set that value to the SeleneResource's meters_per_pixel property. A TFW looks like this:

```
0.00002555
0
0
-0.00002555
-0.0
-0.0
```

### Acceptance Criteria

- [ ] TFW file is found and the meters_per_pixel property is saved onto the resource.

### First step

Write a test

eliotjordan:
Closes #6947 

- **Extract data from TFW file on creation of SeleneResource**
- **Allow any tfw file to be used for meters_per_pixel value**

### DIFF
--- a/app/services/selene_ingest_path_service.rb
+++ b/app/services/selene_ingest_path_service.rb
@@ -10,16 +10,16 @@ class SeleneIngestPathService
     selene_structure.none?(&:nil?)
   end
 
-  # Extract meters per pixel from depthmap tfw file
+  # Extract meters per pixel from a tfw file
   def meters_per_pixel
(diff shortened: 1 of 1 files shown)


tpendragon (commented):
Just a question - I think this means if there isn't a TFW file, the user gets a 500 page?

PR #6964, opened 2025-12-15
Allow Selene files with upper or lower case file extensions

### LINKED ISSUE #6961: Ensure light angle metadata is captured and displayed on Selene file sets
### Summary or User Story

We did a test ingest in prod and do not see this data:
https://figgy.princeton.edu/catalog/effaf1dc-d229-47f6-9801-d378c4484425

### Acceptance Criteria

- [ ] Data is captured and displayed.

### First step

Replicate locally.

eliotjordan:
Closes #6961


hackartisan (dismissed):
oh, good catch, i noticed this issue on one of the numbered primary files, but not on these. 

I wonder if there's a way to make the whole regex case insensitive, would that be nicer to read? not blocking though.

tpendragon:
> I wonder if there's a way to make the whole regex case insensitive, would that be nicer to read? not blocking though.

I wonder if you can append `/i` after the regex like in vim?

PR #6966, opened 2025-12-16
Set state and visibility for new SeleneResources

### LINKED ISSUE #6960: Propagate Selene grandparent visibility and state
### Summary or User Story

Insure the the visibility and state of a Selene Resource's grandparent is propagated during creation.

### Acceptance Criteria
- [ ] A new SeleneResource will have state and visiblity from parent

### First step

Write a test

eliotjordan:
Closes #6960

### DIFF
--- a/app/controllers/scanned_resources_controller.rb
+++ b/app/controllers/scanned_resources_controller.rb
@@ -2,6 +2,7 @@
 class ScannedResourcesController < ResourcesController
   self.resource_class = ScannedResource
   self.change_set_persister = ChangeSetPersister.default
+  before_action :set_selene_visiblity_and_state, only: [:new]
   before_action :validate_selene_ingest_path, only: [:create]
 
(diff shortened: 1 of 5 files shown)


app/controllers/scanned_resources_controller.rb
```diff
@@ -13,6 +14,13 @@ def after_create_success(obj, change_set)
     handle_upload_selene_files(obj)
   end
 
+  def propagate_selene_visiblity_and_state
```
hackartisan: Okay, I think the way this works is that it pre-fills the edit form for the new resource with state and visibility values. It would be nice to have a comment on this method saying that's why we have it.

eliotjordan: Added. And changed the method name to `set_selene_visiblity_and_state`

hackartisan (approved):

Issue #6988, opened 2026-01-07
Honeybadger is not reporting stack traces


eliotjordan:
Honeybadger is not reporting stack traces from errors in Sidekiq jobs since upgrading to ruby 3.4

https://app.honeybadger.io/projects/53391/faults/126385979

<img width="950" height="598" alt="Image" src="https://github.com/user-attachments/assets/9a3faac0-a524-4e69-b7c9-7153dbe08301" />
<img width="1112" height="854" alt="Image" src="https://github.com/user-attachments/assets/34177d0c-8b02-4ccc-861e-d9340e5831c9" />

PR #6989, opened 2026-01-07
Update honeybadger and sidekiq

eliotjordan:
Work towards #6988

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -337,7 +337,7 @@ GEM
       execjs
     coffee-script-source (1.12.2)
     concurrent-ruby (1.3.6)
-    connection_pool (3.0.2)
+    connection_pool (2.5.5)
     content_disposition (1.0.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -337,7 +337,7 @@ GEM
       execjs
     coffee-script-source (1.12.2)
     concurrent-ruby (1.3.6)
-    connection_pool (2.5.5)
+    connection_pool (3.0.2)
```
hackartisan: This might be an issue? It was the problem blocking that dpul PR. see https://github.com/pulibrary/dpul/pull/1636/commits/e8761b2175bbf02cd7500266f18c896671e1600f and https://github.com/pulibrary/dpul/issues/1660#issuecomment-3715258466

eliotjordan: Pushed a new commit to pin connection_pool.

PR #6942, opened 2025-12-09
Check the size of files before accessing them to detect Tigerdata down.

tpendragon:
Closes #6941

PR #7022, opened 2026-01-22
Remove timeouts from disk adapters

### LINKED ISSUE #7014: Remove timeout code introduced in #6952
### Summary or User Story

It doesn't work and we're getting timeouts in Honeybadger, which may be failing requests for users.

#6952 is where it was introduced.

We need to maintain trying to read one byte (so that fallback works if it's a database entry in mediaflux but there's no connection to the Isilon), but don't need to wrap it in a Timeout.

### Acceptance Criteria

- [ ] The code that does `Timeout.` is gone.

### First step

Look at that PR to see what to remove.

### Sudden Priority Justification

Users are possibly getting requests cut off.

hackartisan:
Takes code functionality back to before #6952 and #6942
closes #7014

### DIFF
--- a/app/adapters/fallback_disk_adapter.rb
+++ b/app/adapters/fallback_disk_adapter.rb
@@ -7,8 +7,15 @@ class FallbackDiskAdapter
   end
 
   def find_by(id:)
-    primary_adapter.find_by(id: id)
-  rescue Valkyrie::StorageAdapter::FileNotFound
+    resource = primary_adapter.find_by(id: id)
+    # Check to see if the file's readable, fallback if not.
+    # If Tigerdata doesn't have a solid connection to the Isilon, this will fail
+    File.open(resource.disk_path, "rb") do |file|
+      # Read just one byte.
+      file.read(1)
+    end
+    resource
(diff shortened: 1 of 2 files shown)


tpendragon (changes requested):

app/adapters/fallback_disk_adapter.rb
```diff
@@ -8,15 +8,9 @@ def initialize(primary_adapter:, fallback_adapter:)
 
   def find_by(id:)
     resource = primary_adapter.find_by(id: id)
-    # Check to see if the file's readable, fallback if not.
-    # If Tigerdata doesn't have a solid connection to the Isilon, this will
```
tpendragon: Should we keep this part of the comment?

tpendragon (approved):
Looks good, just a small suggestion.

app/adapters/fallback_disk_adapter.rb
hackartisan: oh, ha, yes, good idea

tpendragon (approved):

PR #7027, opened 2026-01-29
Defer preservation of binary content until after characterization.

### LINKED ISSUE #7025: Preservation should not upload a file to gcs if its checksum does not match characterization data
### Summary or User Story

Currently, a preservation object is created at the time of preservation. the checksum for the preservation object binary node is re-calculated from the file on disk at that time:

https://github.com/pulibrary/figgy/blob/fde186a1102fd631e93d91dc0868ed3d8c58e4d2/app/services/preserver.rb#L56

This newly-calculated checksum is not compared against the characterization data.

In https://github.com/pulibrary/figgy/issues/7013 we see instances where a file uploaded fine to gcs, with gcs confirming the uploaded checksum matches the one we provided, but the file does not match the file in the repository. 

PreserveResourceJob should fail before uploading to google if the computed checksum does not match the checksum calculated during characterization. This could be achieved either by simply using that characterization checksum for the preservation object, or by comparing it to a newly-calculated checksum before starting the upload.

### Acceptance Criteria

- [ ] A file whose checksum doesn't match the checksum found during characterization does not get uploaded to gcs
- [ ] A preservation object chose checksum doesn't match the checksum found during chracterization does not get created and saved

### Sudden Priority justification

The failures as seen in #7013 are continuing to be created; until we keep them from happening to begin with we will have to repeatedly repair the ones we have.

### First step

write a test

tpendragon:
Closes #7025

### DIFF
--- a/spec/change_set_persisters/change_set_persister_spec.rb
+++ b/spec/change_set_persisters/change_set_persister_spec.rb
@@ -1365,7 +1365,7 @@ RSpec.describe ChangeSetPersister do
 
         change_set_persister.save(change_set: change_set)
         expect(PreserveChildrenJob).to have_received(:perform_later).exactly(1).times
-        expect(Preserver).to have_received(:new).exactly(5).times
+        expect(Preserver).to have_received(:new).exactly(4).times
         expect(CleanupFilesJob).not_to have_received(:perform_later)
       end
     end


hackartisan (dismissed):
This looks really good, just one more suggestion about the initial test, thank you!

PR #7078, opened 2026-02-28
Search within using postgres

tpendragon:
## Reasoning

This would let us stop using another dependency (Blacklight IIIF Search), it seems like it's faster than our old implementation, we don't have to build another index, and we don't have to index FileSets.

If this seems okay, I'll clean up the proof of concept and see about getting it going.

This also significantly improves boundary detection for coordinates and phrase highlighting.


hackartisan (approved):
I like these revisions, thanks!

PR #7133, opened 2026-04-20
Use ignore-scripts in CI/CD

hackartisan:
refs pulibrary/princeton_ansible#7007


hackartisan:
Seems like figgy might need these for mapnik?

hackartisan:
yeah, we can't do this on figgy.

see https://github.com/mapnik/node-mapnik/blob/36609d82a111ed9e72de058f19e166a98b12dfe9/package.json#L58

Issue #7131, opened 2026-04-14
Raster failing characterization


tpendragon:
### Summary or User Story

A raster file was uploaded here: https://figgy.princeton.edu/catalog/e3581310-bdfa-4558-b159-99e03c507c16 and is failing to characterize.

https://app.honeybadger.io/projects/53391/faults/127213536

### Acceptance Criteria

- [ ] Rasters can characterize

### Sudden Priority Justification

It's affecting our staff's work if they're unable to reliably upload Raster files.

### First step

Download the file, run some tests.

eliotjordan:
More rasters failing characterization.

932e8e5a-2407-4356-8641-dcf09c52ea58
509222b1-1c58-481b-87ec-cbfa7eeb8148
6d2bb51a-11c7-41a6-ae71-6845f986b4a1

hackartisan:
Here's another one https://figgy.princeton.edu/catalog/f72d492b-57c9-4ce9-bfc0-4590a07ffc37, ingested today (parent resource [MCE_Global 2G mobile coverage, 2022](https://figgy.princeton.edu/catalog/44375e5b-80ef-4a40-a1df-6cd96f498676))

eliotjordan:
These are all re-characterized and have derivatives.

PR #7137, opened 2026-04-22
Skip Tika characterization for raster and vector data

eliotjordan:
Works towards #7131 

Follow up is to recharacterize the datasets in the ticket.

### DIFF
--- a/app/characterization_services/external_metadata_characterization_service.rb
+++ b/app/characterization_services/external_metadata_characterization_service.rb
@@ -18,6 +18,8 @@ class ExternalMetadataCharacterizationService
   #   Valkyrie::Derivatives::FileCharacterizationService.for(file_set, persister).characterize(save: false)
   def characterize(save: true)
     primary_file.mime_type = geo_mime_type
+    primary_file.checksum = MultiChecksum.for(file_object)
+    primary_file.size = File.size(file_object.disk_path)
     @file_set = @persister.save(resource: @file_set) if save
     @file_set
   end
@@ -45,15 +47,15 @@ class ExternalMetadataCharacterizationService
     @document ||= Nokogiri::XML(file_object.read)
   end
 
-  def file_object
-    @file_object ||= Valkyrie::StorageAdapter.find_by(id: primary_file.file_identifiers[0])
-  end
-
   def primary_file
     @file_set.primary_file
   end
 
+  def file_object
+    @file_object ||= Valkyrie::StorageAdapter.find_by(id: primary_file.file_identifiers[0])
+  end
+
   def valid?
-    mime_type == ["application/xml"]
+    primary_file.extname == ".xml"
   end
 end
--- a/app/characterization_services/gdal_characterization_service.rb
+++ b/app/characterization_services/gdal_characterization_service.rb
@@ -69,10 +69,15 @@ class GdalCharacterizationService
     @dataset_path = zip_file_directory
   end
 
+  # Tests if primary file is an xml file
+  def xml_file?
+    primary_file.extname == ".xml"
+  end
+
   # Tests if primary file is a zip file
   # @return [Boolean]
   def zip_file?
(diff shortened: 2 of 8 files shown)


app/characterization_services/external_metadata_characterization_service.rb
```diff
@@ -45,15 +47,19 @@ def document
...
+  end
+
   def file_object
     @file_object ||= Valkyrie::StorageAdapter.find_by(id: primary_file.file_identifiers[0])
   end
 
-  def primary_file
-    @file_set.primary_file
+  def filename
+    Pathname.new(file_object.io.path) if file_object.io.respond_to?(:path) && File.exist?(file_object.io.path)
```
tpendragon: Is there a reason we can't do `file_object.disk_path`?

app/characterization_services/gdal_characterization_service.rb
```diff
@@ -69,10 +69,15 @@ def unzip_original_file
     @dataset_path = zip_file_directory
   end
 
+  # Tests if primary file is an xml file
+  def xml_file?
+    File.extname(primary_file.file_identifiers[0].to_s) == ".xml"
```
tpendragon: Suggestion: Do we need `FileMetadata#extname`?

app/characterization_services/geo_characterization_service.rb
```diff
@@ -16,8 +16,11 @@ def initialize(file_set:, persister:)
   # @example characterize a file and do not persist the changes
   #   Valkyrie::Derivatives::FileCharacterizationService.for(file_set, persister).characterize(save: false)
   def characterize(save: true)
-    @file_set = TikaFileCharacterizationService.new(file_set: file_set, persister: persister).characterize
-    @file_set = scanned_map_characterization_service.characterize if scanned_map_characterization_service.valid?
+    if scanned_map_characterization_service.valid?
+      # Run tika on scanned maps only
+      @file_set = TikaFileCharacterizationService.new(file_set: file_set, persister: persister).characterize
```
tpendragon, line 21: Curious, why do scanned maps need Tika? Or this is just to keep the change smaller?

tpendragon (changes requested):
Just a few questions. Excited to run Tika less!

app/characterization_services/geo_characterization_service.rb
eliotjordan: Yeah, just to keep the changes smaller.

app/characterization_services/external_metadata_characterization_service.rb
eliotjordan: I copied that method from [other characterization services](https://github.com/pulibrary/figgy/blob/main/app/characterization_services/gdal_characterization_service.rb#L47), but, yes, `file_object.disk_path` is definitely better.

app/characterization_services/gdal_characterization_service.rb
eliotjordan: @tpendragon I'm not sure what you mean by this comment.  FileMetadata doesn't have extname method.

tpendragon: I'm saying should we add one, so you can do `primary_file.extname == ".xml"`, or add `FileMetadata#xml?` so you can do `primary_file.xml?`

eliotjordan: Ah... I misread that.  That's a good idea.

eliotjordan: Added a FileMetadata#extname method.

tpendragon: I think you used it in External, but not here.

tpendragon (changes requested):
Just one more spot to use the new extname method.

eliotjordan:
@tpendragon extname method added to gdal characterization service.

tpendragon (approved):

PR #7166, opened 2026-05-26
Collection banner url selector

eliotjordan:
- **Add banner image url to collection model and changeset**
- **Add banner cropping to collection form**
- **Upgrades OSD from v2 to v5, adds Annotorious js library, and adds cropping square to the viewport**
- **Build a crop selector widget**
- **Use OpenSeadragonAreaSelector plugin**
- **Fix collection tests**
- **Add a clear button**
- **Make the hidden field id a component property**
- **Display image banner on Collection show page**
- **Move buttons to admin contols bar to improve display and UX**

image

**Add banner preview to show page:**

image

**Move buttons to admin contols bar to improve display and UX:**

image


tpendragon (changes requested):
This feature works great, I'm super impressed with what y'all were able to do - good work!

Before we merge I think we should have a couple of tests to prevent regression - especially given that this feature won't get used constantly, so we won't hear about it right away if it breaks in an upgrade:

1. A test that the cropping tool appears, to prevent us not knowing when the javascript breaks over time.
2. A test that an edit of the collection with JS (so probably a capybara test that edits a collection that has a banner, saves it, and makes sure the banner image is still there) doesn't override that banner URL - if that widget breaks there's a chance that might happen, and I want to catch it in tests so we don't lose that data.
3. A test of the parseUrl function, so we have a place to iterate as we support more URLs in that space or new versions of IIIF.

sdellis:
@tpendragon We added the tests you requested. It's ready for another review.

tpendragon (approved):

PR #7187, opened 2026-06-09
Add an auto-generated banner_image_id property

### LINKED ISSUE #7173: Add a new field for the figgy id of the item used as banner image on the collection / project
### Summary or User Story

To make https://github.com/pulibrary/dpul-collections/issues/1239 possible, we need the collection/project to also store the id of the resource the crop was made from.

Add this as another field on collection / project

### Implementation details

We think we can just add a field on the model that gets populated by the changeset from the iiif url.

pseudocode something like:

- 60%2Fb5%2Fe5%2F60b5e5365600450db52dbe4d7f92b8cc%2Fintermediate_file
- urldecode to 60/b5/e5/60b5e5365600450db52dbe4d7f92b8cc/intermediate_file
- 60b5e5365600450db52dbe4d7f92b8cc
- split that to get the ID for the derivative tif
- then get its parent, that’s the fileset
  - (this would need a query - we don’t know if that query exists already)
- then get its parent, that’s the resource (use a wayfinder)

### Acceptance Criteria

- [ ] There's a new field on collections / ephemera projects that stores the resouce id of the item used to generate the banner image crop

### First step

Write a test

eliotjordan:
Closes #7173

- Adds banner_image_id to Collection and EphemeraProject
- Adds a change set persister to calculate the figgy id of banner image parent resource

### DIFF
--- a/app/change_set_persisters/change_set_persister.rb
+++ b/app/change_set_persisters/change_set_persister.rb
@@ -37,7 +37,8 @@ class ChangeSetPersister
         CreateProxyFileSets,
         ApplyAuthToken,
         CacheParentId,
-        UpdateCloudFilePermissions
+        UpdateCloudFilePermissions,
+        ApplyBannerImageId
       ],
       after_save: [
         UpdateAuthToken,
--- a/app/change_set_persisters/change_set_persister/apply_banner_image_id.rb
+++ b/app/change_set_persisters/change_set_persister/apply_banner_image_id.rb
@@ -1,0 +1,41 @@
+class ChangeSetPersister
+  class ApplyBannerImageId
+    attr_reader :change_set_persister, :change_set
+    delegate :query_service, :persister, to: :change_set_persister
+    def initialize(change_set_persister:, change_set:, post_save_resource: nil)
+      @change_set = change_set
+      @change_set_persister = change_set_persister
+    end
+
+    def run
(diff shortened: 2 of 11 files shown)


app/queries/find_by_file_metadata_id.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  attr_reader :query_service
+  delegate :resource_factory, to: :query_service
+  delegate :run_query, to: :query_service
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def find_by_file_metadata_id(id:)
+    run_query(query, { file_metadata: [{ id: { id: id.to_s } }] }.to_json).first
```
tpendragon: Suggestion: I think you could do `query_service.custom_queries.find_by_property(property: :file_metadata, value: {id: Valkyrie:::ID.new(id.to_s)}, model: FileSet]`

tpendragon (changes requested):
Just a small query suggestion that I think might make this easier.

sdellis (dismissed):
Looks great! 👍

app/queries/find_by_file_metadata_id.rb
eliotjordan: That's a good idea. Simpler. 👍

tpendragon (approved):

Issue #7174, opened 2026-06-02
Migrate usage statistics gathering from Plausible to Umami


tpendragon:
### Summary or User Story

We've moved from Plausible to Umami. The implementation we did in https://github.com/pulibrary/figgy/issues/6376 and https://github.com/pulibrary/figgy/issues/6378 should be migrated to Umami. 

### Acceptance Criteria

- [ ] We're recording downloads and viewer interactions in Umami.

### First step

Look at our old implementation and see what we can do.

Look at the [Umami Docs](https://docs.umami.is/docs)

eliotjordan:
Complete.

PR #7221, opened 2026-07-23
Instead of Plausible events, send messages to iframe parent window

eliotjordan:
Instead of Plausible events, send messages to iframe parent window that individual webapps can subscribe to and hook to analytics. This allows UV events to be logged per site.

Part of https://github.com/pulibrary/figgy/issues/7174

- **Send message to parent window on UV click**
- **Send message to parent window on download**
- **Silence uv_manager.spec errors**
- **Remove Plausible from Universal Viewer embed**
- **Add tag manager to staging**

Added figgy-staging to umami and tag manager and then added this code to the tag manager to subscribe to messages:

```
    // Listen for UniversalViewer event messages and track in umami
    window.addEventListener("message", function(event) {
      // Exit if message is not from a known origin
      var regex = /princeton\.edu/;
      if (!regex.test(event.origin)) {
        return;
      }

      var event_name = event.data.event;
      if (event_name == 'universal-viewer-click') {
        umami.track(event_name, event.data.data);
      } else if (event_name == 'universal-viewer-download') {
        umami.track(event_name, event.data.data);
      };
    });
```

Now UV click and download events are logged with Umami:
https://analytics-staging.lib.princeton.edu/websites/5ebe40c2-e7a8-453d-a934-96b14dfe0492/events

image

image

If this PR is merged, we can then add that tag code to figgy-prod, finding aids, and dpul.

### DIFF
--- a/app/javascript/test/setup.js
+++ b/app/javascript/test/setup.js
@@ -3,6 +3,10 @@ import _ from 'lodash'
 vi.unmock('lodash')
 _.debounce = vi.fn((fn) => fn);
 
+// Stub HTMLCanvasElement.prototype.getContext used in OpenSeaDragon
+// to prevent "Not implemented" errors during tests.
+HTMLCanvasElement.prototype.getContext = vi.fn()
+
 // jsdom doesn't let you mock window.location anymore, so replace that
 // implementation for all tests so we can mock it. Solution found here:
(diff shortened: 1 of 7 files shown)


tpendragon (approved):
Very cool! I'm excited to see this work.

PR #7227, opened 2026-07-27
Add pulfalight records report

### LINKED ISSUE #7220: Generate a `/reports/pulfalight_records` endpoint
### Summary or User Story

We need to get information about portion notes from Figgy into Finding Aids in a bulk-compatible way. The way we do this for the catalog is a `/reports/mms_records` report that outputs a structure that their indexing pipeline caches and uses. This will support https://github.com/pulibrary/pulfalight/issues/1371

In Pulfalight we only ever index a full collection in one background job, so we can filter our results to just that collection in order to improve performance.

### Technical Implementation

`/reports/pulfalight_records?collection=AC140` should return a hash of all records in the AC140 collection. The response should look the same as it does for the `mms_records` endpoint, so (not a real thing with a portion):

```json
[{
  "AC140_c0329": [{
    "ark": "http://arks.princeton.edu/ark:/88435/dc73666k25k",
    "iiif_manifest_url": "https://figgy.princeton.edu/concern/scanned_resources/5197778a-5f43-4a88-ae38-b3a2e7892614/manifest",
     "is_portion": true,
     "portion_note": "Part",
     "visibility": {"value": "open", "label": "open", "definition": "Open to the world. Anyone can view."}
  }]
}, ...]
```

An example of an item which has a portion note: [Alumni--Reunions, General](https://figgy.princeton.edu/catalog/bea90187-bb7a-49ea-acc8-48f7d8b7c9df)

### Acceptance Criteria

- [ ] `/reports/pulfalight_records` exists
- [ ] The report is restricted to users with a `pulfalight_sync` role (provided by auth token)

### First step

Look at the MMS Records implementation and find the queries we use for finding records by ASpace collection code.

eliotjordan:
Closes #7220

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -98,10 +98,14 @@ class ReportsController < ApplicationController
   def pulfalight_records
     authorize! :show, :pulfalight_report
 
-    respond_to do |format|
-      format.json do
-        render json: pulfalight_report(params[:collection])
+    if params[:collection]
+      respond_to do |format|
+        format.json do
+          render json: pulfalight_report(params[:collection])
+        end
       end
+    else
+      head :bad_request
     end
   end
 
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -100,7 +100,7 @@ class ReportsController < ApplicationController
 
     respond_to do |format|
       format.json do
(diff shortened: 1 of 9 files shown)


hackartisan (changes requested):
This looks great; my only question is what happens if there's no param given to the report route?

hackartisan (approved):
👍🏻

PR #7231, opened 2026-07-29
Disable portion note field until Is Portion is selected

### LINKED ISSUE #7058: Hide portion note field until Is Portion is selected
### Summary or User Story
Follow up to: https://github.com/pulibrary/figgy/pull/7057

Hide portion note field until Is Portion is selected on edit forms. This will help prevent resources with a note but no is_portion value.

### Acceptance Criteria
- [ ] Portion Note field is hidden on form until checkbox is selected.

### First step

Write a test.

eliotjordan:
Closes #7058 

image

image

### DIFF
--- a/app/change_sets/selene_resource_change_set.rb
+++ b/app/change_sets/selene_resource_change_set.rb
@@ -11,8 +11,8 @@ class SeleneResourceChangeSet < ChangeSet
   property :read_groups, multiple: true, required: false
   property :change_set, require: true, default: "selene_resource"
   property :depositor, multiple: false, require: false
-
   property :pending_uploads, multiple: true, required: false
+  property :is_portion, multiple: false, required: false, type: Valkyrie::Types::Bool
   property :portion_note, multiple: false, required: false
   property :meters_per_pixel, multiple: false, required: true, type: ::Types::Params::Float.optional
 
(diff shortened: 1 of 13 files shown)


spec/validators/portion_note_validator_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    context "when there is a portion note and is_portion is set" do
+      it "does not add errors" do
+        record = build_record(portion_note: "Only volume 2.", is_portion: true)
+        validator.validate(record)
+        expect(errors).not_to have_received(:add)
+      end
+    end
+
+    context "when there is a portion note and is_portion is not set" do
```
tpendragon: Do I actually see anything on the edit screen when this is the case? The note's hidden, will that hide the validation error?

tpendragon (changes requested):
Just one edge case I can think of.

spec/validators/portion_note_validator_spec.rb
eliotjordan: You won't see anything on the screen, that's correct. It should be impossible to trigger this in the UI without doing some work in the dev tools to unhide the portion note field.

We could just remove the the validator as it doesn't provide any useful benefit.

tpendragon (approved):

PR #7237, opened 2026-08-07
7229 highlighted

### LINKED ISSUE #7229: Highlighted items should be scoped to a collection
### Summary or User Story

To add highlighted items to a collection we added a feature in Figgy where you could mark an item as "Highlighted." However, when an item is in multiple collections our curators lose the ability to highlight items for a specific collection. They'd like to choose "these four items are highlighted for this collection."

To implement this:

In the "Access and Use" section, there should be a live updating list of checkboxes for featuring in each collection and project an item is a member of, instead of the current checkbox. Current checkbox:

Image

### Open Questions

* What will migration be like for these?
   * For every item that is highlighted, make it highlighted for every collection+project it's a member of. Then people can fix it, DPUL-C looks the same it does now.
* What's deployment like, so that DPUL-C works?
    * Deploy Figgy
    * Fix DPUL-C, Deploy
    * Run migration.
    * Rehydrate

### Acceptance Criteria

- [ ] Highlighted items can be added scoped to a specific collection.

### First step

Write some tests.

eliotjordan:
Closes #7229

- **Scope highlighted resources to specific collections**
- **Add featurable migrator**

image

image

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -152,7 +152,13 @@ class CatalogController < ApplicationController
       segments: true
     }
     config.add_facet_field "file_type_ssim", label: "File Type"
-    config.add_facet_field "featurable_ssim", label: "Highlighted", helper_method: :display_binary_boolean
+    # Ids of featurable collections. We want to be able to facet on them in
+    # queries, but use the "highlights" facet field for display
+    config.add_facet_field "featurable_ssim", show: false
+    config.add_facet_field "highlighted", label: "Highlighted", query: {
+      highlighted: { label: "Yes", fq: "featurable_ssim:[* TO *]" },
+      not_highlighted: { label: "No", fq: "*:* -featurable_ssim:[* TO *]" }
+    }
     config.add_facet_fields_to_solr_request!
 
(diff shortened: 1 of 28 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -152,7 +152,13 @@ def iiif_search
       segments: true
     }
     config.add_facet_field "file_type_ssim", label: "File Type"
-    config.add_facet_field "featurable_ssim", label: "Highlighted", helper_method: :display_binary_boolean
+    # Ids of featurable collections. We want to be able to facet on them in
+    # queries, but use the "highlights" facet field for display
+    config.add_facet_field "featurable_ssim", show: false
```
tpendragon: Can we add a label to this? Otherwise it looks kinda odd.

Image

tpendragon (changes requested):
Just a small thing. Otherwise it works great! The vue component binding to the select element was a nice touch.

app/controllers/catalog_controller.rb
eliotjordan: Added.

tpendragon (approved):

PR #7331, opened 2026-08-31
Remove Date Facet

tpendragon:
In the last 6 months nobody's clicked it, and it's kind of an odd facet
for a workflow application anyways. Removing this means we can get rid
of a large dependency that doesn't update much and maybe unblock
updating view_component.

Screenshot 2026-08-31 at 1 22 49 PM


eliotjordan (approved):

tpendragon:
Kim uses it, analytics isn't tracking this facet.

PR #7363, opened 2026-09-14
Replace pulmap rabbit messages with sidekiq jobs

eliotjordan:
Work towards https://github.com/pulibrary/pulmap/issues/1310

- Adds configuration and sidekiq jobs to index geoblacklight documents directly into solr
- Removes geoblacklight rabbitmq functionality


app/services/pulmap_indexer.rb
```diff
@@ -0,0 +1,21 @@
...
+    connection.update(
+      params: { overwrite: true },
+      data: [document].to_json,
+      headers: { "Content-Type" => "application/json" }
+    )
+    connection.commit if commit
+  end
+
+  def delete(slug:, commit: true)
+    connection.delete_by_query("layer_slug_s:#{RSolr.solr_escape(slug)}")
```
tpendragon, line 12: I'm nervous about the coupling here. Does this mean Figgy's now basically the indexer for pulmap? ...Was it already?

When we bypassed something similar in Pulfalight, we had jobs in Pulfalight hit an endpoint in Figgy, which queued a job, so there was resilience on both sides and the responsibilities were separated.

I think the equivalent here would hitting a pulmap endpoint with an auth key, which enqueues a job that writes to the index. But maybe that's less of a thing when Figgy's already responsible for generating the solr documents. 🤔

eliotjordan: Yeah. Figgy basically does everything. As an example of how much Figgy is in charge here, when you need to do a full re-index, you have to trigger it from Figgy using a service. The pulmap README has docs for doing this in FIggy.

app/jobs/pulmap_index_job.rb
```diff
@@ -0,0 +1,7 @@
+class PulmapIndexJob < ApplicationJob
+  queue_as :high
```
tpendragon: The other jobs queued as high are the PDF OCR jobs, which will fire an alert if they're left too long. High for delete makes sense to me, to handle take downs, but is it necessary for indexing?

app/services/event_generator/geoblacklight_event_generator.rb
```diff
@@ -43,32 +40,20 @@ def valid?(record)
...
-        base_message(type, record).merge("doc" => document_generator(record))
-      end
-
-      def delete_message(type, record)
-        base_message(type, record).merge("id" => slug(record))
+      def sidekiq_queue
+        if bulk
+          :super_low
+        else
+          :high
```
tpendragon: Same question here.

tpendragon (changes requested):
Just a couple questions about the queue priority.

Can you also add an ADR saying that we're going to use Figgy as the indexing layer for Pulmap?

app/services/event_generator/geoblacklight_event_generator.rb
eliotjordan: Now in :low

app/jobs/pulmap_index_job.rb
eliotjordan: Now in :low

tpendragon (approved):

Issue #7364, opened 2026-09-16
Videos are not playing in Safari browser (Clover viewer not working in Safari)


kelea99:
### Summary (Product Owner)

Notified by Near Eastern Studies Department instructor:

I was recently notified by my students that the Amazigh videos on DPUL do not open in Safari, although they do work properly in Chrome. I wanted to bring this to your attention and ask if you could look into whether this is something that can be fixed.
Thank you so much for your help!

**Definitely a clover viewer and safari thing, @eliotjordan had a look** 

https://dpul.princeton.edu/oral-history-nes

https://figgy.princeton.edu/catalog/a18c497a-6df1-470d-abd7-69488d1e7aad

### Impact (Product Owner)
_Please include hard deadlines, if the exhibit is part of an event, the issue is stopping work, etc._
see below. 

### Priority recommendation (Product Owner)

- [ ] asap
- [x] within the next 3 weeks
- [ ] PO will prioritize

## Sudden Priority Justification (Product Owner)
_Required if "asap" or "within the next 3 weeks" is checked. Add "Sudden Priority" and "dls-work-cycle" labels_
While not a hard deadline, this was noticed immediately by students, meaning safari is a primary browser our Princeton students are using for this class.  a work around exists, but im suggesting within the next 3 weeks due to the school year and the very real probability we will hear this same bug problem continuously from others. Probably worth checking figgy functionality in Safari in general.
### Implementation Notes
_To be filled out by DLS._

### Acceptance Criteria
_To be filled out by DLS.Some requirements to consider, including:_

- [ ] The UI implemented for this issue meets accessibility requirements
- [ ] New functionality is documented

### First Step
_To be filled out by DLS._

eliotjordan:
All fixed now.

PR #7369, opened 2026-09-18
Force Chrome to play video with hls.js

eliotjordan:
Work towards #7364

Restores captions to videos played through Clover on Chrome.


sdellis (approved):
Looks good!

sdellis (changes requested):
Oops, looks like it needs test coverage...

